PR #26, opened 2024-06-11
Allow targeting specific commit/tags of dependencies during compilation

hax0kartik:


brunoerg (approved):
ACK 5c46ec2c5d88971812f7402c56727cdb56c0ce8a

PR #45, opened 2024-06-19
add target for script

brunoerg:


rust_bitcoin_lib/src/lib.rs
```diff
@@ -100,6 +101,23 @@ pub unsafe extern "C" fn rust_bitcoin_psbt(data: *const u8, len: usize) -> *mut
...
+#[no_mangle]
+pub unsafe extern "C" fn rust_bitcoin_script(data: *const u8, len: usize) -> bool {
+    // Safety: Ensure that the data pointer is valid for the given length
+    let data_slice = slice::from_raw_parts(data, len);
+
+    let script: Result<(bitcoin::script::ScriptBuf, usize), encode::Error> = encode::deserialize_partial(data_slice);
+    match script {
+        Err(_) => false,
+        Ok(s) => {
+           if s.0.is_op_return() || s.0.len() > 10_000 {
```
hax0kartik, line 113: I think there are unintentional *minor* formatting issues in this line

targets/script.cpp
```diff
@@ -0,0 +1,32 @@
+#include <fuzzer/FuzzedDataProvider.h>
+#include <string>
```
hax0kartik, line 2: This include is not required

targets/script.cpp
```diff
@@ -0,0 +1,32 @@
...
+#include <iostream>
```
hax0kartik, line 3: +1 not required

rust_bitcoin_lib/src/lib.rs
brunoerg: which one?

targets/script.cpp
brunoerg: thanks

rust_bitcoin_lib/src/lib.rs
brunoerg: got it, fixed

PR #46, opened 2024-06-20
Add PSBT deserialization support in btcd wrapper

hax0kartik:
Relevant code in `psbt.cpp` is commented out as BTCD's PSBT code is unstable and still a WIP(?).

### DIFF
--- a/btcd_lib/go.mod
+++ b/btcd_lib/go.mod
@@ -3,9 +3,10 @@ module main
 go 1.21.6
 
 require (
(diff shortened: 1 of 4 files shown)


brunoerg:
if btcd implementation is not ready, I don't think it's worth adding it for now.

hax0kartik:
I believe we could merge this code and let it be since the relevant is commented out anyway. Once BTCD's PSBT code is ready we would just need to uncomment the code.

PR #54, opened 2024-07-02
Replace fuzzer.cpp with bitcoin core's fuzz.cpp 

hax0kartik:


targets/addrv2.cpp
```diff
@@ -29,9 +29,8 @@ std::optional<std::pair<uint64_t, uint64_t>> Addrv2Core(Span<const uint8_t> buff
     return std::make_pair(addrs.size(), clearnet_count);
 }
 
-void Addrv2(FuzzedDataProvider& provider)
+FUZZ_TARGET(Addrv2)
```
brunoerg, line 32:
```suggestion
FUZZ_TARGET(addrv2)
```

targets/bech32.cpp
```diff
@@ -20,10 +20,8 @@ std::string Bech32Core(Span<const uint8_t> buffer)
 }
 
 
-void Bech32(FuzzedDataProvider& provider) 
+FUZZ_TARGET(Bech32)
```
brunoerg, line 23:
```suggestion
FUZZ_TARGET(bech32)
```

same for the others?

brunoerg (approved):
ACK 385d42939dc71a11448d129e5ddd95c4d1875c8f

PR #56, opened 2024-07-04
Add target for generating header_and_shortids from block

hax0kartik:
BTCD doesn't support compact blocks, so this harness only targets Core and Rust-bitcoin.


brunoerg (approved):
lgtm

PR #61, opened 2024-08-12
Add RPC Target

hax0kartik:
This is a rough draft right now because there are certain areas which need to be improved before merging:

* Initialization : Preferably, we should report to the user in case the btcd server has not been executed yet. +1 if we can launch the server ourselves.

* Code Cleanup : Minor Code cleanup is needed


targets/rpc.cpp
```diff
@@ -0,0 +1,450 @@
...
+        return;
+    }
+
+    const bool safe_for_fuzzing = std::find(RPC_COMMANDS_SAFE_FOR_FUZZING.begin(), RPC_COMMANDS_SAFE_FOR_FUZZING.end(), rpc_command) != RPC_COMMANDS_SAFE_FOR_FUZZING.end();
+    if (!safe_for_fuzzing) {
+        return;
+    }
+
+    std::vector<std::string> arguments;
+    LIMITED_WHILE(good_data && fuzzed_data_provider.ConsumeBool(), 100)
```
brunoerg, line 425: Why do we need `good_data`?

hax0kartik: To not push further to the `arguments` incase `ConsumeRPCArgument` has already failed previously.(Check after lien 280

brunoerg: ah ok, I see. Same of Core.

brunoerg:
Still draft or ready for review?

targets/rpc.cpp
```diff
@@ -0,0 +1,450 @@
...
+    "importwallet", // avoid reading from disk
+    "loadtxoutset",   // avoid reading from disk
+    "loadwallet",   // avoid reading from disk
+    "savemempool",           // disabled as a precautionary measure: may take a file path argument in the future
+    "setban",                // avoid DNS lookups
+    "stop",                  // avoid shutdown state
+};
+
+// RPC commands which are safe for fuzzing.
+const std::vector<std::string> RPC_COMMANDS_SAFE_FOR_FUZZING{
```
brunoerg, line 120: Did you check if btcd has all these RPCs?

brunoerg:
```
In file included from init.cpp:25:
./httpserver.h:134:43: error: no matching constructor for initialization of 'std::span<typename string_view::value_type>' (aka 'span<char>')
        WriteReply(nStatus, std::as_bytes(std::span{reply}));
```

targets/rpc.cpp
hax0kartik: Hi, btcd is definitely missing some of these RPCs, however that is not a problem, since btcd in such cases returns a very specific error code which we check before the asserts

hax0kartik:
> Still draft or ready for review?

I believe I am ready for review

hax0kartik:
> ```
> In file included from init.cpp:25:
> ./httpserver.h:134:43: error: no matching constructor for initialization of 'std::span<typename string_view::value_type>' (aka 'span<char>')
>         WriteReply(nStatus, std::as_bytes(std::span{reply}));
> ```

Are you compiling with the most recent commit of bitcoin core? I tested commit `60b816439e` and I see no such issues.

brunoerg:
> Are you compiling with the most recent commit of bitcoin core? I tested commit 60b816439e and I see no such issues.

Yes, and tried on master.

brunoerg (approved):
lgtm

PR #71, opened 2024-12-19
rust-miniscript: use `Descriptor::parse_descriptor`

brunoerg:


brunoerg:
cc: @apoelstra @sanket1729

apoelstra:
This looks right to me, but agreed that this is unfortunate and non-obvious. Will think a bit about whether we can somehow extend `MiniscriptKey` to understand secret keys as well.

PR #87, opened 2025-03-07
Add BOLT-11 Invoice Decoding Implementation

erickcestari:
### Overview

This PR adds the implementation of BOLT-11 invoice decoding functionality for multiple Lightning Network implementations (LND, LDK and nLightning).

### Changes

- Added nLightning implementation for BOLT-11 invoice decoding
- Added LND implementation for BOLT-11 invoice decoding
- Added LDK implementation for BOLT-11 invoice decoding
- Started initial work on doing diferential fuzzing on lightning implementations

### Co-Authors

Níckolas Goline (@nGoline)
Lucas de Oliveira (@lucasxciv)
Erick Cestari (@erickcestari)
Moises Pompilio (@moisesPompilio)


brunoerg:
Some commits should be squashed. No need to have a commit that adds a fuzz corpora and another one that removes it. Same for the commit that adds an implementation and right after update it. By squashing the appropriate commits you will facilitate the review process.

brunoerg (commented):
Concept ACK! Amazing! 🔥

brunoerg:
Needs rebase

erickcestari:
@brunoerg I think now it should be ready for review!

erickcestari:
@brunoerg The PR is now ready for review. Thanks to @moisesPompilio for cleaning up the commit history.

driver.cpp
```diff
@@ -103,6 +103,22 @@ namespace bitcoinfuzz
         }
     }
 
+    void Driver::InvoiceDeserializationTarget(std::span<const uint8_t> buffer) const
+    {
+        FuzzedDataProvider provider(buffer.data(), buffer.size());
+        std::string desc{provider.ConsumeRemainingBytesAsString()};
```
brunoerg: I don't understand why  `deserialize_invoice` uses the buffer instead of using the string directly. Note that `desc` here is unused, probably leftover from a copy/paste. However, I think you could rename this variable and use it on `deserialize_invoice` since all the implementations is converting the buffer to string.

erickcestari: I think i get it. We should send the string directly to those FFI functions rather than relying on byte-to-string conversion functions in each language. Is that why it's better to use strings instead of sending buffers to the implementations?

brunoerg: Yes, and would simplify the code.

brunoerg (approved):
LGTM

I will take a better look on Monday

PR #91, opened 2025-03-11
Update Build Configuration in GitHub Actions Workflow

erickcestari:
### Changes

- Run the workflow on each pr to verify that builds are passing
- Added .NET SDK installation for nlightning builds
- Added proper configuration for all module build steps:
  - Rust Miniscript
  - BTCD
  - LND
  - LDK
  - NLightning

### Benefit
Full build testing of all project modules

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -26,6 +26,11 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version: "stable"
+      
+      - name: Install .NET SDK
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: '8.0.x'
 
       - name: Install LLVM and Clang
         run: |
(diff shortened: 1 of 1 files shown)


brunoerg (commented):
Concept ACK

Nice!

.github/workflows/workflow.yml
```diff
@@ -56,6 +61,43 @@ jobs:
           rustup default nightly
           cd modules/rustbitcoin && make cargo && make
 
+      - name: Build - rust miniscript
+        timeout-minutes: 5
+        run: |
+          export CC=/usr/bin/clang
```
brunoerg: Instead of repeating the export CC/CXX, we could set it "globally", e.g.:
```
      - name: Setup common environment
        run: |
          export CC=/usr/bin/clang
          export CXX=/usr/bin/clang++
          echo "CC=/usr/bin/clang" >> $GITHUB_ENV
          echo "CXX=/usr/bin/clang++" >> $GITHUB_ENV
```

brunoerg:
You're building bitcoinfuzz only with rust bitcoin and Bitcoin Core (`/usr/bin/clang++ -DBITCOIN_CORE -DRUST_BITCOIN -fsanitize=address,fuzzer -Wall -Wextra -std=c++20 -I include -I . -c driver.cpp -o driver.o`) you should set the other modules as well.

brunoerg:
You must drop the commit [1aa1768](https://github.com/brunoerg/bitcoinfuzz/pull/91/commits/1aa176897e8958d7eee98686a06b0f5254df44d4) since it did not work and there is no need to run all targets for now. Since we have problems with rustbitcoin + rustminiscript and btcd + lnd, let's build them but let's leave CXXFLAGS as it was for now.

brunoerg (changes requested):
You must drop the commit [1aa1768](https://github.com/brunoerg/bitcoinfuzz/pull/91/commits/1aa176897e8958d7eee98686a06b0f5254df44d4) since it did not work and there is no need to run all targets for now. Since we have problems with rustbitcoin + rustminiscript and btcd + lnd, let's build them but let's leave CXXFLAGS as it was for now.

brunoerg (approved):
ACK 182c19ca3fdb67e9ad7016aeb33edc01a7dd787a

PR #104, opened 2025-03-19
Fix macOS build issues by explicitly detecting OS

### LINKED ISSUE #102: Linker Errors for Missing Class Definitions (Architecture: arm64)
### Problem
Encountering linker errors when building bitcoinfuzz on macOS (M2 chip). The compilation process works fine, but during the linking stage, I get errors about undefined symbols

### Environment:
`OS: macOS(15.2)`
**Compiler**:
`CC: clang`
`CXX: clang++`
```
clang --version
Homebrew clang version 19.1.3
Target: arm64-apple-darwin24.2.0
Thread model: posix
InstalledDir: /opt/homebrew/Cellar/llvm/19.1.3/bin
Configuration file: /opt/homebrew/etc/clang/arm64-apple-darwin24.cfg
```

```
cat version.hpp | grep "BOOST_VERSION"
#ifndef BOOST_VERSION_HPP
#define BOOST_VERSION_HPP
//  BOOST_VERSION % 100 is the patch level
//  BOOST_VERSION / 100 % 1000 is the minor version
//  BOOST_VERSION / 100000 is the major version
#define BOOST_VERSION 108500
//  BOOST_LIB_VERSION must be defined to be the same as BOOST_VERSION

```

**Steps Taken:**
- Build `module/rust_bitcoin`: `no error`
- Build `module/rust-miniscript`: `no error`
- First build [brunoerg/mako](https://github.com/brunoerg/mako/tree/bitcoinfuzz) in some other directory and used the `libmako.a` file generated by it to build `module/mako`: `no error`
- Build `module/btcd `: `no error`
- Build `module/Bitcoin`: `no error`
- Build `module/LDK`
- Build `module\IND`
- Build `modules/nlightning`: `no error`
- But when building bitcoin fuzz getting the following error:

Image

erickcestari:
The Makefile was previously relying on the UNAME_S environment variable to detect macOS systems, but this variable isn't automatically set in all environments. This caused linking errors on macOS when the Darwin-specific flags weren't applied.

**Changes:**

- Added explicit OS detection using shell uname -s command
- Set UNAME_S variable at the beginning of the Makefile

Should fix: #102

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,9 @@ all: bitcoinfuzz
 CXX := clang++
 CXXFLAGS += -fsanitize=address,fuzzer -Wall -Wextra -std=c++20 -I include -I .
 MODULES := $(wildcard modules/*/module.a)
+UNAME_S := $(shell uname -s)
 
-ifeq ($(UNAME_S), Darwin)
+ifeq ($(UNAME_S),Darwin)
 LDFLAGS = -framework CoreFoundation -Wl,-ld_classic
 endif
 


brunoerg (approved):
ACK bbc399bc1b52e97253d5d819aa8a5e3f2f2ee399

PR #106, opened 2025-03-20
add Embit module for Miniscript parsing

erickcestari:
Add support for Embit, a Python-based Bitcoin library, as a new module to BitcoinFuzz:

- Create Embit module structure with Makefile and necessary files
- Implement Python bindings to call Embit's Miniscript parser
- Add integration with main driver and update build system for Python integration
- Update .gitignore to exclude Python cache files and main.py
- Update workflow to build and run Embit module

The module uses Python C API to bridge between C++ and the Embit Python library for parsing Miniscript expressions in both taproot and non-taproot contexts.

add embit on miniscript_parse

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,22 @@ make
 export CXXFLAGS="$CXXFLAGS -DBTCD"
 ```
 
+### embit
+
+To run the fuzzer with `embit` module, you need to install the `embit` library.
+
+To install the `embit` library, you can use the following command:
+```bash
+cd modules/embit
+pip install -r embit_lib/requirements.txt
+```
+
+```bash
+cd modules/embit
+make
+export CXXFLAGS="$CXXFLAGS -DEMBIT"
+```
+
 ### Bitcoin Core
 
 ```bash
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -31,6 +31,14 @@ jobs:
         uses: actions/setup-dotnet@v4
         with:
           dotnet-version: '8.0.x'
+      
+      - name: Setup Python and install embit
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+      - run: |
+          python -m pip install --upgrade pip
+          pip install -r modules/embit/embit_lib/requirements.txt
 
       - name: Install LLVM and Clang
         run: |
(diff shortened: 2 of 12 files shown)


brunoerg:
Nice!

README.md
```diff
@@ -80,6 +80,21 @@ make
 export CXXFLAGS="$CXXFLAGS -DBTCD"
 ```
 
+### embit
+
+To run the fuzzer with `embit` module, you need to install the `embit` library.
+
+To install the `embit` library, you can use the following command:
```
brunoerg, line 87: Why installing `embit` instead of bringing files here?

erickcestari: Installing the `embit` library is more practical than bringing individual files into our project because:

- Makes it easier to update to newer versions of the library as they're released
- Simplifies maintenance since we don't need to track and update individual files
- Follows the same dependency management approach used by rust-bitcoin and rust-miniscript modules

This is a topic we should standardize to have clear guidelines on when to copy files versus using dependencies. In my opinion, we should always use library dependencies when possible. Only when that's not an option (as with bitcoin-core) should we bring the necessary files directly into our codebase.

modules/embit/embit_lib/embit.py
```diff
@@ -0,0 +1,14 @@
+
+from io import BytesIO
+from embit.descriptor.miniscript import Miniscript
+
+def miniscript_parse(input):
+    try:
+        ms = Miniscript.read_from(BytesIO(input.encode()), taproot=False)
```
brunoerg: Does it check if it is safe as a script on its own (like `IsSafe` from Bitcoin Core)?

erickcestari: Looking again at the code it has a verify function that does some sanity checks. I will add it.

https://github.com/diybitcoinhardware/embit/blob/master/src/embit/descriptor/miniscript.py

brunoerg: Cool, I'll test it again.

README.md
brunoerg: I agree that it simplify maintenance, but it would not allow us to test specific changes and we would only be able to test releases.

erickcestari: You can install old or unreleased versions of Embit using Pip by specifying a commit hash or branch. I plan to improve this by adding a requirements.txt file, which I believe will be more maintainable and easier to install/change embit versions.

brunoerg: Fair enough, let's move on this way.

brunoerg (approved):
ACK 217a6942ced6321ccabbe3c5673915859ed496c2

Issue #96, opened 2025-03-13
Add Lightning Core support to invoice_deserialization target


erickcestari:
We need to integrate Lightning Core BOLT11 invoice parsing into our existing invoice_deserialization target.

### Technical Challenges
- Lightning Core has a complex architecture with numerous interdependent modules
- The project uses an extensive Makefile build system
- Integration will require careful extraction of the relevant components

### Proposed Solution

1. Identify the specific BOLT11 invoice parsing modules in Lightning Core
2. Create a new wrapper file that exposes only the necessary decode function from Lightning Core
3. Copy the required dependencies and source files to our project
4. Create a simplified Makefile that compiles only the necessary Lightning Core components and links them directly with our C++ code, leveraging C/C++ interoperability through extern "C" declarations

PR #110, opened 2025-03-23
Fix: Add Lightning Core support to invoice_deserialization target #96

rajanarahul93:
### 🛠 Problem  
BitcoinFuzz lacked support for BOLT11 invoice deserialization, making it impossible to decode Lightning Core invoices. The challenge was integrating Lightning Core’s BOLT11 parsing without breaking existing modules.  

### 🔍 What Was Fixed?  
1. **Linked Rust's `ldk_des_invoice` function properly** in `InvoiceBridge.cs` to ensure C# can call the Rust function correctly.  
2. **Updated `Makefile`** to ensure `libldk_lib.a` is built before linking with C++.  
3. **Fixed `lib.rs` function signature** to properly expose `ldk_des_invoice` using `extern "C"`.  
4. **Improved error handling** in C# and Rust to prevent crashes when parsing invalid invoices.  

### ✅ Changes Implemented  
- 📌 **`InvoiceBridge.cs`** → Fixed DllImport path & added debug logs.  
- 📌 **`Makefile`** → Ensured Rust library builds before linking.  
- 📌 **`lib.rs`** → Improved error handling and function compatibility.  

### 🔬 Testing  
After rebuilding (`make clean && make`), BOLT11 invoices successfully decode, and the function returns `true` for valid invoices.  

 **BitcoinFuzz now supports BOLT11 invoice parsing!** 

solves issue #96

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 all: bitcoinfuzz
 
 CXX := clang++
-CXXFLAGS += -fsanitize=address,fuzzer -Wall -Wextra -std=c++20 -I include -I .
+CXXFLAGS += -fsanitize=address,fuzzer -Wall -Wextra -std=c++20 -I include -I . -Lmodules/ldk/ldk_lib -lldk_lib
 MODULES := $(wildcard modules/*/module.a)
 UNAME_S := $(shell uname -s)
 
@@ -9,7 +9,7 @@ ifeq ($(UNAME_S),Darwin)
 LDFLAGS = -framework CoreFoundation -Wl,-ld_classic
 endif
 
-bitcoinfuzz: main.cpp driver.o include/bitcoinfuzz/basemodule.o
+bitcoinfuzz: modules/ldk/ldk_lib/libldk_lib.a main.cpp driver.o include/bitcoinfuzz/basemodule.o
 	$(CXX) $(CXXFLAGS) $(LDFLAGS) main.cpp $(MODULES) driver.o include/bitcoinfuzz/basemodule.o -o bitcoinfuzz
 
 driver.o: driver.cpp driver.h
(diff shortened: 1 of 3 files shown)


pingu-73:
Hi thanks for the pr!
I have technical concerns regarding your implementation:
- creating a `interdependency between nlightning and ldk` modules defeats the purpose of differential fuzzing
- These changes would make nlightning and ldk return `identical results b'coz they're using the same code path`, which doesn't help identify discrepancies between implementations

ig your approach `might be deviating` from what was intended in #96 
@erickcestari you confirm if these concerns are valid?

FWIW could you please reduce the use of emojis while creating a PR

rajanarahul93:
Hey @pingu-73 , thanks for taking the time to review my PR!  

I see your point about the interdependency between `nlightning` and `ldk`. My main goal was to get BOLT11 invoice deserialization working, but I get that this might not align with the original intent of #96, especially when it comes to differential fuzzing.  

You're right—having both modules use the same code path kind of defeats the purpose of comparing different implementations. One way to fix this could be to keep `nlightning` handling its own deserialization logic separately while still supporting `ldk` in a way that doesn't interfere. Would that be a better approach?  

Let me know what you think, and I can make adjustments accordingly. Also, noted on keeping PR descriptions more formal—I'll cut down on the emojis next time!  

Looking forward to your thoughts!

pingu-73:
> One way to fix this could be to keep nlightning handling its own deserialization logic separately while still supporting ldk in a way that doesn't interfere. Would that be a better approach?
> Let me know what you think, and I can make adjustments accordingly.

Please revisit the solution proposed by @erickcestari in #96 it's very elaborate and provides a solid starting point. Also refer to other Bitcoin modules present in the repo as they all follow a similar ideology.
If you’re still unsure about the approach I suggest commenting on the issue to ask for clarification. I’ll be happy to explain what is meant in #96 but please make sure to `thoroughly review the proposed solution first`.

rajanarahul93:
Okay !!

brunoerg:
It has nothing to do with #96 and do not add anything related to https://github.com/ElementsProject/lightning. Can you explain it better? Otherwise, I can close it for now.

rajanarahul93:
Hey @brunoerg , thanks for reviewing this!

I now understand that my changes don’t align with #96 . My goal was to enable BOLT11 invoice deserialization, but I see that I may have taken the wrong approach.

To clarify:

I linked ldk_des_invoice to enable BOLT11 parsing, assuming it was needed for Lightning Core invoice support.

However, if this is not relevant to #96 and shouldn’t be part of BitcoinFuzz, I can reevaluate my approach.

Would you recommend a different direction, or should I close this PR and reassess the solution based on the repository's existing structure? Let me know how you'd like to proceed.

moisesPompilio:
Hello

This issue #96 is the implementation of a new module in the project, this module is [core lightning](https://github.com/ElementsProject/lightning). This module will be part of the project's lightning implementation coverage. It will basically have to have the deserialization of invoices from Bolt11 to core lightning, but I think it is not a good first issue because core lightning is complex. I recommend closing this PR, because its implementation is not correct to solve this issue. But you can still try to make this issue if you want.

PR #111, opened 2025-03-24
add embit descriptor parsing support and fix embit miniscript_parse

erickcestari:
This PR updates embit by modifying `miniscript_parse` to use the higher-level `from_string` function and adding support for `descriptor_parse`.

- Refactor Python interface code by extracting common parsing logic into call_python_parser function 
- Add new descriptor_parse function to Python layer using embit's Descriptor.from_string() 
- Add corresponding C++ binding function embit_descriptor_parse 
- Update miniscript parsing to use Miniscript.from_string() instead of read_from(BytesIO()) 
- Export the new descriptor parse function in the header file Add embit on miniscript_parse workflow

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -128,7 +128,7 @@ jobs:
       - name: Test - descriptor_parse
         timeout-minutes: 5
         run: |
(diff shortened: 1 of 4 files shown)


brunoerg (approved):
ACK abef12f4c0fc601ff2a9c65fa2b01cec56627910

PR #115, opened 2025-03-25
enhance Invoice Deserialization to Return Detailed Information

### LINKED ISSUE #112: Improve `deserialize_invoice` target
We currently have the `deserialize_invoice` target which basically pass a string to the implementations and they return a boolean indicating whether the invoice was successfully deserialized. We could improve it to instead of returning a boolean, return some contents from the invoice.

- Which fields would be interesting to get?
- How could we return them (concatenated string)?

erickcestari:
This pr improves the deserialize_invoice functionality across all implementations:

- Change return type from bool to string to provide rich invoice content
- Implement consistent format for invoice data across all implementations: HASH=<payment_hash>;AMOUNT=<milli_sats>;DESCRIPTION=<desc>;RECIPIENT=<pubkey>; EXPIRY=<seconds>;TIMESTAMP=<unix_time>;ROUTING_HINTS=<count>;MIN_CLTV=<value>
- Ensure consistent memory management for returned strings in all modules

These changes enable consistent testing across different Lightning implementations, allowing for detailed comparison of parsed invoice fields rather than just success/failure validation.

Closes #112

### DIFF
--- a/modules/ldk/ldk_lib/src/lib.rs
+++ b/modules/ldk/ldk_lib/src/lib.rs
@@ -1,32 +1,104 @@
-use lightning_invoice::Currency;
+use lightning_invoice::{
+    Bolt11InvoiceDescriptionRef, Bolt11SemanticError, Currency, ParseOrSemanticError,
+};
+use std::ffi::CString;
+use std::os::raw::c_char;
 use std::{ffi::CStr, str::FromStr};
 
+unsafe fn str_to_c_string(input: &str) -> *mut c_char {
+    CString::new(input).unwrap().into_raw()
+}
+
 #[no_mangle]
-pub unsafe extern "C" fn ldk_des_invoice(input: *const std::os::raw::c_char) -> bool {
+pub unsafe extern "C" fn ldk_des_invoice(input: *const std::os::raw::c_char) -> *mut c_char {
     if input.is_null() {
-        return false;
+        return str_to_c_string("");
     }
 
     // Convert C string to Rust string
     let c_str = match CStr::from_ptr(input).to_str() {
         Ok(s) => s,
-        Err(_) => return false,
+        Err(_) => return str_to_c_string(""),
     };
 
-    match lightning_invoice::SignedRawBolt11Invoice::from_str(c_str) {
+    match lightning_invoice::Bolt11Invoice::from_str(c_str) {
         Ok(invoice) => {
-            // If the destination pubkey was provided as a tagged field, use that
-            // to verify the signature, otherwise recover it from the signature
-            let is_signature_valid = if let Some(_) = invoice.payee_pub_key() {
-                invoice.check_signature()
-            } else {
-                invoice.recover_payee_pub_key().is_ok()
-            };
-            let is_currency_bitcoin = invoice.currency() == Currency::Bitcoin;
-            let has_payment_hash = invoice.payment_hash().is_some();
+            if invoice.currency() != Currency::Bitcoin {
+                return str_to_c_string("");
+            }
+            let mut result = String::new();
 
-            has_payment_hash && is_currency_bitcoin && is_signature_valid
+            result.push_str("HASH=");
+            result.push_str(&invoice.payment_hash().to_string());
+
+            result.push_str(";AMOUNT=");
+            if let Some(amount) = invoice.amount_milli_satoshis() {
+                result.push_str(&amount.to_string());
+            }
+
+            result.push_str(";DESCRIPTION=");
+            if let Bolt11InvoiceDescriptionRef::Direct(direct_description) = invoice.description() {
+                result.push_str(&direct_description.to_string());
+            }
+
+            let invoice_payee_pub_key = match invoice.payee_pub_key() {
+                Some(payee) => payee.clone(),
+                None => invoice.recover_payee_pub_key(),
+            };
+
+            result.push_str(";RECIPIENT=");
+            result.push_str(&invoice_payee_pub_key.to_string());
+
+            result.push_str(";EXPIRY=");
+            result.push_str(&invoice.expiry_time().as_secs().to_string());
+
+            result.push_str(";TIMESTAMP=");
+            result.push_str(
+                &invoice
+                    .clone()
+                    .into_signed_raw()
+                    .raw_invoice()
+                    .data
+                    .timestamp
+                    .as_unix_timestamp()
+                    .to_string(),
+            );
+
+            result.push_str(";ROUTING_HINTS=");
+            result.push_str(&invoice.private_routes().len().to_string());
+
+            result.push_str(";MIN_CLTV=");
+            result.push_str(&invoice.min_final_cltv_expiry_delta().to_string());
+
+            str_to_c_string(&result)
         }
-        Err(_) => false,
+        // Handle invoices without payment secrets by returning null
+        // This is needed because some Lightning implementations don't require payment secrets,
+        // and we need to maintain compatibility with these implementations
+        Err(ParseOrSemanticError::SemanticError(Bolt11SemanticError::NoPaymentSecret)) => {
+            std::ptr::null_mut()
+        }
+        // Handle invoices with multiple payment hashes by returning null
+        // This is needed because some Lightning implementations don't require payment to have only one hash,
+        // and we need to maintain compatibility with these implementations
+        Err(ParseOrSemanticError::SemanticError(Bolt11SemanticError::MultiplePaymentHashes)) => {
+            std::ptr::null_mut()
+        }
+        // Handle invoices with multiple descriptions hashes by returning null
+        // This is needed because some Lightning implementations don't require to have only one description,
+        // and we need to maintain compatibility with these implementations
+        Err(ParseOrSemanticError::SemanticError(Bolt11SemanticError::MultipleDescriptions)) => {
+            std::ptr::null_mut()
+        }
+        Err(_) => str_to_c_string(""),
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn ldk_free_string(ptr: *mut c_char) {
+    if !ptr.is_null() {
+        unsafe {
+            let _ = CString::from_raw(ptr);
+        }
     }
 }
--- a/modules/lnd/lnd_wrapper/wrapper.go
+++ b/modules/lnd/lnd_wrapper/wrapper.go
@@ -7,16 +7,18 @@ package main
 import "C"
 
 import (
+	"fmt"
 	"runtime"
+	"strings"
 
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/lightningnetwork/lnd/zpay32"
 )
 
 //export LndDeserializeInvoice
-func LndDeserializeInvoice(cInvoiceStr *C.char) C.int {
+func LndDeserializeInvoice(cInvoiceStr *C.char) *C.char {
 	if cInvoiceStr == nil {
-		return 0
+		return C.CString("")
 	}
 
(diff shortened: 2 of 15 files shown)


erickcestari:
While running fuzzing tests on the current commit, I noticed that LND sets certain default values when they are missing from the invoice. This causes the fuzzing process to crash consistently, as LDK and NLightning do not assign default values.

Default values in LND:

`MIN_CLTV = 18`

`EXPIRY = 3600`

brunoerg:
Needs rebase

morehouse:
> While running fuzzing tests on the current commit, I noticed that LND sets certain default values when they are missing from the invoice. This causes the fuzzing process to crash consistently, as LDK and NLightning do not assign default values.

That's annoying.  A few options:

1. Ignore these fields for now.
2. Modify your fork of LND to export the `minFinalCLTVExpiry` and `expiry` fields.  Then access them directly.
3. Use [reflection](https://stackoverflow.com/questions/42664837/how-to-access-unexported-struct-fields) to extract the actual values of `minFinalCLTVExpiry` and `expiry`.
4. Modify the wrappers for other implementations to set the same defaults as LND.

modules/ldk/ldk_lib/src/lib.rs
```diff
@@ -1,16 +1,22 @@
...
+unsafe fn str_to_c_string(input: &str) -> *mut c_char {
+    CString::new(input).unwrap().into_raw()
+}
 
 #[no_mangle]
-pub unsafe extern "C" fn ldk_des_invoice(input: *const std::os::raw::c_char) -> bool {
+pub unsafe extern "C" fn ldk_des_invoice(input: *const std::os::raw::c_char) -> *mut c_char {
     if input.is_null() {
-        return false;
+        return str_to_c_string("0");
```
morehouse: Nit:  can an empty string be returned instead?

brunoerg:
> > While running fuzzing tests on the current commit, I noticed that LND sets certain default values when they are missing from the invoice. This causes the fuzzing process to crash consistently, as LDK and NLightning do not assign default values.
> 
> That's annoying. A few options:
> 
> 1. Ignore these fields for now.
> 2. Modify your fork of LND to export the `minFinalCLTVExpiry` and `expiry` fields.  Then access them directly.
> 3. Use [reflection](https://stackoverflow.com/questions/42664837/how-to-access-unexported-struct-fields) to extract the actual values of `minFinalCLTVExpiry` and `expiry`.
> 4. Modify the wrappers for other implementations to set the same defaults as LND.

I think we can ignore these fields for now.

modules/ldk/ldk_lib/src/lib.rs
```diff
@@ -24,9 +30,65 @@ pub unsafe extern "C" fn ldk_des_invoice(input: *const std::os::raw::c_char) ->
...
             let has_payment_hash = invoice.payment_hash().is_some();
+            if has_payment_hash && is_currency_bitcoin && is_signature_valid {
+                let mut result = String::new();
+
+                result.push_str("HASH=");
+                result.push_str(&invoice.payment_hash().unwrap().0.to_string());
+
+                result.push_str(";AMOUNT=");
+                if let Some(amount) = invoice.amount_pico_btc()  {
+                    let amount_milisatoshis = amount/10;
```
brunoerg: If you need the milisatoshis, you could use `amount_milli_satoshis` instead of `amount_pico_btc`. Otherwise, I think this `amount/10` could bring some issues.

modules/ldk/ldk_lib/src/lib.rs
```diff
@@ -1,16 +1,22 @@
...
     }
 
     // Convert C string to Rust string
     let c_str = match CStr::from_ptr(input).to_str() {
         Ok(s) => s,
-        Err(_) => return false,
+        Err(_) => return str_to_c_string("0"),
     };
 
     match lightning_invoice::SignedRawBolt11Invoice::from_str(c_str) {
```
brunoerg: Per our conversation, we should use `Bolt11Invoice` since it does the validations.

erickcestari:
> > While running fuzzing tests on the current commit, I noticed that LND sets certain default values when they are missing from the invoice. This causes the fuzzing process to crash consistently, as LDK and NLightning do not assign default values.
> 
> That's annoying. A few options:
> 
> 1. Ignore these fields for now.
> 2. Modify your fork of LND to export the `minFinalCLTVExpiry` and `expiry` fields.  Then access them directly.
> 3. Use [reflection](https://stackoverflow.com/questions/42664837/how-to-access-unexported-struct-fields) to extract the actual values of `minFinalCLTVExpiry` and `expiry`.
> 4. Modify the wrappers for other implementations to set the same defaults as LND.

Good news. Using Bolt11invoice struct, it has correct default cltv and expiry time like LND

brunoerg:
Squash the commits?

modules/ldk/ldk_lib/src/lib.rs
```diff
@@ -1,32 +1,78 @@
...
-            // If the destination pubkey was provided as a tagged field, use that
-            // to verify the signature, otherwise recover it from the signature
-            let is_signature_valid = if let Some(_) = invoice.payee_pub_key() {
-                invoice.check_signature()
-            } else {
-                invoice.recover_payee_pub_key().is_ok()
-            };
             let is_currency_bitcoin = invoice.currency() == Currency::Bitcoin;
-            let has_payment_hash = invoice.payment_hash().is_some();
+            if is_currency_bitcoin  {
```
morehouse: Nit: we can simplify and reduce nesting a bit:

```rust
if invoice.currency() != Currency::Bitcoin {
  return str_to_c_string("0");
}
...
```

modules/lnd/lnd_wrapper/wrapper.go
```diff
@@ -7,16 +7,18 @@ package main
...
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/lightningnetwork/lnd/zpay32"
 )
 
 //export LndDeserializeInvoice
-func LndDeserializeInvoice(cInvoiceStr *C.char) C.int {
+func LndDeserializeInvoice(cInvoiceStr *C.char) *C.char {
 	if cInvoiceStr == nil {
-		return 0
+		return C.CString("0")
```
morehouse: Nit:  can we use an empty string?

modules/lnd/lnd_wrapper/wrapper.go
```diff
@@ -26,14 +28,51 @@ func LndDeserializeInvoice(cInvoiceStr *C.char) C.int {
...
 
-	_, err := zpay32.Decode(invoiceStr, network)
+	invoice, err := zpay32.Decode(invoiceStr, network)
 	if err != nil {
-		return 0
+		return C.CString("0")
 	}
 
-	runtime.GC()
+	var result string
```
morehouse: Nit:  strings.Builder will be more efficient with all these concatenations

modules/lnd/module.cpp
```diff
@@ -9,11 +9,12 @@ namespace bitcoinfuzz
...
         Lnd::Lnd(void) : BaseModule("Lnd") {}
 
-        std::optional<bool> Lnd::deserialize_invoice(std::string str) const
+        std::optional<std::string> Lnd::deserialize_invoice(std::string str) const
         {
-            bool result = LndDeserializeInvoice(str.c_str());
-            return result;
+            auto result = LndDeserializeInvoice(str.c_str());
+            std::string result_str(result);
+            LndFreeString(result);
```
morehouse: Nit:  I think we could actually use `free` here, and remove `LndFreeString` entirely.

modules/nlightning/src/Bolts/BOLT11/InvoiceBridge.cs
```diff
@@ -2,26 +2,76 @@
...
             string? invoiceString = Marshal.PtrToStringUTF8(invoiceStringPtr);
             if (string.IsNullOrEmpty(invoiceString))
             {
-                return false;
+                return IntPtr.Zero;
             }
 
-            _ = Invoice.Decode(invoiceString);
-            return true;
+            string result = string.Empty;
```
morehouse: Nit:  I'm not a C# expert, but I think StringBuilder is the more efficient way to do this.

morehouse (commented):
Basically LGTM, with some nits.

Only reviewed the NLightning stuff lightly -- I have 0 experience with C#.

brunoerg:
I could build and run the target (during the weekend) with no issues - will take a better look on the code today.

modules/ldk/ldk_lib/src/lib.rs
```diff
@@ -1,32 +1,104 @@
...
+            result.push_str(";RECIPIENT=");
+            result.push_str(&invoice_payee_pub_key.to_string());
+
+            result.push_str(";EXPIRY=");
+            result.push_str(&invoice.expiry_time().as_secs().to_string());
+
+            result.push_str(";TIMESTAMP=");
+            result.push_str(
+                &invoice
+                    .clone()
```
brunoerg, line 58: nit: Why do we need to clone here?

brunoerg (approved):
ACK e6756c89b24251989be579bdc0b3fc574ad52f1b

modules/ldk/ldk_lib/src/lib.rs
erickcestari: The `clone()` is used because the `into_signed_raw()` method consumes (takes ownership of) the invoice. 

Just out of curiosity about rust: "into_" is commonly used for methods that consume the original value and transform it into something else.

Without clone:
```
error[E0382]: borrow of moved value: invoice
    --> src/lib.rs:67:30
     |
25   |         Ok(invoice) => {
     |            ------- move occurs because invoice has type Bolt11Invoice, which does not implement the Copy trait
...
58   |                     .into_signed_raw()
     |                      ----------------- invoice moved due to this method call
...
67   |             result.push_str(&invoice.private_routes().len().to_string());
     |                              ^^^^^^^ value borrowed here after move
     |
note: Bolt11Invoice::into_signed_raw takes ownership of the receiver self, which moves invoice
```

brunoerg: Make sense, thanks.

modules/ldk/ldk_lib/src/lib.rs
```diff
@@ -1,32 +1,104 @@
...
+            result.push_str(";MIN_CLTV=");
+            result.push_str(&invoice.min_final_cltv_expiry_delta().to_string());
+
+            str_to_c_string(&result)
+        }
+        // Handle invoices without payment secrets by returning null
+        // This is needed because some Lightning implementations don't require payment secrets,
+        // and we need to maintain compatibility with these implementations
+        Err(ParseOrSemanticError::SemanticError(Bolt11SemanticError::NoPaymentSecret)) => {
+            std::ptr::null_mut()
```
morehouse, line 79: So empty string means invalid invoice, and nullptr means ignore the invoice when comparing.

Is there a comment explaining this distinction somewhere in the driver?  If not, it might make sense to add one.

morehouse (approved):
Haven't tested, but code looks good.

modules/ldk/ldk_lib/src/lib.rs
brunoerg: > So empty string means invalid invoice, and nullptr means ignore the invoice when comparing.

Yes, we use `nullptr` when we want to skip/ignore stuff. That's why the targets are all `std::optional`. An example of usage: https://github.com/brunoerg/bitcoinfuzz/issues/118. But yes, a comment explaining it would be really helpful.

PR #117, opened 2025-03-27
Refactor Embit module naming and implement descriptor_parse

erickcestari:
This pr streamlines the Embit module by cleaning up function names and implementing the missing descriptor_parse functionality. Changes include:

- Rename embit_miniscript_miniscript_parse to embit_miniscript_parse to eliminate redundant naming
- Rename embit_miniscript_descriptor_parse to embit_descriptor_parse for consistency
- Add implementation for descriptor_parse method in the Embit class
- Add proper header includes in module.cpp and remove unused imports
- Fix the Embit descriptor_parse function call to target the correct implementation

These changes complete the Embit interface implementation and improve overall code organization.

I forgot to add the descriptor function into the embit module.cpp... Now this is fixed!

### DIFF
--- a/modules/embit/module.cpp
+++ b/modules/embit/module.cpp
@@ -1,8 +1,7 @@
 #include <span>
 
 #include "module.h"
-
-extern "C" bool embit_miniscript_miniscript_parse(const char* input);
+#include "embit_lib/embit_lib.h"
 
 namespace bitcoinfuzz
 {
@@ -12,9 +11,12 @@ namespace bitcoinfuzz
 
         std::optional<bool> Embit::miniscript_parse(std::string str) const
         {
-            bool result = embit_miniscript_miniscript_parse(str.c_str());
-            return result;
+            return embit_miniscript_parse(str.c_str());
         }
 
+        std::optional<bool> Embit::descriptor_parse(std::string str) const
+        {
+            return embit_descriptor_parse(str.c_str());
+        }
     }
 }
--- a/modules/embit/embit_lib/embit_lib.cpp
+++ b/modules/embit/embit_lib/embit_lib.cpp
@@ -72,7 +72,7 @@ cleanup:
     return success;
 }
 
-extern "C" bool embit_miniscript_miniscript_parse(const char* input) {
+extern "C" bool embit_miniscript_parse(const char* input) {
     return call_python_parser(input, "miniscript_parse");
 }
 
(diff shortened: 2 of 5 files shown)


modules/embit/module.cpp
```diff
@@ -12,9 +11,14 @@ namespace bitcoinfuzz
...
         std::optional<bool> Embit::miniscript_parse(std::string str) const
         {
-            bool result = embit_miniscript_miniscript_parse(str.c_str());
+            bool result = embit_miniscript_parse(str.c_str());
             return result;
         }
 
+        std::optional<bool> Embit::descriptor_parse(std::string str) const
+        {
+            bool result = embit_descriptor_parse(str.c_str());
```
brunoerg: For both descriptor and miniscript parse you can return embit_descriptor_parse/embit_miniscript_parse directly. No need to have the `bool result`.

brunoerg:
I think Embit misses a lot of descriptor stuff? Any descriptor with thresh or combo makes it fail. cc @odudex

odudex:
As far as I know, there are no issues with thresholds. Only booleans, `0` and `1` being used as arguments for logical operators `and`, `or`.
Can you elaborate on the combo issue?

brunoerg:
> As far as I know, there are no issues with thresholds. Only booleans, `0` and `1` being used as arguments for logical operators `and`, `or`. Can you elaborate on the combo issue?

With the following descriptor `combo(02FFFFFFFFFFFFFFFF778fecFFFFFFfFFFFfFFFFFFFfDFFFFFFFF6FFFeFFFFFC2F)`, I get from Embit the error "Invalid descriptor (starts with 'combo(0')"

brunoerg (approved):
ACK 5234dd8ca6e7050547aac067477a37f62e54dd7b

odudex:
> I get from Embit the error "Invalid descriptor (starts with 'combo(0')"

`combo` is not implemented indeed.

brunoerg:
So, I think the embit module could return `std::nullopt` for the fragments they haven't implemented yet instead of crashing due to errors. It can be done in follow-up, no need to be here, sounds good?

odudex:
> instead of crashing due to errors

In application it won't crash, error will be raised and dealt with, script won't be loaded, but application keeps running fine.
An issue can be created for the implementation. But for now, AFAIK, no device or compatible coordinator demands it.

brunoerg:
> In application it won't crash, error will be raised and dealt with, script won't be loaded, but application keeps running fine.
> An issue can be created for the implementation. But for now, AFAIK, no device or compatible coordinator demands it.

Sorry, @odudex. I meant crash during our fuzzing campaign, not related to any application. So, instead of returning false due to non-implemented fragments which would cause a discrepance with other implementations that can deal with it, we could simply return std::nullopt and avoid comparison here. It would mean "we do not have it, skip".

brunoerg:
Just created #118 issue

PR #121, opened 2025-04-01
[Feature Enhancement] Adding support for PSBT parse fuzzing target

pingu-73:
## Feature Added:
Adds a new fuzzing target for psbt parsing in bitcoin core and rust-bitcoin

## Changes Made:
- Current bitcoin core module doesn't have psbt.h file, Have added psbt.* and it's dependencies from [bitcoin](https://github.com/bitcoin/bitcoin/tree/master/src).
- Modified module/bitcoin/module.* to add psbt parse function.


pingu-73:
cc: @brunoerg 
Is there some other way to add psbt support for bitcoin core other than copying psbt and its dependency from [bitcoin](https://github.com/bitcoin/bitcoin/tree/master/src)?

brunoerg (changes requested):
Thanks for it, but please, review all the files you're including from Bitcoin Core. I'm sure that there are a lot of files that have nothing to do with PSBT. (e.g. addrman, banman, bip324)

pingu-73:
> Thanks for it, but please, review all the files you're including from Bitcoin Core. I'm sure that there are a lot of files that have nothing to do with PSBT. (e.g. addrman, banman, bip324)

they are not directly associated with psbt.* but psbt has dependencies on some files that were not present initially and they have dependencies on `addrman, banman, bip324, leveldb`.

PR #125, opened 2025-04-03
bolt11: use custom mutator

### LINKED ISSUE #120: bolt11: use `core-lightning` custom mutator

brunoerg:
Fixes #120 

This PR adds the custom mutator from core lightning's bolt11 target to the `deserialize_invoice` target here. I did some modifications due to C++ and the bech32 functions used are from Bitcoin Core (so it's required to build bitcoin core module).

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -146,7 +146,7 @@ jobs:
 
       - name: Test - deserialize_invoice
         run: |
-          export CXXFLAGS="-DLDK -DLND -DNLIGHTNING -DCUSTOM_MUTATOR_BOLT11"
+          export CXXFLAGS="-DLDK -DLND -DCUSTOM_MUTATOR_BOLT11"
           make
           ASAN_OPTIONS=detect_leaks=0 FUZZ=deserialize_invoice ./bitcoinfuzz -runs=100
       
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -66,6 +66,11 @@ jobs:
           export BOOST_LIB_DIR=/usr/lib/x86_64-linux-gnu/
           cd modules/bitcoin && make
 
+      - name: Build - bolt11mutator
+        timeout-minutes: 40
+        run: |
+          cd modules/bolt11mutator && make
+
       - name: Build - rust bitcoin
         timeout-minutes: 5
         run: |
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -146,7 +146,7 @@ jobs:
 
       - name: Test - deserialize_invoice
         run: |
-          export CXXFLAGS="-DLDK -DLND -DNLIGHTNING"
+          export CXXFLAGS="-DLDK -DLND -DNLIGHTNING -DCUSTOM_MUTATOR_BOLT11"
           make
           ASAN_OPTIONS=detect_leaks=0 FUZZ=deserialize_invoice ./bitcoinfuzz -runs=100
       
--- a/main.cpp
+++ b/main.cpp
@@ -41,7 +41,7 @@
 #endif
 
 #ifdef CUSTOM_MUTATOR_BOLT11
-#include "modules/bitcoin/bech32.h"
+#include <modules/bolt11mutator/bech32.h>
 #endif
 
 std::shared_ptr<bitcoinfuzz::Driver> driver = nullptr;
--- a/main.cpp
+++ b/main.cpp
@@ -1,4 +1,6 @@
+#include <algorithm>
 #include <memory>
+#include <cstring>
 #include "driver.h"
 #include <bitcoinfuzz/basemodule.h>
 
@@ -38,7 +40,285 @@
 #include <modules/embit/module.h>
 #endif
 
+#ifdef CUSTOM_MUTATOR_BOLT11
+#include "modules/bitcoin/bech32.h"
+#endif
+
 std::shared_ptr<bitcoinfuzz::Driver> driver = nullptr;
+
+#ifdef CUSTOM_MUTATOR_BOLT11
+// We use a custom mutator to produce an input corpus that consists entirely of
+// correctly encoded bech32 strings. This enables us to efficiently fuzz the
+// bolt11 decoding logic without the fuzzer getting stuck on fuzzing the bech32
+// decoding/encoding logic. This custom mutator is originally from core-lightning:
+// https://github.com/ElementsProject/lightning/blob/3a7a1fad4eb56522b6ab590e53c695e2fb08e7e2/tests/fuzz/fuzz-bolt11.c#L59
+//
+// This custom mutator does the following things:
+//   1. Attempt to bech32 decode the given input (returns the encoded dummy
+//      invoice on failure).
+//   2. Mutate either the human readable or data part of the invoice using
+//      libFuzzer's default mutator `LLVMFuzzerMutate`.
+//   3. Attempt to bech32 encode the mutated hrp and data (returns the endcoded
+//      dummy on failure).
+//   4. Write the encoded result to `fuzz_data` if its size does not exceed
+//      `max_size`, otherwise return the encoded dummy invoice.
+extern "C" size_t LLVMFuzzerMutate(uint8_t *Data, size_t Size, size_t MaxSize);
+extern "C" size_t LLVMFuzzerCustomMutator(uint8_t *fuzz_data, size_t size, size_t max_size,
+			       unsigned int seed);
+extern "C" size_t LLVMFuzzerCustomCrossOver(const uint8_t *in1, size_t in1_size, const uint8_t *in2,
+				 size_t in2_size, uint8_t *out, size_t max_out_size,
+				 unsigned seed);
+
+// Encodes a dummy bolt11 invoice into `fuzz_data` and returns the size of the
+// encoded string.
+static size_t initial_input(uint8_t *fuzz_data, size_t max_size)
+{
+    constexpr std::string_view dummy =
+        "lnbc16lta047pp5h6lta047h6lta047h6lta047h6lta047h6lta047h6lta047"
+	    "h6lqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq"
+	    "qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqxnht6w";
+
+    size_t output_size = std::min(max_size, dummy.size());
+    std::memcpy(fuzz_data, dummy.data(), output_size);
+    return output_size;
+}
+
+size_t LLVMFuzzerCustomMutator(uint8_t *fuzz_data, size_t size, size_t max_size,
+                               unsigned int seed)
+{
+    // A minimum size of 9 prevents hrp_maxlen <= 0 and data_maxlen <= 0.
+    if (size < 9)
+        return initial_input(fuzz_data, max_size);
+
+    // Interpret fuzz input as string
+    std::string input(reinterpret_cast<char*>(fuzz_data), size);
+
+    size_t data_maxlen = input.size() - 8;
+    size_t hrp_maxlen = input.size() - 6;
+
+    // Attempt to bech32 decode the input
+    bech32::DecodeResult decoded = bech32::Decode(input, bech32::CharLimit::CUSTOM_MUTATOR);
+    if (decoded.encoding != bech32::Encoding::BECH32) {
+        // Decoding failed, this should only happen when starting from
+        // an empty corpus.
+        return initial_input(fuzz_data, max_size);
+    }
+
+    auto data = decoded.data;
+    auto hrp = decoded.hrp;
+
+    // Mutate either the hrp or data
+    std::srand(seed);
+    switch (std::rand() % 2) {
+        case 0: { // Mutate hrp
+            // Make sure we have a buffer that's large enough for mutation
+            std::vector<uint8_t> hrp_buffer(hrp.begin(), hrp.end());
+            // Reserve enough space for the maximum mutation size
+            hrp_buffer.resize(hrp_maxlen);
+
+            // Mutate the buffer
+            size_t new_len = LLVMFuzzerMutate(hrp_buffer.data(),
+                                             hrp.size(),
+                                             hrp_maxlen);
+
+            // Convert back to string with the new length
+            hrp = std::string(reinterpret_cast<char*>(hrp_buffer.data()), new_len);
+
+            // Sanitize hrp - ensure only valid ASCII characters (33-126) and not uppercase
+            for (char& c : hrp) {
+                if (c < 33 || c > 126) {
+                    c = 'a' + (c % 26); // Replace with a valid lowercase letter
+                } else if (c >= 'A' && c <= 'Z') {
+                    c = c - 'A' + 'a'; // Convert uppercase to lowercase
+                }
+            }
+            break;
+        }
+        case 1: { // Mutate data
+            size_t original_data_size = data.size();
+            // Make sure we have a buffer that's large enough for mutation
+            data.resize(std::max(data.size(), data_maxlen));
+
+            size_t new_len = LLVMFuzzerMutate(data.data(),
+                                                original_data_size,
+                                                data_maxlen);
+
+            // It ensures that all values remain valid 5-bit values
+            for (auto& val : data) {
+                val &= 0x1F;
+            }
+            break;
+        }
+    }
+
+    if (!hrp.empty()) {
+        for (const char& c : hrp) {
+            if (c >= 'A' && c <= 'Z') {
+                return initial_input(fuzz_data, max_size);
+            }
+        }
+    }
+
+    std::string output = bech32::Encode(bech32::Encoding::BECH32, hrp, data);
+
+    if (output.length() > max_size) {
+        return initial_input(fuzz_data, max_size);
+    }
+
+    std::memcpy(fuzz_data, output.data(), output.length());
+    return output.length();
+}
+
+static std::vector<uint8_t> insert_part(std::span<const uint8_t> in1,
+                                      std::span<const uint8_t> in2,
+                                      size_t max_out_size)
+{
+    size_t in1_size = in1.size();
+    size_t in2_size = in2.size();
+    std::vector<uint8_t> out;
+
+    if (in1_size >= max_out_size)
+        return out;
+    if (in1_size == 0 || in2_size == 0)
+        return out;
+
+    size_t max_insert_size = max_out_size - in1_size;
+    if (max_insert_size > in2_size) max_insert_size = in2_size;
+    size_t insert_begin = std::rand() % in1_size;
+    size_t insert_size = (std::rand() % max_insert_size) + 1;
+
+    size_t in2_begin = std::rand() % (in2_size - insert_size + 1);
+
+    size_t total_size = in1_size + insert_size;
+    out.reserve(total_size);
+
+    out.insert(out.end(), in1.begin(), in1.begin() + insert_begin);
+    out.insert(out.end(), in2.begin() + in2_begin, in2.begin() + in2_begin + insert_size);
+    out.insert(out.end(), in1.begin() + insert_begin, in1.end());
+
+    return out;
+}
+
+static std::vector<uint8_t> overwrite_part(std::span<const uint8_t> in1,
+                                          std::span<const uint8_t> in2,
+                                          size_t max_out_size)
+{
+    std::vector<uint8_t> out;
+    out.reserve(max_out_size);
+
+    if (in1.empty() || in2.empty()) {
+        return out;
+    }
+
+    // Copy in1 to out first (limited by max_out_size)
+    size_t in1_bytes = std::min(in1.size(), max_out_size);
+    out.insert(out.end(), in1.begin(), in1.begin() + in1_bytes);
+
+    if (out.empty()) return out;
+
+    size_t pos = std::rand() % out.size();
+    size_t max_possible = std::min(in2.size(), out.size() - pos);
+    size_t overwrite_length = (std::rand() % max_possible) + 1;
+
+    size_t in2_start = 0;
+    if (in2.size() > overwrite_length) {
+        in2_start = std::rand() % (in2.size() - overwrite_length + 1);
+    }
+
+    // Overwrite portion of out with random substring from in2
+    std::copy_n(in2.begin() + in2_start, overwrite_length, out.begin() + pos);
+
+    return out;
+}
+
+static std::vector<uint8_t> cross_over(std::span<const uint8_t> in1,
+                                      std::span<const uint8_t> in2,
+                                      size_t max_out_size,
+                                      unsigned seed)
+{
+    std::srand(seed);
+    if (std::rand() % 2) {
+        return insert_part(in1, in2, max_out_size);
+    } else {
+        return overwrite_part(in1, in2, max_out_size);
+    }
+}
+
+size_t LLVMFuzzerCustomCrossOver(const uint8_t *in1, size_t in1_size, const uint8_t *in2,
+                               size_t in2_size, uint8_t *out, size_t max_out_size,
+                               unsigned seed)
+{
+    if (in1_size < 9 || in2_size < 9)
+        return 0;
+
+    // Interpret fuzz inputs as strings
+    std::string input1(reinterpret_cast<const char*>(in1), in1_size);
+    std::string input2(reinterpret_cast<const char*>(in2), in2_size);
+
+    // Attempt to bech32 decode the inputs
+    bech32::DecodeResult result1 = bech32::Decode(input1, bech32::CharLimit::CUSTOM_MUTATOR);
+    if (result1.encoding != bech32::Encoding::BECH32) {
+        // Decoding failed
+        return 0;
+    }
+
+    bech32::DecodeResult result2 = bech32::Decode(input2, bech32::CharLimit::CUSTOM_MUTATOR);
+    if (result2.encoding != bech32::Encoding::BECH32) {
+        // Decoding failed
+        return 0;
+    }
+
+    std::string hrp1 = result1.hrp;
+    std::string hrp2 = result2.hrp;
+    std::vector<uint8_t> data1 = result1.data;
+    std::vector<uint8_t> data2 = result2.data;
+
+    std::srand(seed);
+    std::string out_hrp;
+    std::vector<uint8_t> out_data;
+
+    if (std::rand() % 2) {
+        // Cross-over the HRP
+        out_data = data1;
+
+        std::span<const uint8_t> hrp1_span(reinterpret_cast<const uint8_t*>(hrp1.data()), hrp1.size());
+        std::span<const uint8_t> hrp2_span(reinterpret_cast<const uint8_t*>(hrp2.data()), hrp2.size());
+
+        size_t max_out_data_size = max_out_size - data1.size() - 8;
+
+        auto out_hrp_vec = cross_over(
+            hrp1_span,
+            hrp2_span,
+            max_out_data_size,
+            static_cast<unsigned>(std::rand()));
+
+        // Convert back to string and ensure null termination
+        out_hrp = std::string(out_hrp_vec.begin(), out_hrp_vec.begin() + out_hrp_vec.size());
+    } else {
+        // Cross-over the data part
+        out_hrp = hrp1;
+
+        size_t max_out_data_size = max_out_size - hrp1.size() - 8;
+
+        out_data = cross_over(
+            data1,
+            data2,
+            max_out_data_size,
+            static_cast<unsigned>(std::rand()));
+    }
+
+    // Encode the output
+    std::string encoded = bech32::Encode(bech32::Encoding::BECH32, out_hrp, out_data);
+
+    if (encoded.size() > max_out_size) {
+        return 0;
+    }
+
+    // Copy the result to out buffer
+    std::memcpy(out, encoded.data(), encoded.size());
+    return encoded.size();
+}
+#endif
 
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size)
 {
--- a/driver.cpp
+++ b/driver.cpp
@@ -112,10 +112,7 @@ namespace bitcoinfuzz
         {
             std::optional<std::string> res{module.second->deserialize_invoice(invoice)};
             if (!res.has_value()) continue;
-            if (last_response.has_value()) {
-                assert(*res == *last_response);
-            }
-
+            if (last_response.has_value()) assert(*res == *last_response);
             last_response = res.value();
         }
     }
@@ -133,7 +130,7 @@ namespace bitcoinfuzz
             std::optional<std::string> res{module.second->address_parse(address)};
             if(!res.has_value()) continue;
 
-            if(last_response.has_value()) 
+            if(last_response.has_value())
             {
                 if(*res != *last_response)
                 {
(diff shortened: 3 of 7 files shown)


brunoerg:
cc @morehouse @erickcestari

erickcestari:
I am having three issues, may some of these are not related to this pr:

1. When trying to build it, I got some errors.
```
make
clang++ -DLDK -DLND -DCUSTOM_MUTATOR_BOLT11 -fsanitize=address,fuzzer -Wall -Wextra -std=c++20 -I include -I .  main.cpp modules/ldk/module.a modules/lnd/module.a driver.o include/bitcoinfuzz/basemodule.o -o bitcoinfuzz 
main.cpp:65:10: error: no member named 'memcpy' in namespace 'std'
   65 |     std::memcpy(fuzz_data, dummy.data(), output_size);
      |     ~~~~~^
main.cpp:57:56: warning: unused parameter 'size' [-Wunused-parameter]
   57 | static size_t initial_input(uint8_t *fuzz_data, size_t size, size_t max_size)
      |                                                        ^
main.cpp:182:10: error: no member named 'memcpy' in namespace 'std'
  182 |     std::memcpy(fuzz_data, output.data(), output.length());
      |     ~~~~~^
main.cpp:197:10: error: no member named 'memcpy' in namespace 'std'
  197 |     std::memcpy(out, in1, insert_begin);
      |     ~~~~~^
main.cpp:198:10: error: no member named 'memcpy' in namespace 'std'
  198 |     std::memcpy(out + insert_begin, in2 + in2_begin, insert_size);
      |     ~~~~~^
main.cpp:199:10: error: no member named 'memcpy' in namespace 'std'
  199 |     std::memcpy(out + insert_begin + insert_size, in1 + insert_begin,
      |     ~~~~~^
main.cpp:216:10: error: no member named 'memcpy' in namespace 'std'
  216 |     std::memcpy(out, in1, in1_size);
      |     ~~~~~^
main.cpp:217:10: error: no member named 'memcpy' in namespace 'std'
  217 |     std::memcpy(out + overwrite_begin, in2 + in2_begin, overwrite_size);
      |     ~~~~~^
main.cpp:305:10: error: no member named 'memcpy' in namespace 'std'
  305 |     std::memcpy(out, encoded.data(), encoded.size());
      |     ~~~~~^
1 warning and 8 errors generated.
make: *** [Makefile:16: bitcoinfuzz] Error 1
```

Adding this include fixes the errors.
`#include <cstring> `

2. After including cstring and trying to compile i got this error:

```
make
clang++ -DLDK -DLND -DCUSTOM_MUTATOR_BOLT11 -fsanitize=address,fuzzer -Wall -Wextra -std=c++20 -I include -I .  main.cpp modules/ldk/module.a modules/lnd/module.a driver.o include/bitcoinfuzz/basemodule.o -o bitcoinfuzz 
main.cpp:58:56: warning: unused parameter 'size' [-Wunused-parameter]
   58 | static size_t initial_input(uint8_t *fuzz_data, size_t size, size_t max_size)
      |                                                        ^
1 warning generated.
/usr/bin/ld: /tmp/main-4a10ba.o: in function `LLVMFuzzerCustomMutator':
main.cpp:(.text.LLVMFuzzerCustomMutator[LLVMFuzzerCustomMutator]+0x4f1): undefined reference to `bech32::Decode(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bech32::CharLimit)'
/usr/bin/ld: main.cpp:(.text.LLVMFuzzerCustomMutator[LLVMFuzzerCustomMutator]+0x1b12): undefined reference to `bech32::Encode(bech32::Encoding, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::vector<unsigned char, std::allocator<unsigned char> > const&)'
/usr/bin/ld: /tmp/main-4a10ba.o: in function `LLVMFuzzerCustomCrossOver':
main.cpp:(.text.LLVMFuzzerCustomCrossOver[LLVMFuzzerCustomCrossOver]+0x597): undefined reference to `bech32::Decode(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bech32::CharLimit)'
/usr/bin/ld: main.cpp:(.text.LLVMFuzzerCustomCrossOver[LLVMFuzzerCustomCrossOver]+0x6b6): undefined reference to `bech32::Decode(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bech32::CharLimit)'
/usr/bin/ld: main.cpp:(.text.LLVMFuzzerCustomCrossOver[LLVMFuzzerCustomCrossOver]+0x1158): undefined reference to `bech32::Encode(bech32::Encoding, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::vector<unsigned char, std::allocator<unsigned char> > const&)'
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [Makefile:16: bitcoinfuzz] Error 1
```

3. When trying to build the bitcoin core I am receiving this error:
```
CMake Error in CMakeLists.txt:
  No cmake_minimum_required command is present.  A line of code such as

    cmake_minimum_required(VERSION 4.0)

  should be added at the top of the file.  The version specified may be lower
  if you wish to support older CMake versions for this project.  For more
  information run "cmake --help-policy CMP0000".


-- Configuring incomplete, errors occurred!
make: *** [Makefile:35: ../univalue/build/libunivalue.a] Error 1
```

Similar to the one that appeared on this ci: https://github.com/brunoerg/bitcoinfuzz/actions/runs/14203350756/job/39795216201#step:11:228

erickcestari:
Other thing that I noticed is that our ci is not recompiling the bitcoinfuzz for each test in CI. #128

brunoerg:
I will put on the docs that to use this custom mutator it's required to build the Bitcoin Core module for now. So fixing the Bitcoin Core module build should work, I'm checking it. 

It's weird that this build error wasn't occurring before, since we did not change anything on this? (?)

brunoerg:
rebased

brunoerg:
CI failure is unrelated. It's a crash on miniscript_parse target.

main.cpp
```diff
@@ -38,8 +39,275 @@
...
+				 unsigned seed);
+
+// Encodes a dummy bolt11 invoice into `fuzz_data` and returns the size of the
+// encoded string.
+static size_t initial_input(uint8_t *fuzz_data, size_t size, size_t max_size)
+{
+    const std::string dummy =
+        "lnbc16lta047pp5h6lta047h6lta047h6lta047h6lta047h6lta047h6lta047"
+	    "h6lqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq"
+	    "qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqxnht6w";
```
morehouse: Nit: this can be a compile-time constant:

```suggestion
    constexpr std::string_view dummy =
        "lnbc16lta047pp5h6lta047h6lta047h6lta047h6lta047h6lta047h6lta047"
	    "h6lqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq"
	    "qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqxnht6w";
```

main.cpp
```diff
@@ -38,8 +39,275 @@
 #include <modules/embit/module.h>
 #endif
 
+#ifdef CUSTOM_MUTATOR_BOLT11
+#include "modules/bitcoin/bech32.h"
+#endif
+
 std::shared_ptr<bitcoinfuzz::Driver> driver = nullptr;
 
+#ifdef CUSTOM_MUTATOR_BOLT11
```
morehouse, line 49: Can this whole `ifdef` be moved to `modules/bitcoin/bech32.h` or somewhere else, so that `main.cpp` doesn't get polluted?

main.cpp
```diff
@@ -38,8 +39,275 @@
...
+//   2. Mutate either the human readable or data part of the invoice using
+//      libFuzzer's default mutator `LLVMFuzzerMutate`.
+//   3. Attempt to bech32 encode the mutated hrp and data (returns the endcoded
+//      dummy on failure).
+//   4. Write the encoded result to `fuzz_data` if its size does not exceed
+//      `max_size`, otherwise return the encoded dummy invoice.
+extern "C" size_t LLVMFuzzerMutate(uint8_t *Data, size_t Size, size_t MaxSize);
+
+size_t LLVMFuzzerCustomMutator(uint8_t *fuzz_data, size_t size, size_t max_size,
+                               unsigned int seed)
```
morehouse, line 87: I'm not super familiar with all the license rules, but this mutator is mostly copied from CLN.  Probably bitcoinfuzz needs an open source license in order to use this code...

It's probably also good to credit CLN as the original source of the mutator.

main.cpp
```diff
@@ -38,8 +39,275 @@
...
+// This custom mutator does the following things:
+//   1. Attempt to bech32 decode the given input (returns the encoded dummy
+//      invoice on failure).
+//   2. Mutate either the human readable or data part of the invoice using
+//      libFuzzer's default mutator `LLVMFuzzerMutate`.
+//   3. Attempt to bech32 encode the mutated hrp and data (returns the endcoded
+//      dummy on failure).
+//   4. Write the encoded result to `fuzz_data` if its size does not exceed
+//      `max_size`, otherwise return the encoded dummy invoice.
+extern "C" size_t LLVMFuzzerMutate(uint8_t *Data, size_t Size, size_t MaxSize);
```
morehouse: Nit: the `LLVMFuzzerMutate` definition should be moved further up so it doesn't separate the `LLVMFuzzerCustomMutator` definition from its function comment.

main.cpp
```diff
@@ -38,8 +39,275 @@
...
+        // Decoding failed, this should only happen when starting from
+        // an empty corpus.
+        return initial_input(fuzz_data, size, max_size);
+    }
+
+    auto data = decoded.data;
+    auto hrp = decoded.hrp;
+
+    size_t data_maxlen = input.size() > 8 ? std::min(input.size() - 8, static_cast<size_t>(65)) : 0;
+    size_t hrp_maxlen = input.size() > 6 ? std::min(input.size() - 6, static_cast<size_t>(83)) : 0;
```
morehouse: Nit:  IMO brace initialization is cleaner than `static_cast`:

```suggestion
    size_t data_maxlen = input.size() > 8 ? std::min(input.size() - 8, size_t{65}) : 0;
    size_t hrp_maxlen = input.size() > 6 ? std::min(input.size() - 6, size_t{83}) : 0;
```

main.cpp
```diff
@@ -38,8 +39,275 @@
...
+            if (hrp_maxlen > 0) {
+                // Make sure we have a buffer that's large enough for mutation
+                std::vector<uint8_t> hrp_buffer(hrp.begin(), hrp.end());
+                // Reserve enough space for the maximum mutation size
+                hrp_buffer.resize(hrp_maxlen);
+
+                // Mutate the buffer
+                size_t new_len = LLVMFuzzerMutate(hrp_buffer.data(),
+                                                 hrp.size(),
+                                                 hrp_maxlen - 1);
```
morehouse: Can we use `hrp_maxlen` without subtracting 1?  IIRC CLN's mutator left an extra byte to add the null terminator, which is needed for CLN's parsing.  But I don't think bitcoinfuzz needs that null terminator.

main.cpp
```diff
@@ -38,8 +39,275 @@
...
+
+                // Ensure minimum length and validate - at least 1 character
+                new_len = std::max(new_len, static_cast<size_t>(1));
```
morehouse:
```suggestion
                new_len = std::max(new_len, size_t{1});
```

main.cpp
```diff
@@ -38,8 +39,275 @@
...
+                std::vector<uint8_t> hrp_buffer(hrp.begin(), hrp.end());
+                // Reserve enough space for the maximum mutation size
+                hrp_buffer.resize(hrp_maxlen);
+
+                // Mutate the buffer
+                size_t new_len = LLVMFuzzerMutate(hrp_buffer.data(),
+                                                 hrp.size(),
+                                                 hrp_maxlen - 1);
+
+                // Ensure minimum length and validate - at least 1 character
```
morehouse: If libFuzzer truncates the HRP or data part to length 0, why do we ignore that and enforce len 1 instead?

main.cpp
```diff
@@ -38,8 +39,275 @@
...
+        }
+        case 1: { // Mutate data
+            if (data_maxlen > 0) {
+                // Make sure we have a buffer that's large enough for mutation
+                data.resize(std::max(data.size(), data_maxlen));
+
+                size_t datalen = data.size();
+                size_t new_len = LLVMFuzzerMutate(data.data(),
+                                                 std::min(datalen, data_maxlen),
+                                                 data_maxlen);
```
morehouse: I don't think this is doing the right thing...

1. `data` get's resized to `data_maxlen`.
2. `datalen` gets set to `data_maxlen`.
3. `LLVMFuzzerMutate` is called with `data_maxlen`.

Shouldn't `LLVMFuzzerMutate` be called with the original `data` size, not the resized size?

main.cpp
```diff
@@ -38,8 +39,275 @@
...
+
+                new_len = std::max(new_len, static_cast<size_t>(1));
```
morehouse:
```suggestion
                new_len = std::max(new_len, size_t{1});
```

main.cpp
morehouse: Where do the 65 and 83 magic constants come from?

main.cpp
```diff
@@ -38,8 +39,275 @@
...
+        }
+    }
+
+    if (hrp.empty() || hrp.size() > 83) {
+        return initial_input(fuzz_data, size, max_size);
+    } else if (!hrp.empty()) {
+        for (const char& c : hrp) {
+            if (c >= 'A' && c <= 'Z') {
+                return initial_input(fuzz_data, size, max_size);
+            }
```
morehouse: How do we get in this case, given that we enforce lowercase above?

main.cpp
```diff
@@ -38,8 +39,275 @@
...
+        // Convert strings to vectors for cross_over function
+        std::vector<uint8_t> hrp1_vec(hrp1.begin(), hrp1.end());
+        std::vector<uint8_t> hrp2_vec(hrp2.begin(), hrp2.end());
+        std::vector<uint8_t> out_hrp_vec(max_out_size - data1.size() - 8);
+
+        size_t out_hrp_size = cross_over(
+            hrp1_vec.data(), hrp1_vec.size(),
+            hrp2_vec.data(), hrp2_vec.size(),
+            out_hrp_vec.data(), out_hrp_vec.size(),
+            (unsigned)std::rand());
```
morehouse: static_cast?

brunoerg: Done

main.cpp
brunoerg: It's from the BIP173 and other implementations but feel free to ignore, I just removed them.

main.cpp
brunoerg: Yes, you're right. I just fixed it.

main.cpp
brunoerg: I've changed this part and removed it.

main.cpp
brunoerg: Yes, done.

main.cpp
brunoerg: Done

main.cpp
brunoerg: I just added a license here and credited CLN for this mutator.

main.cpp
brunoerg: I'm thinking about a better place to move it to. I agree that `main.cpp` isn't the most elegant place for it.

main.cpp
brunoerg: Done

main.cpp
brunoerg: Done

main.cpp
brunoerg: Done

main.cpp
brunoerg: We enforce only if we mutate the hrp part, right? If it takes the path to mutate the data, the HRP is left unchanged, and if it originally contained uppercase letters, those aren't sanitized.

brunoerg:
Thanks, @morehouse. Just addressed your points.

brunoerg:
I just added a module called `bolt11mutator`. It contains only the bech32 files from Bitcoin Core, very light and simple. This way we don't need to compile the entire Bitcoin Core module to be able to use the bech32 stuff for the mutator.

.github/workflows/workflow.yml
```diff
@@ -52,7 +52,11 @@ jobs:
       - name: Install Boost
         run: |
           sudo apt install libboost-all-dev
-      
+
+      - name: Install libsodium
+        run: |
+          sudo apt install libsodium-dev
```
morehouse: I don't see libsodium required anywhere in this PR.

.github/workflows/workflow.yml
```diff
@@ -146,19 +155,18 @@ jobs:
 
       - name: Test - deserialize_invoice
         run: |
-          export CXXFLAGS="-DLDK -DLND -DNLIGHTNING"
```
morehouse, line 149: Why is `-DNLIGHTNING` removed?

main.cpp
```diff
@@ -38,8 +39,271 @@
 #include <modules/embit/module.h>
 #endif
 
+#ifdef CUSTOM_MUTATOR_BOLT11
+#include <modules/bolt11mutator/bech32.h>
+#endif
+
 std::shared_ptr<bitcoinfuzz::Driver> driver = nullptr;
 
+#ifdef CUSTOM_MUTATOR_BOLT11
```
morehouse, line 49: Can we move this to the new `bolt11mutator` module?

modules/bitcoin/bech32.h
```diff
@@ -37,6 +37,7 @@ enum class Encoding {
  *  and we would never encode an address with such a massive value */
 enum CharLimit : size_t {
     BECH32 = 90,            //!< BIP173/350 imposed character limit for Bech32(m) encoded addresses. This guarantees finding up to 4 errors.
+    CUSTOM_MUTATOR = 100'000,
```
morehouse: Can we remove this now that there's a separate module for the custom mutator?

main.cpp
```diff
@@ -38,8 +39,271 @@
...
+    }
+
+    auto data = decoded.data;
+    auto hrp = decoded.hrp;
+
+    // Mutate either the hrp or data
+    std::srand(seed);
+    switch (std::rand() % 2) {
+        case 0: { // Mutate hrp
+            if (hrp_maxlen > 0) {
```
morehouse: I don't think it's possible for `hrp_maxlen` to be 0. We short circuit if `size < 9`, so the min value for `hrp_maxlen` is 3.

main.cpp
```diff
@@ -38,8 +39,271 @@
...
+                size_t original_data_size = data.size();
+                // Make sure we have a buffer that's large enough for mutation
+                data.resize(std::max(data.size(), data_maxlen));
+
+                size_t new_len = LLVMFuzzerMutate(data.data(),
+                                                 original_data_size,
+                                                 data_maxlen);
+
+                new_len = std::max(new_len, size_t{1});
+                data.resize(new_len);
```
morehouse: This in-place resize and mutation of `data` should also be possible for `hrp` -- `std::string` allows resizing too.

main.cpp
```diff
@@ -38,8 +39,271 @@
...
+                        c = 'a' + (c % 26); // Replace with a valid lowercase letter
+                    } else if (c >= 'A' && c <= 'Z') {
+                        c = c - 'A' + 'a'; // Convert uppercase to lowercase
+                    }
+                }
+            }
+            break;
+        }
+        case 1: { // Mutate data
+            if (data_maxlen > 0) {
```
morehouse: AFAICT `data_maxlen` is always at least 1 here, so we can remove this check.

main.cpp
```diff
@@ -38,8 +39,271 @@
...
+                size_t original_data_size = data.size();
+                // Make sure we have a buffer that's large enough for mutation
+                data.resize(std::max(data.size(), data_maxlen));
+
+                size_t new_len = LLVMFuzzerMutate(data.data(),
+                                                 original_data_size,
+                                                 data_maxlen);
+
+                new_len = std::max(new_len, size_t{1});
```
morehouse: Can we remove this as well, like we did for the HRP?

main.cpp
```diff
@@ -38,8 +39,271 @@
...
+                data.resize(new_len);
+
+                for (auto& val : data) {
+                    val &= 0x1F;
+                }
```
morehouse: A comment would be helpful here.

main.cpp
```diff
@@ -38,8 +39,271 @@
...
+            }
+            break;
+        }
+    }
+
+    if (hrp.empty() || hrp.size() > 83) {
```
morehouse: We should not be enforcing the BIP 173 limits.  BOLT 11 explicitly allows larger invoices:  https://github.com/lightning/bolts/blob/master/11-payment-encoding.md#requirements

main.cpp
```diff
@@ -38,8 +39,271 @@
...
+        return initial_input(fuzz_data, size, max_size);
```
morehouse: Can we just allow an empty HRP or data part to see how implementations handle it?

main.cpp
```diff
@@ -38,8 +39,271 @@
...
+        }
+    }
+
+    if (data.empty() || data.size() > 65) {
+        return initial_input(fuzz_data, size, max_size);
+    }
+
+    std::string output = bech32::Encode(bech32::Encoding::BECH32, hrp, data);
+
+    if (output.empty() || output.length() > max_size) {
```
morehouse: I don't think it's possible for `output` to be empty here.

main.cpp
```diff
@@ -38,8 +39,271 @@
...
+    std::string out_hrp;
+    std::vector<uint8_t> out_data;
+
+    if (std::rand() % 2) {
+        // Cross-over the HRP
+        out_data = data1;
+
+        // Convert strings to vectors for cross_over function
+        std::vector<uint8_t> hrp1_vec(hrp1.begin(), hrp1.end());
+        std::vector<uint8_t> hrp2_vec(hrp2.begin(), hrp2.end());
```
morehouse: Do the HRPs actually need to be converted?  Maybe we can just use `data()` and `size()` on the strings themselves.

We might also be able to do this for `out_hrp`.

main.cpp
```diff
@@ -38,8 +39,271 @@
...
+        size_t max_out_data_size = max_out_size - hrp1.size() - 8;
+        std::vector<uint8_t> out_data_vec(max_out_data_size);
+
+        size_t out_data_size = cross_over(
+            data1.data(), data1.size(),
+            data2.data(), data2.size(),
+            out_data_vec.data(), max_out_data_size,
+            (unsigned)std::rand());
+
+        out_data.assign(out_data_vec.begin(), out_data_vec.begin() + out_data_size);
```
morehouse: We can use `out_data` directly.  Also should use `static_cast`:

```suggestion
        out_data.resize(max_out_size - hrp1.size() - 8);

        size_t out_data_size = cross_over(
            data1.data(), data1.size(),
            data2.data(), data2.size(),
            out_data.data(), out_data.size(),
            static_cast<unsigned>(std::rand()));

        out_data.resize(out_data_size);
```

main.cpp
```diff
@@ -38,8 +39,271 @@
...
+    }
+
+    // Encode the output
+    std::string encoded = bech32::Encode(bech32::Encoding::BECH32, out_hrp, out_data);
+
+    if (encoded.empty() || encoded.size() > max_out_size) {
```
morehouse: I don't think `encoded` can be empty here.

main.cpp
```diff
@@ -38,8 +39,271 @@
...
+    overwrite_size = std::min(overwrite_size, in2_size);
+    size_t in2_begin = std::rand() % (in2_size - overwrite_size + 1);
+
+    std::memcpy(out, in1, in1_size);
+    std::memcpy(out + overwrite_begin, in2 + in2_begin, overwrite_size);
+
+    return in1_size;
+}
+
+size_t cross_over(const uint8_t *in1, size_t in1_size, const uint8_t *in2,
```
morehouse: This function can be `static`.  Or better yet, all these helpers can go in an unnamed namespace.

main.cpp
```diff
@@ -38,8 +39,271 @@
...
+                size_t in2_size, uint8_t *out, size_t max_out_size, unsigned seed)
```
morehouse: Nit:  these parameters could be C++-ified using `std::span` for inputs and `std::vector` for the output.

.github/workflows/workflow.yml
brunoerg: Because there is a known issue on nlightning and with the custom mutator, 100 runs were enough to catch it and cause a crash.

modules/bitcoin/bech32.h
brunoerg: Sure. I'm going to remove it.

.github/workflows/workflow.yml
brunoerg: I've dropped it. It was a workaround to test a thing.

main.cpp
brunoerg: I just removed this check.

main.cpp
brunoerg: Done.

main.cpp
brunoerg: Done.

main.cpp
brunoerg: Done. I removed for both.

main.cpp
brunoerg: Yes, you're right. I just removed it.

main.cpp
brunoerg: Sounds good, I removed it.

main.cpp
brunoerg: Sure, I just removed this check.

main.cpp
brunoerg: I've changed cross_over and simplified this part a little bit.

main.cpp
brunoerg: Done.

main.cpp
brunoerg: Just made it `static` but I intend to move it to the `bolt11mutator` module.

main.cpp
brunoerg: I just removed this check.

main.cpp
brunoerg: Sure, it's my goal. I will do it in a follow-up since I've addressed the other points and I'm running out of time for now.

brunoerg:
Rebased

main.cpp
brunoerg: I removed the conversion and it's now using `std::span` for the cross_over.

main.cpp
```diff
@@ -38,8 +40,290 @@
...
+        for (const char& c : hrp) {
+            if (c >= 'A' && c <= 'Z') {
+                return initial_input(fuzz_data, max_size);
+            }
+        }
+    }
+
+    if (data.empty() || data.size() > 65) {
+        return initial_input(fuzz_data, max_size);
+    }
```
morehouse: We shouldn't enforce the 65 limit, and I don't see a strong need to require `data` to be non-empty either.

```suggestion
```

main.cpp
```diff
@@ -38,8 +40,290 @@
...
+
+    // Copy in1 to out first (limited by max_out_size)
+    size_t in1_bytes = std::min(in1.size(), max_out_size);
+    out.insert(out.end(), in1.begin(), in1.begin() + in1_bytes);
+
+    // Choose overwrite position
+    if (!out.empty()) {
+        size_t pos = std::rand() % out.size();
+        size_t overwrite_length = std::min(in2.size(), out.size() - pos);
```
morehouse: This will *always* overwrite the max amount from `in2`.  We still want some randomness here so that a substring of `in2` could be overwritten instead.

main.cpp
```diff
@@ -38,8 +40,290 @@
...
+
+        // Overwrite portion of out with in2
+        std::copy_n(in2.begin(), overwrite_length, out.begin() + pos);
```
morehouse: This will always start overwriting with the beginning of `in2`, which is also undesired.  We want some random substring of `in2` to be used.

main.cpp
```diff
@@ -38,8 +40,290 @@
...
+
+    // Insert first part of in1
+    size_t bytes_to_copy = std::min(pos, max_out_size);
+    out.insert(out.end(), in1.begin(), in1.begin() + bytes_to_copy);
+
+    // Insert in2
+    if (out.size() < max_out_size) {
+        size_t remaining = max_out_size - out.size();
+        size_t in2_bytes = std::min(in2.size(), remaining);
+        out.insert(out.end(), in2.begin(), in2.begin() + in2_bytes);
```
morehouse: This *always* inserts the maximum substring of `in2`.  We want to insert some random substring instead.

main.cpp
```diff
@@ -38,8 +40,290 @@
...
+    std::srand(seed);
+    std::string out_hrp;
+    std::vector<uint8_t> out_data;
+
+    if (std::rand() % 2) {
+        // Cross-over the HRP
+        out_data = data1;
+
+        std::span<const uint8_t> hrp1_span(reinterpret_cast<const uint8_t*>(hrp1.data()), hrp1.size());
+        std::span<const uint8_t> hrp2_span(reinterpret_cast<const uint8_t*>(hrp2.data()), hrp2.size());
```
morehouse, line 285: This is uglier than I thought.  I was hoping we could just pass the string directly to `cross_over` and it would get implicitly converted to a span.

But I guess the issue is that `char` doesn't convert implicitly to `uint8_t`?

morehouse (changes requested):
The modified `insert_part` and `overwrite_part` functions have behavior that is too restrictive, so they won't produce as many unique combinations as the original mutator from CLN.

main.cpp
brunoerg: Yes, it makes sense, I've removed it.

main.cpp
brunoerg: Just changed it to select a random substring.

main.cpp
brunoerg: Done.

main.cpp
brunoerg: Done.

main.cpp
```diff
@@ -38,8 +40,289 @@
...
+    if (in1.empty() || in2.empty()) {
+        return out;
+    }
+
+    // Copy in1 to out first (limited by max_out_size)
+    size_t in1_bytes = std::min(in1.size(), max_out_size);
+    out.insert(out.end(), in1.begin(), in1.begin() + in1_bytes);
+
+    // Choose overwrite position
+    if (!out.empty() && !in2.empty()) {
```
morehouse: At this point `in2` can't be empty, or we would have returned above.

And we could reduce nesting by inverting the condition:

```c++
if (out.empty()) return out;
```

main.cpp
```diff
@@ -38,8 +40,289 @@
...
+        size_t pos = std::rand() % out.size();
+        size_t max_possible = std::min(in2.size(), out.size() - pos);
+        size_t overwrite_length = max_possible > 0 ? (std::rand() % max_possible) + 1 : 0;
```
morehouse: AFAICT `max_possible` cannot be 0 here.

We already checked that `in2.size() > 0`, and `out.size() - pos` is always at least 1.

main.cpp
```diff
@@ -38,8 +40,289 @@
...
+
+        if (overwrite_length > 0) {
```
morehouse: Likewise, since we added 1 above, `overwrite_length` is always at least 1.

main.cpp
brunoerg: Done

main.cpp
brunoerg: Done

main.cpp
brunoerg: Done.

morehouse (approved):
Code LGTM, haven't tested.

PR #126, opened 2025-04-03
Fix: Add Lightning Core support to invoice_deserialization target (#96)

rajanarahul93:
### 🛠 Problem  
Previously, my PR **[#110](https://github.com/brunoerg/bitcoinfuzz/pull/110)** was closed because my approach did not align with the intent of **issue #96**.  
The goal of this issue is to implement a **new module** for **Core Lightning** to support BOLT11 invoice deserialization.  
In my earlier PR, I mistakenly introduced an interdependency between `nlightning` and `ldk`, which was not the intended approach.  

### 🔄 What I Did Differently  
- Followed the correct approach as suggested by the maintainers.  
- Ensured `nlightning` handles **its own deserialization logic** instead of relying on `ldk`.  
- Implemented the necessary functions for BOLT11 invoice parsing within the `nlightning` module.  
- Removed unnecessary dependencies to maintain the differential fuzzing concept.  

### 📷 Proof of Successful Compilation  
Screenshot 2025-04-03 222631

### 🔬 Testing  
- The module now compiles successfully.  
- Invoice deserialization works as expected without breaking other modules.  

@brunoerg

### DIFF
--- a/modules/nlightning/module.cpp
+++ b/modules/nlightning/module.cpp
@@ -34,11 +34,11 @@ namespace bitcoinfuzz
     {
         FreeStringFunc NLightning::freeString = nullptr;
         DecodeInvoiceFunc NLightning::decodeInvoice = nullptr;
+        DecodeInvoiceFunc NLightning::decodeLightningCoreInvoice = nullptr;  // Initialize new function pointer
         CleanupResources NLightning::cleanupResources = nullptr;
 
         NLightning::NLightning(void) : BaseModule("NLightning")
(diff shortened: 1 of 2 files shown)


brunoerg (commented):
NACK

Sorry, but this PR has nothing to do with adding core-lightning to this project. You're simply modifying nlightning which is a completely different project.

brunoerg:
This PR has nothing to do with adding core-lightning to this project.

PR #136, opened 2025-04-08
Update lnd to latest commit

erickcestari:
I think we should avoid using the `replace github.com/lightningnetwork/lnd => github.com/erickcestari/lnd v0.0.0-20250407140637-aaf0a19b06b0` directive and instead depend on `lnd` directly as a module, without relying on a fork.

However, when I tried using it directly, I ran into some errors. So for now, let's just update the fork to point to the latest commit:

https://github.com/lightningnetwork/lnd/commit/aaf0a19b06b035570c1c9e787288b154bcf0cb65

Then we can open a PR later to remove the fork and use the upstream lnd module directly.


brunoerg (approved):
ACK 71ab136995e8af8197e44e06f3c2efae4da5ce8a

morehouse:
@erickcestari: what changes are on your fork, and can they be submitted upstream?

Edit:  nevermind.  Just saw https://github.com/bitcoinfuzz/bitcoinfuzz/pull/137

PR #140, opened 2025-04-09
[Feature Enhancement] added bolt7 fuzz target

pingu-73:


pingu-73:
@brunoerg `miniscript_parse` tests are failing https://github.com/bitcoinfuzz/bitcoinfuzz/actions/runs/14360413569/job/40260166180?pr=140

brunoerg:
> @brunoerg `miniscript_parse` tests are failing https://github.com/bitcoinfuzz/bitcoinfuzz/actions/runs/14360413569/job/40260166180?pr=140

I'm aware, thanks. I will try to fix it by today.

brunoerg:
Can you please rebase?

.github/workflows/workflow.yml
```diff
@@ -166,4 +166,4 @@ jobs:
         run: |
           export CXXFLAGS="-LDK -LND"
           make
-          ASAN_OPTIONS=detect_leaks=1 FUZZ=bolt7_channel_announcement ./bitcoinfuzz -runs=100
\ No newline at end of file
+          ASAN_OPTIONS=detect_leaks=0 FUZZ=bolt7_channel_announcement ./bitcoinfuzz -runs=100
```
brunoerg: If you're going to disable `detect_leaks` so it's better to remove it instead of setting to 0.

.github/workflows/workflow.yml
```diff
@@ -162,3 +162,8 @@ jobs:
           make
           ASAN_OPTIONS=detect_leaks=0 FUZZ=psbt_parse ./bitcoinfuzz -runs=100
 
+      - name: Test - bolt7_channel_announcement
+        run: |
+          export CXXFLAGS="-LDK -LND"
```
morehouse, line 167: These flags look wrong.  Don't you get errors without the `-D` prefix?

modules/lnd/lnd_wrapper/wrapper.go
```diff
@@ -73,4 +76,36 @@ func LndDeserializeInvoice(cInvoiceStr *C.char) *C.char {
...
+	r := bytes.NewReader(dataSlice)
+
+	// decode as ChannelAnnouncement1 (legacy ver)
+	var msg1 lnwire.ChannelAnnouncement1
+	if err := msg1.Decode(r, 0); err != nil {
+		return C.CString("")
+	}
+
+	result := fmt.Sprintf(
+		"NODE1=%x;NODE2=%x;BITCOIN_KEY1=%x;BITCOIN_KEY2=%x;SHORT_CHAN_ID=%v;CHAIN_HASH=%x;FEATURES=%v;",
```
morehouse, line 97: What about the signatures?

modules/ldk/ldk_lib/src/lib.rs
```diff
@@ -101,4 +105,42 @@ pub extern "C" fn ldk_free_string(ptr: *mut c_char) {
...
+            let block = announcement.contents.short_channel_id >> 40;
+            let tx = (announcement.contents.short_channel_id >> 16) & 0xFFFFFF;
+            let output = announcement.contents.short_channel_id & 0xFFFF;
+            result.push_str(&format!("SHORT_CHAN_ID={}:{}:{};", block, tx, output));
+            
+            result.push_str(&format!("CHAIN_HASH={};", announcement.contents.chain_hash.to_string().to_lowercase()));
+            
+            // TODO!!
+            // can't do result.push_str(&format!("FEATURES={:?};", announcement.contents.features));
+            // so creating a representation that looks like FEATURES=&{map[1:{} 5:{} 9:{} ...]} from lnd
```
morehouse, line 139: I'd suggest a format like this: `FEATURES=1,5,9,...;`, with each wrapper manipulating the parsed features into this format.

driver.cpp
```diff
@@ -202,6 +202,57 @@ namespace bitcoinfuzz
...
+                remove_features(res_no_features);
+                remove_features(last_response_no_features);
+
+                if (*res != *last_response)
+                {
+                    std::cout << "BOLT7 announcement mismatch between " << last_module_name  << " and " << module.first << "!" << "\n";
+                    std::cout << "  " << last_module_name << ": " << last_response_for_logging << "\n";
+                    std::cout << "  " << module.first << ": " << res_for_logging << "\n";
+                }
+                assert(res_no_features == last_response_no_features);
```
morehouse, line 249: If we aren't comparing features, we should just omit them from the response strings, so the workaround here can be removed.

But we should be able to compare features by using a consistent string format.

morehouse (commented):
There's dozens of lightning messages that could be fuzzed this way.  Why did you choose `channel_announcement` messages in particular?

While I think more fuzzing is good, I'm a bit skeptical that we'll find many bugs in parsing these messages, since the parsing logic is fairly simple compared to things like bolt11 or bolt12 invoices.

Maybe more bugs could be found by comparing the *validation* results of the announcements, for example using LND's `netann.ValidateChannelAnn`.  Of course all naively-generated announcements will be invalid due to incorrect signatures, but we could write a custom mutator that creates valid signatures sometimes.

PR #141, opened 2025-04-09
Remove payment secret workaround

### LINKED ISSUE #138: bolt11: remove payment secret workaround
Previously some lightning implementations required payment secrets in all invoices while others did not.  But now that https://github.com/lightning/bolts/pull/1242 has been merged to the spec, all implementations should start requiring payment secrets.

We should make sure all implementations update to comply with the latest spec change, and then remove our workaround here:

https://github.com/brunoerg/bitcoinfuzz/blob/e8ef846811381c9326f8a9b9335a112caab4d505/modules/ldk/ldk_lib/src/lib.rs#L75-L80

erickcestari:
Now that the payment secret is needed by both the writer and the reader, we should remove the ldk workaround

Closes #138

### DIFF
--- a/modules/ldk/ldk_lib/src/lib.rs
+++ b/modules/ldk/ldk_lib/src/lib.rs
@@ -72,12 +72,6 @@ pub unsafe extern "C" fn ldk_des_invoice(input: *const std::os::raw::c_char) -> 
 
             str_to_c_string(&result)
         }
-        // Handle invoices without payment secrets by returning null
-        // This is needed because some Lightning implementations don't require payment secrets,
-        // and we need to maintain compatibility with these implementations
-        Err(ParseOrSemanticError::SemanticError(Bolt11SemanticError::NoPaymentSecret)) => {
-            std::ptr::null_mut()
-        }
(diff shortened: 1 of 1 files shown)


brunoerg (approved):
code review ACK 1a110b0771050e7ec617b145fe16152e516c4625

morehouse:
I think we need to wait on this until CLN, LND (and others?) also reject such invoices.

Otherwise the fuzz test will be constantly failing.

brunoerg:
> I think we need to wait on this until CLN, LND (and others?) also reject such invoices.

What is the current state of it?

erickcestari:
> > I think we need to wait on this until CLN, LND (and others?) also reject such invoices.
> 
> What is the current state of it?

The LND implementation will be completed in two phases: first, the initial phase must be merged, and only then will LND start enforcing the payment secret requirement in the second phase

https://github.com/lightningnetwork/lnd/pull/9752

The CLN is waiting for the new changes on bolts.
https://github.com/ElementsProject/lightning/pull/8221

brunoerg:
Update: CLN has removed it, now we're just waiting for LND.

PR #156, opened 2025-04-30
ci: Add macos job

### LINKED ISSUE #154: ci: add macOS job
We currently only build and run bitcoinfuzz on Ubuntu, it would be great if we had a macOS step. For example, https://github.com/bitcoinfuzz/bitcoinfuzz/pull/153 works fine on Linux and CI passes, however, the Makefile has some commands that do not work on macOS that could be caught by CI.

erickcestari:
- Added a GitHub Actions job to run CI on macos-13 alongside ubuntu-24.04.
- Updated from ubuntu-latest to ubuntu-24.04 for more consistent builds.
- Added conditional steps to install LLVM, Clang, Boost, and other build dependencies specific to each OS.
- Adjusted environment setup and build commands to support both platforms.

Closes #154

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-24.04]
 
     steps:
       - name: Checkout repo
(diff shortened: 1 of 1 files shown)


brunoerg (approved):
ACK 028e9d51a537e73f70c5d7102997ddd6d48d46cb

PR #157, opened 2025-05-02
instrument rust modules with SanitizerCoverage

### LINKED ISSUE #151: Instrument rust modules with SanitizerCoverage
As far as I can tell, none of the rust modules use [SanitizerCoverage](https://clang.llvm.org/docs/SanitizerCoverage.html) instrumentation, which means libFuzzer is blind when fuzzing them.  We can still find bugs this way, as long as the non-Rust modules we're fuzzing against have proper instrumentation.  But to maximize bug-finding power we should instrument Rust modules too.

It looks like [cargo-fuzz](https://github.com/rust-fuzz/cargo-fuzz) instruments rust programs with the necessary SanitizerCoverage [flags](https://github.com/rust-fuzz/cargo-fuzz/blob/main/src/project.rs#L175-L179), so probably there's some simple way to use cargo-fuzz for this purpose.  Otherwise, it's also possible to add this instrumentation manually, but I'm not sure how tricky that will be.

brunoerg:
Fixes #151

### DIFF
--- a/modules/ldk/Makefile
+++ b/modules/ldk/Makefile
@@ -5,7 +5,12 @@ CXXFLAGS += -Wall -Wextra -fsanitize=address,fuzzer -std=c++20 -I ../../include
 LDK_LIB_PATH := ./ldk_lib/target/release/libldk_lib.a
 
 cargo:
-	cd ldk_lib && RUSTFLAGS="-Z sanitizer=address" cargo build --release
+	cd ldk_lib && RUSTFLAGS="-Z sanitizer=address" cargo rustc --release -- -C passes='sancov-module' \
+    -C llvm-args=-sanitizer-coverage-inline-8bit-counters \
+    -C llvm-args=-sanitizer-coverage-trace-compares \
+    -C llvm-args=-sanitizer-coverage-pc-table \
+    -C llvm-args=-sanitizer-coverage-level=4 \
+    -C llvm-args=-simplifycfg-branch-fold-threshold=0
 
 module.a: module.o $(LDK_LIB_PATH)
 	bash ../../merge.sh module.a $(LDK_LIB_PATH) module.o
--- a/modules/rustbitcoin/Makefile
+++ b/modules/rustbitcoin/Makefile
@@ -6,7 +6,12 @@ CXXFLAGS += -Wall -Wextra -fsanitize=address,fuzzer -std=c++20 -I ../../include
 RUST_LIB_PATH := ./rust_bitcoin_lib/target/release/librust_bitcoin_lib.a
 
 cargo:
-	cd rust_bitcoin_lib && RUSTFLAGS="-Z sanitizer=address" cargo build --release
+	cd rust_bitcoin_lib && RUSTFLAGS="-Z sanitizer=address" cargo rustc --release -- -C passes='sancov-module' \
+    -C llvm-args=-sanitizer-coverage-inline-8bit-counters \
+    -C llvm-args=-sanitizer-coverage-trace-compares \
+    -C llvm-args=-sanitizer-coverage-pc-table \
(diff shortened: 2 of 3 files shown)


brunoerg:
I ran `FUZZ=psbt_parse ./bitcoinfuzz -runs=10000000` with only rust-bitcoin set and got:

This PR:
```
#10000000	DONE   cov: 540 ft: 705 corp: 35/13793b lim: 4096 exec/s: 90909 rss: 461Mb
```

Master:
```
#10000000	DONE   cov: 341 ft: 341 corp: 1/1b lim: 4096 exec/s: 140845 rss: 463Mb
```

Coverage is being tracked now.

brunoerg:
cc @morehouse @apoelstra

morehouse:
A couple flags that `cargo fuzz` uses and might provide more benefit for us:
- `-C llvm-args=-simplifycfg-branch-fold-threshold=0`: this disables an optimization that reduces coverage feedback
- `-C llvm-args=-sanitizer-coverage-level=4`: same as level 3 but also instruments indirect calls

brunoerg:
> A couple flags that `cargo fuzz` uses and might provide more benefit for us:
> 
> * `-C llvm-args=-simplifycfg-branch-fold-threshold=0`: this disables an optimization that reduces coverage feedback
> * `-C llvm-args=-sanitizer-coverage-level=4`: same as level 3 but also instruments indirect calls

Interesting, I just addressed them.

modules/ldk/Makefile
```diff
@@ -5,7 +5,12 @@ CXXFLAGS += -Wall -Wextra -fsanitize=address,fuzzer -std=c++20 -I ../../include
 LDK_LIB_PATH := ./ldk_lib/target/release/libldk_lib.a
 
 cargo:
-	cd ldk_lib && RUSTFLAGS="-Z sanitizer=address" cargo build --release
+	cd ldk_lib && RUSTFLAGS="-Z sanitizer=address" cargo rustc --release -- -C passes='sancov-module' \
+    -C llvm-args='-sanitizer-coverage-inline-8bit-counters' \
+    -C llvm-args='-sanitizer-coverage-trace-compares' \
+    -C llvm-args='-sanitizer-coverage-pc-table' \
+    -C llvm-args='-sanitizer-coverage-level=4' \
+    -C llvm-args=-simplifycfg-branch-fold-threshold=0
```
morehouse: Nit: inconsistent use of quotes

```suggestion
    -C llvm-args=-sanitizer-coverage-inline-8bit-counters \
    -C llvm-args=-sanitizer-coverage-trace-compares \
    -C llvm-args=-sanitizer-coverage-pc-table \
    -C llvm-args=-sanitizer-coverage-level=4 \
    -C llvm-args=-simplifycfg-branch-fold-threshold=0
```

or

```suggestion
    -C llvm-args='-sanitizer-coverage-inline-8bit-counters' \
    -C llvm-args='-sanitizer-coverage-trace-compares' \
    -C llvm-args='-sanitizer-coverage-pc-table' \
    -C llvm-args='-sanitizer-coverage-level=4' \
    -C llvm-args='-simplifycfg-branch-fold-threshold=0'
```

morehouse (approved):

modules/ldk/Makefile
brunoerg: Done. Thanks.

apoelstra:
concept ACK supporting this but I continue to think it should be possible to disable for purposes of quickly building a high-coverage corpus.

brunoerg:
> concept ACK supporting this but I continue to think it should be possible to disable for purposes of quickly building a high-coverage corpus.

I'm going to work on a way to enable/disable it in follow-up.

PR #166, opened 2025-05-09
Add Target for BOLT12 offer decode

### LINKED ISSUE #162: Add Target for BOLT12 Offer Decode
Add a new target to our differential fuzzing  to test BOLT12 offer decoding functionality across multiple Bitcoin implementations.

erickcestari:
This PR adds a new fuzzing target for BOLT12 offer decoding to our differential fuzzing. It implements the `deserialize_offer` target and adds support for it in the core-lightning and LDK modules.

Closes #162

### DIFF
--- a/modules/clightning/module.cpp
+++ b/modules/clightning/module.cpp
@@ -2,6 +2,7 @@
 
 extern "C" {
     #include "common/bolt11.h"
+    #include "common/bolt12.h"
     #include "bitcoin/pubkey.h"
     #include "common/node_id.h"
     #include "common/utils.h"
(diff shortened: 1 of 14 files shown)


brunoerg:
Needs rebase.

modules/clightning/module.cpp
```diff
@@ -80,6 +81,96 @@ std::string clightning_des_invoice(const std::string& input) {
     return result.str();
 }
 
+std::string clightning_des_offer(const std::string& input) {
+    char* fail = nullptr;
+    const struct chainparams* params = chainparams_for_network("bitcoin");
+    const char *b12 = input.c_str();
+    size_t b12len = input.size();
```
morehouse: Nit:  Since we're passing the buffer size, we shouldn't need to add the null terminator with `c_str()`.

And since these variables are only used once, I think we should just use `input.data()` and `input.size()` directly below.

modules/clightning/module.cpp
```diff
@@ -92,5 +183,10 @@ namespace bitcoinfuzz
         {
             return clightning_des_invoice(str.c_str());
         }
+
+        std::optional<std::string> CLightning::deserialize_offer(std::string str) const
+        {
+            return clightning_des_offer(str.c_str());
```
morehouse: `c_str()` looks unnecessary here and probably causes 1 or 2 extra copies of `str` to be made  (first to add null byte, then again to convert back to `std::string`).
```suggestion
            return clightning_des_offer(str);
```

modules/clightning/module.cpp
```diff
@@ -80,6 +81,96 @@ std::string clightning_des_invoice(const std::string& input) {
     return result.str();
 }
 
+std::string clightning_des_offer(const std::string& input) {
```
morehouse: Nit: could use a `std::string_view` here

modules/ldk/ldk_lib/src/lib.rs
```diff
@@ -94,11 +96,105 @@ pub unsafe extern "C" fn ldk_des_invoice(input: *const std::os::raw::c_char) ->
...
+            be_flags.reverse();
+            result.push_str(be_flags.to_hex_string(Case::Lower).as_str());
+
+            result.push_str(";ABSOLUTE_EXPIRY=");
+            if let Some(absolute_expiry) = offer.absolute_expiry() {
+                result.push_str(absolute_expiry.as_secs().to_string().as_str());
+            }
+
+            offer.paths().iter().for_each(|path| {
+                path.blinded_hops().iter().for_each(|hop| {
```
morehouse, line 164: Note that we're smushing all the hops together and losing the path delineation in our output.

I doubt we'll miss bugs this way, but it's theoretically possible.  Anyhow we can always fix that in the future if we want.

morehouse (approved):
Code LGTM.  Haven't tested

modules/clightning/module.cpp
erickcestari: Fixed

modules/clightning/module.cpp
erickcestari: Fixed

modules/ldk/ldk_lib/src/lib.rs
erickcestari: I changed to:

```rust
offer.paths().iter().enumerate().for_each(|(i, path)| {
    path.blinded_hops().iter().for_each(|hop| {
        result.push_str(&format!(";PATH_{}_HOP=", i));
        result.push_str(hop.blinded_node_id.to_string().as_str());
    });
});
```

modules/clightning/module.cpp
erickcestari: Fixed

morehouse (approved):

brunoerg (approved):
tested ACK f04b7f28898e399ad1e374febd2ce89f071da6b2

PR #169, opened 2025-05-13
fix: handle amount encoding for compatibility with non-LND implementations

erickcestari:
When integrating with multiple Lightning implementations, we observed compatibility issues when dealing with large Bitcoin amounts. This is due to differences in how amounts are represented internally:

- LND uses milliSatoshis for internal representation
- LDK and other implementations use picoSatoshis, which can overflow when dealing with extremely large amounts

### DIFF
--- a/modules/lnd/lnd_wrapper/wrapper.go
+++ b/modules/lnd/lnd_wrapper/wrapper.go
@@ -42,7 +42,16 @@ func LndDeserializeInvoice(cInvoiceStr *C.char) *C.char {
 
 	sb.WriteString(";AMOUNT=")
 	if invoice.MilliSat != nil {
-		sb.WriteString(fmt.Sprintf("%d", *invoice.MilliSat))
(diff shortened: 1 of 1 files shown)


brunoerg:
Cool. 

CI failure is unrelated.

brunoerg (approved):
ACK 4a5ce53ff298480c1a18d1520219849069e32df7

morehouse:
Thinking about this some more... Does it always work?

We observed the amount field differing only in the top bit for one of the test inputs, but if there's actually a 10^9 difference in the stored values, we probably need to mask out a lot more than the top bit.

erickcestari:
> Thinking about this some more... Does it always work?
> 
> We observed the amount field differing only in the top bit for one of the test inputs, but if there's actually a 10^9 difference in the stored values, we probably need to mask out a lot more than the top bit.

Yes, you're right. The solution could be to multiply and then divide LND’s milliSatoshis by 10, simulating the same behavior as other implementations that use picoBitcoins.

PR #177, opened 2025-05-15
simulate expiry overflow to ensure consistent behavior

erickcestari:


morehouse:
If we move forward with this, we should add comments explaining why we do this in each module.  Also would be good to open an issue to remove the workaround if/when https://github.com/lightningnetwork/lnd/issues/9808 is fixed.

morehouse (approved):
LGTM

PR #183, opened 2025-05-22
fix: Fix Rust FFI string truncation on embedded null bytes

erickcestari:
Replace null-terminated string handling with length-based approach in Rust FFI functions to prevent data loss when input contains null bytes.

- Add length parameter to `ldk_des_invoice`, `ldk_des_offer`, and `rust_bitcoin_address_parse`
- Use `slice::from_raw_parts` instead of `CStr::from_ptr` in Rust
- Pass `str.data()` and `str.length()` from C++ instead of `str.c_str()`

### DIFF
--- a/modules/ldk/ldk_lib/ldk_lib.h
+++ b/modules/ldk/ldk_lib/ldk_lib.h
@@ -1,7 +1,7 @@
 #include <cstdint>
 
-extern "C" char* ldk_des_invoice(const char* input);
+extern "C" char* ldk_des_invoice(const char* input, size_t length);
 
-extern "C" char* ldk_des_offer(const char* input);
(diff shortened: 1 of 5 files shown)


erickcestari:
After reviewing the code more thoroughly. I noticed that this doesn't make sence since this project tests multiple implementations via FFI to ensure they handle identical inputs the same way. The null termination "issue" is actually correct behavior - all implementations should receive the same truncated input at null bytes.
"Fixing" only the Rust implementation would mean:

- Rust gets: "lno\x00ano1zc..." (full string)
- Other implementations get: "lno" (truncated)

This defeats the purpose of differential testing since implementations would be processing different data.

Closing the pr...

PR #184, opened 2025-05-22
fix: Handle string length properly in c-lightning offer_decode

erickcestari:
Use strnlen() to get truncated string length instead of full buffer size to match C string semantics

### DIFF
--- a/modules/clightning/module.cpp
+++ b/modules/clightning/module.cpp
@@ -85,7 +85,10 @@ std::string clightning_des_offer(const std::string_view input) {
     char* fail = nullptr;
     const struct chainparams* params = chainparams_for_network("bitcoin");
 
-    struct tlv_offer *offer = offer_decode(tmpctx, input.data(), input.size(), nullptr, params, &fail);
+    // Get the truncated length of the string (in case it contains null bytes)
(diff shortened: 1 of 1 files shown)


modules/clightning/module.cpp
```diff
@@ -85,7 +85,10 @@ std::string clightning_des_offer(const std::string_view input) {
     char* fail = nullptr;
     const struct chainparams* params = chainparams_for_network("bitcoin");
 
-    struct tlv_offer *offer = offer_decode(tmpctx, input.data(), input.size(), nullptr, params, &fail);
+    // Get the truncated length of the string (in case it contains null bytes)
+    size_t c_string_len = strnlen(input.data(), input.size());
```
brunoerg, line 89: `strnlen` finds the actual string length up to the first null byte, right?

erickcestari: Yes, it is written in the docs:
```
Find the length of STRING, but scan at most MAXLEN characters.
If no '\0' terminator is found in that many characters, return MAXLEN.
```

brunoerg (approved):
code review ACK 1f355ab74ef22b6fd90817803f0e5672be85f05a

Issue #181, opened 2025-05-17
add psbt target for bitcoin-core


yuvicc:

brunoerg:
Concept ACK. It currently supports btcd and rust-bitcoin.

PR #186, opened 2025-05-23
add psbt target in bitcoin core

yuvicc:
Adresses #181

### DIFF
--- a/modules/bitcoin/module.cpp
+++ b/modules/bitcoin/module.cpp
@@ -16,6 +16,8 @@
 #include "validation.h"
 #include "core_io.h"
 #include "key_io.h"
+#include "psbt.h"
+#include "span.h"
 
 namespace {
 class FuzzedSignatureChecker : public BaseSignatureChecker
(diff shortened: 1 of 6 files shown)


modules/bitcoin/module.cpp
```diff
@@ -366,5 +368,88 @@ std::optional<std::string> Bitcoin::addrv2_parse(std::span<const uint8_t> buffer
...
+
+    if (buffer.empty()) {
+        return std::nullopt;
+    }
+
+    PartiallySignedTransaction psbt;
+    std::string error;
+
+    // Attempt to decode raw psbt from buffer
+    if(!DecodeRawPSBT(psbt, MakeByteSpan(buffer), error)) {
```
brunoerg:
```suggestion
    if (!DecodeRawPSBT(psbt, MakeByteSpan(buffer), error)) {
```

modules/bitcoin/module.cpp
```diff
@@ -366,5 +368,88 @@ std::optional<std::string> Bitcoin::addrv2_parse(std::span<const uint8_t> buffer
...
+        return std::nullopt;
```
brunoerg: If Bitcoin Core could not decode a PSBT or the transaction is not valid we should not return a nullopt. It should return something that indicates that it fails (e.g. an empty string). We only use `std::nullopt` to sinalize that a module does not support something.

yuvicc: Correct, I'll fix that!

yuvicc: Done!

modules/bitcoin/module.cpp
```diff
@@ -366,5 +368,87 @@ std::optional<std::string> Bitcoin::addrv2_parse(std::span<const uint8_t> buffer
...
+    if (!DecodeRawPSBT(psbt, MakeByteSpan(buffer), error)) {
+        return std::string{};
+    }
+
+    std::string result;
+
+    try {
+        // Check if it's a valid transaction
+        if (!psbt.tx) {
+            return std::nullopt;
```
brunoerg: Same here for not returning a std::nullopt.

modules/bitcoin/module.cpp
```diff
@@ -366,5 +368,87 @@ std::optional<std::string> Bitcoin::addrv2_parse(std::span<const uint8_t> buffer
...
+                         std::to_string(static_cast<int64_t>(txout.nValue)) + ";";
+                
+                // Output script as hex string
+                result += "out" + std::to_string(i) + "script=" + 
+                         HexStr(txout.scriptPubKey) + ";";
+            }
+        }
+        
+    } catch (const std::exception& e) {
+        return std::nullopt;
```
brunoerg:
```suggestion
        return "";
```

modules/bitcoin/module.cpp
yuvicc: Ah, Got it!

brunoerg (approved):
ACK fcedb19faf2ed18e84597f6a10cf92065cccb79e; successfully ran it locally.

PR #188, opened 2025-05-28
Improve bitcoin build

erickcestari:
* Add `cmake_minimum_required(VERSION 3.16)` to be able to build with CMAKE 4 version
* Set default BOOST_LIB_DIR to Linux and macOS systems.


brunoerg (changes requested):
Since this sets default BOOST_LIB_DIR based on OS, we could remove the BOOST_LIB_DIR instructions from README.

brunoerg (approved):
ACK 56521c7e210c2811786799b8c50942c040215e09

PR #189, opened 2025-05-28
Fix Python error handling in Embit module

erickcestari:
### Changes
- **Refactor**: Consolidated `embit_lib.cpp` into `module.cpp` file
- **Fix**: Clear Python errors before GIL release to prevent persistent error state

### Problem
The Embit module was experiencing null main_module errors during fuzzing due to Python errors not being cleared between function calls, causing import failures.


brunoerg (approved):
ACK 8f48eee81d319e2b9a1b7477f5b8f0edf971cbe6

PR #190, opened 2025-05-30
fix: iterate over offer_chains and output all chain hashes

erickcestari:
### Changes

* Iterate over `offer_chains` and include all chains in the output
* Separate each chain with a semicolon (`;`)


brunoerg (approved):
code review ACK e23ccf2fc935d279e62fbb42372a4d49b54141a8

PR #193, opened 2025-05-30
fix: use tal_count() instead of NULL termination

erickcestari:
Arrays from talloc are not guaranteed NULL-terminated. Under memory pressure, this caused segfaults when iterating past array bounds.


brunoerg (approved):
code review ACK 2dda58c74038532f82fb0c65d4beb30ac0b965df

PR #205, opened 2025-06-12
Use core-lightning's BOLT12 custom mutator to enhance fuzz testing

### LINKED ISSUE #165: bolt12: Use core-lightning's BOLT12 custom mutator to enhance fuzz testing
https://github.com/ElementsProject/lightning/blob/master/tests/fuzz/bolt12.h

erickcestari:
The custom mutator decodes bech32 inputs, mutates the decoded data, then re-encodes to ensure valid bech32 strings reach the fuzzer target.

## Changes

- Added BOLT12 offer fuzzing with custom mutator/crossover for `lno` bech32 strings
- Refactored `bolt11mutator` → `custommutator` module for reusability
- Added `EncodeNoChecksum`/`DecodeNoChecksum` functions to bech32 library
- Extracted crossover logic into separate `customcrossover.h/cpp `files
- Updated CI to test BOLT12 offer deserialization using the `-DCUSTOM_MUTATOR_BOLT12_OFFER`

Closes: #165


main.cpp
```diff
@@ -41,12 +41,207 @@
...
 
 
+#if defined(CUSTOM_MUTATOR_BOLT11) || defined(CUSTOM_MUTATOR_BOLT12_OFFER)
+#include <modules/custommutator/bech32.h>
+#include <modules/custommutator/customcrossover.h>
+#endif
+
+#ifdef CUSTOM_MUTATOR_BOLT12_OFFER
+extern const std::string bech32_hrp = "lno";
+extern const std::string dummy_initial_input = "lno1zcss9mk8y3wkklfvevcrszlmu23kfrxh49px20665dqwmn4p72pksese";
```
erickcestari: I was unsure whether we should start with an offer that includes almost all fields to maximize coverage early on, or keep it minimal. I think using a fully populated offer from the beginning might cause the fuzzer to miss some bugs.

brunoerg: >  I think using a fully populated offer from the beginning might cause the fuzzer to miss some bugs.

Why?

erickcestari: Starting with a minimal offer lets the fuzzer discover bugs in edge cases where fields are missing or added incrementally. Many parsing bugs only surface when handling incomplete/minimal inputs - if we start with everything populated, the fuzzer mainly tests deletion mutations and might miss bugs that occur when building up from minimal valid offers.

main.cpp
```diff
@@ -46,8 +46,202 @@
...
 std::shared_ptr<bitcoinfuzz::Driver> driver = nullptr;
 
+#ifdef CUSTOM_MUTATOR_BOLT12_OFFER
+// A custom mutator that decodes the bech32 input, mutates the decoded input,
+// and then re-encodes the mutated input. This produces an input corpus that
+// consists entirely of correctly encoded bech32 strings, enabling efficient
+// fuzzing of the bolt12 decoding logic without the fuzzer getting stuck on
+// fuzzing the bech32 decoding logic. This custom mutator is originally from
+// core-lightning:
+// https://github.com/ElementsProject/lightning/blob/master/tests/fuzz/bolt12.h#L59
```
brunoerg: You can link it to a stable version instead of master branch.

erickcestari: Fixed

modules/custommutator/bech32.cpp
```diff
@@ -370,6 +370,21 @@ std::string Encode(Encoding encoding, const std::string& hrp, const data& values
...
 
+/** Encode a Bech32 or Bech32m string without the checksum. */
+std::string EncodeNoChecksum(const std::string& hrp, const data& values) {
+    // First ensure that the HRP is all lowercase. BIP-173 and BIP350 require an encoder
+    // to return a lowercase Bech32/Bech32m string, but if given an uppercase HRP, the
+    // result will always be invalid.
+    for (const char& c : hrp) assert(c < 'A' || c > 'Z');
+
+    std::string ret;
+    ret.reserve(hrp.size() + 1 + values.size() + CHECKSUM_SIZE);
```
morehouse: There's extra bytes here we don't need.
```suggestion
    ret.reserve(hrp.size() + 1 + values.size());
```

modules/custommutator/bech32.cpp
```diff
@@ -399,6 +414,33 @@ DecodeResult Decode(const std::string& str, CharLimit limit) {
...
+            return {};
+        }
+        values[i] = rev;
+    }
+    std::string hrp;
+    hrp.reserve(pos);
+    for (size_t i = 0; i < pos; ++i) {
+        hrp += LowerCase(str[i]);
+    }
+    return {Encoding::BECH32, std::move(hrp), data(values.begin(), values.end() - CHECKSUM_SIZE)};
```
morehouse: Yikes, doesn't this truncate the last `CHECKSUM_SIZE` bytes?

modules/custommutator/bech32.h
```diff
@@ -44,6 +44,8 @@ enum CharLimit : size_t {
  *  assertion error. Encoding must be one of BECH32 or BECH32M. */
 std::string Encode(Encoding encoding, const std::string& hrp, const std::vector<uint8_t>& values);
 
+std::string EncodeNoChecksum(const std::string& hrp, const std::vector<uint8_t>& values);
```
morehouse: Nit: function comment is missing

modules/custommutator/customcrossover.cpp
```diff
@@ -0,0 +1,76 @@
+#include "customcrossover.h"
+
+ std::vector<uint8_t> insert_part(std::span<const uint8_t> in1,
+                                      std::span<const uint8_t> in2,
+                                      size_t max_out_size)
```
morehouse: Nit: parameter indentation seems off in this file.

modules/custommutator/customcrossover.h
```diff
@@ -0,0 +1,26 @@
...
+// Overwrites a random part of in1 with a slice from in2, up to max_out_size.
+std::vector<uint8_t> overwrite_part(std::span<const uint8_t> in1,
+                                    std::span<const uint8_t> in2,
+                                    size_t max_out_size);
+
+// Randomly selects insert_part or overwrite_part using the provided seed.
+std::vector<uint8_t> cross_over(std::span<const uint8_t> in1,
+                                      std::span<const uint8_t> in2,
+                                      size_t max_out_size,
+                                      unsigned seed);
```
morehouse: Nit: indentation doesn't align for these parameters.

main.cpp
```diff
@@ -41,12 +41,207 @@
 #endif
 
 #ifdef CUSTOM_MUTATOR_BOLT11
-#include <modules/bolt11mutator/bech32.h>
+#include <modules/custommutator/bech32.h>
+#include <modules/custommutator/customcrossover.h>
 #endif
```
morehouse: Looks like this can be removed entirely since these files are already included below.

main.cpp
morehouse: As long as the corpus eventually has both inputs (i.e. a minimal one `lno1` and a more complete one `lno1zcss9mk8y3wkklfvevcrszlmu23kfrxh49px20665dqwmn4p72pksese`), we shouldn't need to worry about the differences in fuzzing, since inputs are selected from the corpus for mutation.  The initial input here just acts as a single seed input for the corpus.

My slight preference is to set the initial input here to a minimal `lno1` for simplicity, but either way should be fine.

morehouse: I don't think there's any value in using `extern` here.  Also these can be compile-time constants.
```suggestion
constexpr std::string_view bech32_hrp = "lno";
constexpr std::string_view dummy_initial_input = "lno1zcss9mk8y3wkklfvevcrszlmu23kfrxh49px20665dqwmn4p72pksese";
```

Also let's just move these definitions to the below ifdef instead?

In fact `dummy_initial_input` could be defined inside `initial_input`.

main.cpp
```diff
@@ -41,12 +41,207 @@
...
+// core-lightning:
+// https://github.com/ElementsProject/lightning/blob/release-v25.02.2/tests/fuzz/bolt12.h#L54
+extern "C" size_t LLVMFuzzerMutate(uint8_t *Data, size_t Size, size_t MaxSize);
+extern "C" size_t LLVMFuzzerCustomMutator(uint8_t *fuzz_data, size_t size, size_t max_size,
+                                          unsigned int seed);
+extern "C" size_t LLVMFuzzerCustomCrossOver(const uint8_t *in1, size_t in1_size, const uint8_t *in2,
+                                            size_t in2_size, uint8_t *out, size_t max_out_size,
+                                            unsigned seed);
+
+extern const std::string bech32_hrp;
```
morehouse: We define `bech32_hrp` in this file, so no need to declare it extern.
```suggestion
```

main.cpp
```diff
@@ -41,12 +41,207 @@
...
+        std::memcpy(fuzz_data, encoded_data.data(), encoded_size);
+
+        return encoded_size;
+    }
+    catch (const std::exception &e)
+    {
+        //std::cout << "Exception: " << e.what() << std::endl;
+        // In case of any exception, fall back to initial input
+        return initial_input(fuzz_data, max_size);
+    }
```
morehouse: How can an exception occur?

main.cpp
```diff
@@ -41,12 +41,207 @@
...
+ * and then re-encodes the mutated input. This produces an input corpus that
+ * consists entirely of correctly encoded bech32 strings, enabling efficient
+ * fuzzing of the bolt12 decoding logic without the fuzzer getting stuck on
+ * fuzzing the bech32 decoding logic. */
+extern "C" size_t LLVMFuzzerCustomMutator(uint8_t *fuzz_data, size_t size,
+                                          size_t max_size, unsigned int seed)
+{
+    try
+    {
+        if (size < 9)
```
morehouse: Why must `size` be at least 9?

main.cpp
```diff
@@ -41,12 +41,207 @@
...
+            return initial_input(fuzz_data, max_size);
+        std::string input_str(reinterpret_cast<char *>(fuzz_data), size);
+
+        // Check if input contains the correct HRP before decoding
+        if (input_str.length() < bech32_hrp.length() + 1 ||
+            input_str.substr(0, bech32_hrp.length()) != bech32_hrp ||
+            (input_str.length() > bech32_hrp.length() && input_str[bech32_hrp.length()] != '1'))
+        {
+            return initial_input(fuzz_data, max_size);
+        }
```
morehouse: Do we actually need to reject incorrect HRPs?

main.cpp
```diff
@@ -41,12 +41,207 @@
...
+
+        if (decoded_data.empty())
+        {
+            return initial_input(fuzz_data, max_size);
+        }
+
+        if (decoded_data.size() > max_size)
+        {
+            return initial_input(fuzz_data, max_size);
+        }
```
morehouse: Let's avoid a data copy:

```suggestion
        bech32::DecodeResult decoded_result = bech32::DecodeNoChecksum(input_str, bech32::CharLimit::CUSTOM_MUTATOR);
        std::vector<uint8_t> &decoded_data = decoded_result.data;

        if (decoded_data.empty())
        {
            return initial_input(fuzz_data, max_size);
        }

        if (decoded_data.size() > max_size)
        {
            return initial_input(fuzz_data, max_size);
        }
```

or alternatively:

```suggestion
        bech32::DecodeResult decoded = bech32::DecodeNoChecksum(input_str, bech32::CharLimit::CUSTOM_MUTATOR);

        if (decoded.data.empty())
        {
            return initial_input(fuzz_data, max_size);
        }

        if (decoded.data.size() > max_size)
        {
            return initial_input(fuzz_data, max_size);
        }
```

main.cpp
```diff
@@ -41,12 +41,207 @@
...
+        if (encoded_data.empty()) {
+            std::cout << "encoded_data.empty()";
+            return initial_input(fuzz_data, max_size);
+        }
+        size_t encoded_size = encoded_data.length();
+
+        if (encoded_size > max_size)
+        {
+            return initial_input(fuzz_data, max_size);
+        }
```
morehouse: Nit:  Let's use a consistent curly-brace style within this function.

main.cpp
```diff
@@ -41,12 +41,207 @@
...
+
+        // Mutate the data part of the decoded input
+        std::vector<uint8_t> mutated_data = decoded_data;
+
+        // Resize the vector to the maximum possible size BEFORE mutation.
+        mutated_data.resize(std::max(mutated_data.size(), max_size));
+
+        size_t mutated_size = LLVMFuzzerMutate(mutated_data.data(),
+                                            decoded_data.size(),
+                                            max_size);
```
morehouse: Nit: indentation seems off

main.cpp
```diff
@@ -41,12 +41,207 @@
...
+        {
+            return initial_input(fuzz_data, max_size);
+        }
+
+
+        // Mutate the data part of the decoded input
+        std::vector<uint8_t> mutated_data = decoded_data;
+
+        // Resize the vector to the maximum possible size BEFORE mutation.
+        mutated_data.resize(std::max(mutated_data.size(), max_size));
```
morehouse: Shouldn't we just truncate to `max_size` if necessary?

```suggestion
        mutated_data.resize(max_size));
```

main.cpp
```diff
@@ -41,12 +41,207 @@
...
+        }
+
+        if (decoded_data.size() > max_size)
+        {
+            return initial_input(fuzz_data, max_size);
+        }
+
+
+        // Mutate the data part of the decoded input
+        std::vector<uint8_t> mutated_data = decoded_data;
```
morehouse: Do we need to copy data *again*?  I think we can just mutate the decoded vector directly...

main.cpp
```diff
@@ -41,12 +41,207 @@
...
+        mutated_data.resize(mutated_size);
+
+        // It ensures that all values remain valid 5-bit values
+        for (uint8_t& val : mutated_data) {
+            val &= 0x1F;
+        }
+        
+        // Encode the mutated input
+        std::string encoded_data = bech32::EncodeNoChecksum(bech32_hrp, mutated_data);
+        if (encoded_data.empty()) {
```
morehouse: How would `encoded_data` be empty?

main.cpp
```diff
@@ -41,12 +41,207 @@
...
+            std::cout << "encoded_data.empty()";
+            return initial_input(fuzz_data, max_size);
+        }
+        size_t encoded_size = encoded_data.length();
```
morehouse: Nit:  seems like `encoded_size` is an unnecessary variable

main.cpp
```diff
@@ -41,12 +41,207 @@
...
+        // Decode first input
+        std::string input1_str(reinterpret_cast<const char*>(in1), in1_size);
+        
+        // Check if input1 contains the correct HRP
+        if (input1_str.length() < bech32_hrp.length() + 1 ||
+            input1_str.substr(0, bech32_hrp.length()) != bech32_hrp ||
+            (input1_str.length() > bech32_hrp.length() && 
+             input1_str[bech32_hrp.length()] != '1')) {
+            return 0;
+        }
```
morehouse: Do we actually need to reject incorrect HRPs?

main.cpp
```diff
@@ -41,12 +41,207 @@
...
+
+        bech32::DecodeResult decoded_result1 = bech32::DecodeNoChecksum(input1_str, 
+                                                    bech32::CharLimit::CUSTOM_MUTATOR);
```
morehouse: Nit: indentation seems off

main.cpp
```diff
@@ -41,12 +41,207 @@
...
+        std::vector<uint8_t> decoded_data1 = decoded_result1.data;
```
morehouse: Let's avoid this unnecessary copy

main.cpp
```diff
@@ -41,12 +41,207 @@
...
+            // std::cout << "// Check if input2 contains the correct HRP";
+            return 0;
+        }
+
+        bech32::DecodeResult decoded_result2 = bech32::DecodeNoChecksum(input2_str, 
+                                                    bech32::CharLimit::CUSTOM_MUTATOR);
+        std::vector<uint8_t> decoded_data2 = decoded_result2.data;
+        
+        if (decoded_data2.empty()) {
+            // std::cout << "decoded_data2.empty()";
```
morehouse: Please delete unused comments.

main.cpp
```diff
@@ -41,12 +41,207 @@
...
+            input2_str.substr(0, bech32_hrp.length()) != bech32_hrp ||
+            (input2_str.length() > bech32_hrp.length() && 
+             input2_str[bech32_hrp.length()] != '1')) {
+            // std::cout << "// Check if input2 contains the correct HRP";
+            return 0;
+        }
+
+        bech32::DecodeResult decoded_result2 = bech32::DecodeNoChecksum(input2_str, 
+                                                    bech32::CharLimit::CUSTOM_MUTATOR);
+        std::vector<uint8_t> decoded_data2 = decoded_result2.data;
```
morehouse: Unnecessary copy

main.cpp
```diff
@@ -41,12 +41,207 @@
...
+        }
+
+        // Perform cross-over on decoded data
+        std::vector<uint8_t> crossed_data = cross_over(decoded_data1, decoded_data2, 
+                                                       max_out_size, seed);
+        
+        if (crossed_data.empty()) {
+            // std::cout << "crossed_data.empty()";
+            return 0;
+        }
```
morehouse: Is this check necessary?  If the mutated data is empty, shouldn't the rest of the mutator still work properly?

main.cpp
```diff
@@ -41,12 +41,207 @@
...
+
+        // Encode the crossed-over data
+        std::string encoded_data = bech32::EncodeNoChecksum(bech32_hrp, crossed_data);
+        size_t encoded_size = encoded_data.length();
```
morehouse: Nit: `encoded_size` seems like an unnecessary variable

main.cpp
```diff
@@ -41,12 +41,207 @@
...
+        }
+
+        std::memcpy(out, encoded_data.data(), encoded_size);
+        return encoded_size;
+
+    } catch (const std::exception& e) {
+        //std::cout << "Exception: " << e.what() << std::endl;
+        // In case of any exception, fall back to cross_over_fail
+        return 0;
+    }
```
morehouse: How can an exception happen?

morehouse (changes requested):

modules/custommutator/bech32.cpp
erickcestari: Oh, yes. I will fix it.

main.cpp
erickcestari: The idea of defining it externally is to reuse this CUSTOM_MUTATOR_BOLT12 for other BOLT12 targets, such as invoice_request and invoice.

main.cpp
erickcestari: I was using it for debbuging. I will remove those comments/prints/catchs

main.cpp
erickcestari: Removed.

main.cpp
```diff
@@ -40,13 +40,155 @@
...
+{
+    std::string input_str(reinterpret_cast<char *>(fuzz_data), size);
+
+    // Check if input contains the correct HRP before decoding
+    if (input_str.length() < bech32_hrp.length() + 1 ||
+        input_str.substr(0, bech32_hrp.length()) != bech32_hrp ||
+        (input_str.length() > bech32_hrp.length() && input_str[bech32_hrp.length()] != '1'))
+    {
+        return initial_input(fuzz_data, max_size);
+    }
```
morehouse: What benefit do we get from preventing mutation of inputs with incorrect HRPs?

I don't see why we should do this -- below we're going to correct the HRP on re-encoding anyway.

main.cpp
```diff
@@ -40,13 +40,155 @@
...
+
+    size_t mutated_size = LLVMFuzzerMutate(decoded_data.data(),
+                                           decoded_data.size(),
+                                           max_size);
+    decoded_data.resize(mutated_size);
+
+    // It ensures that all values remain valid 5-bit values
+    for (uint8_t& val : decoded_data) {
+        val &= 0x1F;
+    }
```
morehouse, line 113: Nit:  In this function, there's currently two curly brace styles:
```c
if (...)
{
  ...
}
```

and 

```c
if (...) {
  ...
}
```

Let's pick one and be consistent.

main.cpp
```diff
@@ -40,13 +40,155 @@
...
+    {
+        return initial_input(fuzz_data, max_size);
+    }
+
+    // Resize the vector to the maximum possible size BEFORE mutation.
+    decoded_data.resize(max_size);
+
+    size_t mutated_size = LLVMFuzzerMutate(decoded_data.data(),
+                                           decoded_data.size(),
+                                           max_size);
```
morehouse: The second argument should be the size of `decoded_data` *before* we resized it.  Otherwise the mutation will treat the newly added null bytes as part of the original input.

main.cpp
```diff
@@ -40,13 +40,155 @@
...
+
+/* A custom mutator that decodes the bech32 input, mutates the decoded input,
+ * and then re-encodes the mutated input. This produces an input corpus that
+ * consists entirely of correctly encoded bech32 strings, enabling efficient
+ * fuzzing of the bolt12 decoding logic without the fuzzer getting stuck on
+ * fuzzing the bech32 decoding logic. */
+extern "C" size_t LLVMFuzzerCustomMutator(uint8_t *fuzz_data, size_t size,
+                                          size_t max_size, unsigned int seed)
+{
+    std::string input_str(reinterpret_cast<char *>(fuzz_data), size);
```
morehouse: I think `input_str` could be a `string_view`, and `DecodeNoChecksum` could take a `string_view` param, which would avoid another copy.

main.cpp
```diff
@@ -40,13 +40,155 @@
...
+    // Decode first input
+    std::string input1_str(reinterpret_cast<const char*>(in1), in1_size);
+    
+    // Check if input1 contains the correct HRP
+    if (input1_str.length() < bech32_hrp.length() + 1 ||
+        input1_str.substr(0, bech32_hrp.length()) != bech32_hrp ||
+        (input1_str.length() > bech32_hrp.length() && 
+            input1_str[bech32_hrp.length()] != '1')) {
+        return 0;
+    }
```
morehouse: Same comment as above -- I think we can remove the HRP checks.

main.cpp
```diff
@@ -40,13 +40,155 @@
...
+
+    bech32::DecodeResult decoded_result1 = bech32::DecodeNoChecksum(input1_str, bech32::CharLimit::CUSTOM_MUTATOR);
+
+    if (decoded_result1.data.empty()) {
+        return 0;
+    }
```
morehouse: I think we can omit the `empty` checks for simplicity -- `cross_over` should correctly handle the case where either vector is empty.

main.cpp
morehouse: Sure, but we don't do that yet, and the ifdefs are currently the same (which needs to be changed later anyway), so let's just combine them for now?

morehouse (changes requested):

main.cpp
erickcestari: Fixed!

main.cpp
erickcestari: Yes, it doesn't make sense. Fixed!

main.cpp
erickcestari: I picked this one, since it is the most used in our project:

```rust
if (...) {
  ...
}
```

main.cpp
erickcestari: Fixed!

main.cpp
erickcestari: Removed

main.cpp
erickcestari: Removed

main.cpp
erickcestari: Okay

modules/custommutator/bech32.cpp
```diff
@@ -284,7 +284,7 @@ inline unsigned char LowerCase(unsigned char c)
 }
 
 /** Return indices of invalid characters in a Bech32 string. */
-bool CheckCharacters(const std::string& str, std::vector<int>& errors)
+bool CheckCharacters(const std::string_view& str, std::vector<int>& errors)
```
morehouse: `string_view`s are lightweight and are best passed by value (here and elsewhere).

```suggestion
bool CheckCharacters(std::string_view str, std::vector<int>& errors)
```

main.cpp
```diff
@@ -40,13 +40,123 @@
...
+/* A custom cross-over mutator that decodes the bech32 inputs before cross-over
+ * mutating them. Like LLVMFuzzerCustomMutator, this enables more efficient
+ * fuzzing of bolt12 offers, invoice requests, and invoices. */
+extern "C" size_t LLVMFuzzerCustomCrossOver(const uint8_t *in1, size_t in1_size,
+                                            const uint8_t *in2, size_t in2_size,
+                                            uint8_t *out, size_t max_out_size,
+                                            unsigned seed)
+{
+    // Decode first input
+    std::string input1_str(reinterpret_cast<const char*>(in1), in1_size);
```
morehouse: `string_view` avoids a copy here and for `input2_str`.

main.cpp
```diff
@@ -40,13 +40,123 @@
...
+    if (decoded_data.size() > max_size) {
+        return initial_input(fuzz_data, max_size);
+    }
+
+    // Resize the vector to the maximum possible size BEFORE mutation.
+    decoded_data.resize(max_size);
+
+    size_t mutated_size = LLVMFuzzerMutate(decoded_data.data(),
+                                           decoded_data_size,
+                                           max_size);
```
morehouse: Nit: it's clearer to have `decoded_data_size` right next to where it is used
```suggestion

    if (decoded_data.size() > max_size) {
        return initial_input(fuzz_data, max_size);
    }

    // Resize the vector to the maximum possible size BEFORE mutation.
    size_t decoded_data_size = decoded_data.size();
    decoded_data.resize(max_size);

    size_t mutated_size = LLVMFuzzerMutate(decoded_data.data(),
                                           decoded_data_size,
                                           max_size);
```

modules/custommutator/bech32.cpp
erickcestari: Fixed!

main.cpp
erickcestari: Fixed

main.cpp
erickcestari: Fixed

modules/custommutator/bech32.h
```diff
@@ -44,6 +44,10 @@ enum CharLimit : size_t {
  *  assertion error. Encoding must be one of BECH32 or BECH32M. */
 std::string Encode(Encoding encoding, const std::string& hrp, const std::vector<uint8_t>& values);
 
+/** Encode a Bech32 or Bech32m string without the checksum. If hrp contains uppercase characters, this will cause an
+ *  assertion error. Encoding must be one of BECH32 or BECH32M. */
+std::string EncodeNoChecksum(const std::string_view hrp, const std::vector<uint8_t>& values);
```
morehouse: Nit:  `string_view` already prevents mutation of the string it points to.  The `const` here only prevents us from reassigning `str` to a different string.  Usually we omit such consts in parameters (here and elsewhere).

```suggestion
std::string EncodeNoChecksum(std::string_view hrp, const std::vector<uint8_t>& values);
```

morehouse (approved):

modules/custommutator/bech32.h
erickcestari: Fixed!

morehouse (approved):
Code LGTM

PR #213, opened 2025-06-25
bitcoin: drop Boost usage

fanquake:
I was following the instructions for a manual build of the bitcoin module, and it wasn't clear why Core would have a dependency on Boost. It seems like everything builds fine without it? The only (real) reference to Boost I could find is in the `interfaces/handler.h` header.


brunoerg (approved):
ACK f2590b539fb64cf987001b6e66c82f417241e0d5

Nice catch! It's probably a leftover from a previous approach.

PR #217, opened 2025-06-27
Add Target for Bitcoin p2p message parsing

erickcestari:
Add a new target parse_p2p_message for differential fuzzing of Bitcoin v1 P2P messages

### Changes
- Add btcd suport for the target
- Add rust-bitcoin suport for the target
- Add custom mutator to create valid checksums

### DIFF
--- a/modules/btcd/module.cpp
+++ b/modules/btcd/module.cpp
@@ -1,4 +1,5 @@
 #include <span>
+#include <cstring>
 
 #include "module.h"
 #include "btcd_wrapper/libbtcd_wrapper.h"
@@ -16,6 +17,27 @@ namespace bitcoinfuzz
                 .length = static_cast<int>(input_data.size())};
 
             return BTCDEvalScript(script_data, /*flags=*/0) == 1;
+        }
+
+        std::optional<std::string> Btcd::parse_p2p_message(std::span<const uint8_t> buffer) const
+        {
+            ByteArray message_data{
+                .data = reinterpret_cast<char *>(const_cast<uint8_t *>(buffer.data())),
+                .length = static_cast<int>(buffer.size())};
+
+            const auto message_res = BTCDParseP2PMessage(message_data);
+        
+            if (message_res == nullptr) {
+                return std::nullopt;
+            }
+            if (strlen(message_res) == 0) {
+                free(message_res);
+                return std::nullopt;
+            }
+
+            std::string res(message_res);
+            free(message_res);
+            return res;
         }
 
         std::optional<std::string> Btcd::script_asm(std::span<const uint8_t> buffer) const
--- a/modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
+++ b/modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
@@ -1,3 +1,4 @@
+use bitcoin::absolute::Decodable;
 use bitcoin::address::Address;
 use bitcoin::bip152::HeaderAndShortIds;
 use bitcoin::block::BlockUncheckedExt;
@@ -5,7 +6,8 @@ use bitcoin::consensus::{deserialize_partial, encode, serialize};
 use bitcoin::script::{ScriptBuf, ScriptExt};
 use bitcoin::Block;
 use p2p::address::AddrV2;
-use p2p::message::AddrV2Payload;
+use p2p::message::{AddrV2Payload, RawNetworkMessage};
+use p2p::Magic;
 use std::ffi::CStr;
 use std::ffi::CString;
 use std::os::raw::c_char;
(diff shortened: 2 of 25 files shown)


brunoerg:
Needs rebase

modules/btcd/module.cpp
```diff
@@ -18,6 +18,19 @@ namespace bitcoinfuzz
...
         }
 
+        std::optional<std::string> Btcd::parse_p2p_message(std::span<const uint8_t> buffer) const
+        {
+            ByteArray message_data{
+                .data = reinterpret_cast<char *>(const_cast<uint8_t *>(buffer.data())),
+                .length = static_cast<int>(buffer.size())};
+
+            const auto message_res = BTCDParseP2PMessage(message_data);
+            if (message_res == "") return std::nullopt;
```
brunoerg: `BTCDParseP2PMessage` returns `char*` so you have to use strcmp instead of simply doing `== ""`?

modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
```diff
@@ -71,6 +74,24 @@ pub unsafe extern "C" fn rust_bitcoin_script(data: *const u8, len: usize) -> *mu
     }
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn rust_bitcoin_parse_p2p_message(data: *const u8, len: usize) -> *mut c_char {
```
brunoerg: You should also skip rust-bitcoin when the error is "unsupported segwit version".

brunoerg (changes requested):
I'm not sure about the approach here. Bitcoin Core and btcd are full node implementations, when you call `ReadMessageN` from btcd, I beilieve it reads the message and validates it. However, I'm not sure how much validation is done by simplying decoding into `RawNetworkMessage` for rust-bitcoin. cc @apoelstra

I did a light run on this PR and it crashed due to:

```sh
P2P message parsing failed
Module: Rustbitcoin
Result: tx
Module: Btcd
Result: 0
``` 

btcd failed to parse the `tx` message due to `readScript: transaction output public key script is larger than the max allowed size [count 18446744073709551615, max 4000000]`. I think rust-bitcoin is not supposed to do this check when simplying decoding to `RawNetworkMessage`. Not sure.

modules/btcd/module.cpp
erickcestari: Taking a closer look. The `BTCDParseP2PMessage` will always return some value: "0" or the Command string. So this check isn't necessary.

It looks this problem happens in 

```cpp
std::optional<std::string> Btcd::script_asm(std::span<const uint8_t> buffer) const
        {
            ByteArray script_data{
                .data = reinterpret_cast<char *>(const_cast<uint8_t *>(buffer.data())),
                .length = static_cast<int>(buffer.size())};

            const auto script_asm_res = BTCDScriptAsm(script_data);
            if (script_asm_res == "") return std::nullopt;
            std::string res(script_asm_res);
            free(script_asm_res);
            return res;
        }
```

We should probrably add checks like this:
```cpp
std::optional<std::string> Btcd::script_asm(std::span<const uint8_t> buffer) const
        {
            ByteArray script_data{
                .data = reinterpret_cast<char *>(const_cast<uint8_t *>(buffer.data())),
                .length = static_cast<int>(buffer.size())};

            const auto script_asm_res = BTCDScriptAsm(script_data);
            
            // Check if the result is null
            if (script_asm_res == nullptr) {
                return std::nullopt;
            }
            
            // Check if the result is an empty string
            if (strlen(script_asm_res) == 0) {
                free(script_asm_res);
                return std::nullopt;
            }
            
            std::string res(script_asm_res);
            free(script_asm_res);
            return res;
        }
```

modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
erickcestari: Done!

brunoerg:
Needs rebase.

modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
```diff
@@ -6,7 +6,11 @@ use std::str::{FromStr, Utf8Error};
 
 use bitcoin::address::Address;
 use bitcoin::consensus::{deserialize_partial, encode, serialize};
+use bitcoin::consensus::Decodable;
+use bitcoin::consensus::{deserialize_partial, encode};
```
brunoerg: Why are you importing `deserialize_partial` and `encode` again? I think this is causing CI to fail.

erickcestari: Fixed!

brunoerg:
Needs rebase

brunoerg (approved):
ACK a418fe1161367ec38e106e6bd7434b2211bdc135

brunoerg:
I'm still not super convinced about this because of Bitcoin Core. It would need a specific ProcessMessage for us - I mean, we would need to write a tiny ProcessMessage to be used here. But anyway, it might be fun to play as is with the current modules (btcd, rust-bitcoin).

PR #219, opened 2025-07-10
fix: decode the address before passing it to DecodeString

yuvicc:
We need to decode the address string before passing it to the `DecodeString` method so that it can be detected as `INVALID` if it is an invalid address string.


brunoerg:
code review ACK 037756be297952a12402d593e464db5b013a2bc1

PR #222, opened 2025-07-10
  Return block hash on successful deserialization (#97)

### LINKED ISSUE #97: Return block hash on `block_deserialization` success
Instead of just returning `std::optional<bool>`,` block_deserialization` functions could return the block hash. This would allow us to compare hashes for successful deserialization and extend fuzzing to cover hash validation.

**Proposed behavior**:

- Failed blocks return `false` or `0`.
- Cases that currently return null in `std::optional<bool> `return an empty string.
- Successful deserialization returns the block hash

bc1cindy:
Changes:
  - Modified `deserialize_block` functions to return block hash instead of boolean
  - Updated all modules (Bitcoin Core, BTCD, Rust-Bitcoin) to return hash string
  - Updated driver logic to compare hash strings

  This allows comparing block hashes between different implementations during fuzzing.
- Fixes #97

### DIFF
--- a/modules/bitcoin/module.cpp
+++ b/modules/bitcoin/module.cpp
@@ -284,22 +284,22 @@ std::optional<bool> Bitcoin::miniscript_parse(std::string str) const
     return false;
 }
 
-std::optional<std::vector<bool>> Bitcoin::deserialize_block(std::span<const uint8_t> buffer) const
+std::optional<std::string> Bitcoin::deserialize_block(std::span<const uint8_t> buffer) const
 {
     DataStream ds{buffer};
     CBlock block;
-    std::vector<bool> res{};
     try {
         ds >> TX_WITH_WITNESS(block);
     } catch (const std::ios_base::failure&) {
-        return res;
+        return std::nullopt;
     }
     if (block.vtx.empty()) {
-        res.push_back(false);
-    } else {
-        res.push_back(!IsBlockMutated(block, true));
+        return "0";
     }
-    return res;
+    if (IsBlockMutated(block, true)) {
+        return "0";
+    }
+    return block.GetHash().ToString();
 }
 
 std::optional<std::string> Bitcoin::script_asm(std::span<const uint8_t> buffer) const
--- a/modules/btcd/btcd_wrapper/wrapper.go
+++ b/modules/btcd/btcd_wrapper/wrapper.go
@@ -133,7 +133,7 @@ func BTCDDesBlock(scriptData C.ByteArray) *C.char {
 		return C.CString("0")
 	}
 
-	return C.CString("true")
+	return C.CString(block.Hash().String())
 }
 
 //export BTCDFreeString
--- a/modules/btcd/module.cpp
+++ b/modules/btcd/module.cpp
@@ -31,7 +31,7 @@ namespace bitcoinfuzz
             return res;
         }
 
-        std::optional<std::vector<bool>> Btcd::deserialize_block(std::span<const uint8_t> buffer) const
+        std::optional<std::string> Btcd::deserialize_block(std::span<const uint8_t> buffer) const
         {
             ByteArray script_data{
                 .data = reinterpret_cast<char *>(const_cast<uint8_t *>(buffer.data())),
@@ -40,8 +40,10 @@ namespace bitcoinfuzz
             auto pointer{BTCDDesBlock(script_data)};
             std::string result(pointer);
             BTCDFreeString(pointer);
-            std::vector<bool> final_result{"true" == result};
-            return final_result;
+            if (result == "unsupported segwit version") {
+                return std::nullopt;
+            }
+            return result;
         }
 
         std::optional<std::string> Btcd::addrv2_parse(std::span<const uint8_t> buffer) const
--- a/modules/rustbitcoin/module.cpp
+++ b/modules/rustbitcoin/module.cpp
@@ -17,7 +17,7 @@ namespace bitcoinfuzz
             return result;
         }
 
-        std::optional<std::vector<bool>> Rustbitcoin::deserialize_block(std::span<const uint8_t> buffer) const
+        std::optional<std::string> Rustbitcoin::deserialize_block(std::span<const uint8_t> buffer) const
         {
             auto pointer{rust_bitcoin_des_block(buffer.data(), buffer.size())};
             std::string result(pointer);
@@ -25,8 +25,7 @@ namespace bitcoinfuzz
             if (result == "unsupported segwit version") {
                 return std::nullopt;
             }
-            std::vector<bool> final_result{"true" == result};
-            return final_result;
+            return result;
         }
 
         std::optional<std::string> Rustbitcoin::address_parse(std::string str) const
--- a/modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
+++ b/modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
@@ -41,8 +41,13 @@ pub unsafe extern "C" fn rust_bitcoin_des_block(
 
     match res {
         Ok(res) => {
+            let block_hash = res.0.block_hash();
             let is_valid = res.0.validate().is_ok();
-            return str_to_c_string(is_valid.to_string().as_str());
+            if is_valid {
+                return str_to_c_string(&block_hash.to_string());
+            } else {
+                return str_to_c_string("0");
+            }
         }
         Err(err) => {
             if err.to_string().starts_with("unsupported segwit version") {
--- a/driver.cpp
+++ b/driver.cpp
@@ -30,11 +30,11 @@ namespace bitcoinfuzz
 
     void Driver::BlockDeserializationTarget(std::span<const uint8_t> buffer) const
     {
-        std::optional<std::vector<bool>> last_response{std::nullopt};
+        std::optional<std::string> last_response{std::nullopt};
         for (auto& module : modules)
         {
-            std::optional<std::vector<bool>> res{module.second->deserialize_block(buffer)};
-            if (!res.has_value() || res->empty()) continue;
+            std::optional<std::string> res{module.second->deserialize_block(buffer)};
+            if (!res.has_value()) continue;
             if (last_response.has_value()) assert(*res == *last_response);
             last_response = res.value();
         }
--- a/include/bitcoinfuzz/basemodule.cpp
+++ b/include/bitcoinfuzz/basemodule.cpp
@@ -13,7 +13,7 @@ namespace bitcoinfuzz
         return std::nullopt;
     }
 
-    std::optional<std::vector<bool>> BaseModule::deserialize_block(std::span<const uint8_t> buffer) const
+    std::optional<std::string> BaseModule::deserialize_block(std::span<const uint8_t> buffer) const
     {
         return std::nullopt;
     }
--- a/include/bitcoinfuzz/basemodule.h
+++ b/include/bitcoinfuzz/basemodule.h
@@ -18,7 +18,7 @@ namespace bitcoinfuzz
         BaseModule(const std::string &name) : name(name) {}
 
         virtual std::optional<std::string> script_parse(std::span<const uint8_t> buffer) const;
-        virtual std::optional<std::vector<bool>> deserialize_block(std::span<const uint8_t> buffer) const;
+        virtual std::optional<std::string> deserialize_block(std::span<const uint8_t> buffer) const;
         virtual std::optional<bool> script_eval(const std::vector<uint8_t>& input_data, unsigned int flags, size_t version) const;
         virtual std::optional<bool> descriptor_parse(std::string str) const;
         virtual std::optional<bool> miniscript_parse(std::string str) const;
--- a/modules/bitcoin/module.h
+++ b/modules/bitcoin/module.h
@@ -15,7 +15,7 @@ namespace bitcoinfuzz
         public:
             Bitcoin(void);
             std::optional<std::string> script_parse(std::span<const uint8_t> buffer) const override;
-            std::optional<std::vector<bool>> deserialize_block(std::span<const uint8_t> buffer) const override;
+            std::optional<std::string> deserialize_block(std::span<const uint8_t> buffer) const override;
             std::optional<bool> script_eval(const std::vector<uint8_t>& input_data, unsigned int flags, size_t version) const override;
             std::optional<bool> descriptor_parse(std::string str) const override;
             std::optional<bool> miniscript_parse(std::string str) const override;
--- a/modules/btcd/module.h
+++ b/modules/btcd/module.h
@@ -15,7 +15,7 @@ namespace bitcoinfuzz
             Btcd(void);
             std::optional<bool> script_eval(const std::vector<uint8_t>& input_data, unsigned int flags, size_t version) const override;
             std::optional<std::string> script_asm(std::span<const uint8_t> buffer) const override;
-            std::optional<std::vector<bool>> deserialize_block(std::span<const uint8_t> buffer) const override;
+            std::optional<std::string> deserialize_block(std::span<const uint8_t> buffer) const override;
             std::optional<std::string> address_parse(std::string str) const override;
             std::optional<std::string> psbt_parse(std::span<const uint8_t> buffer) const override;
             std::optional<std::string> addrv2_parse(std::span<const uint8_t> buffer) const override;
--- a/modules/rustbitcoin/module.h
+++ b/modules/rustbitcoin/module.h
@@ -15,7 +15,7 @@ namespace bitcoinfuzz
         public:
             Rustbitcoin(void);
             std::optional<std::string> script_parse(std::span<const uint8_t> buffer) const override;
-            std::optional<std::vector<bool>> deserialize_block(std::span<const uint8_t> buffer) const override;
+            std::optional<std::string> deserialize_block(std::span<const uint8_t> buffer) const override;
             std::optional<std::string> address_parse(std::string str) const override;
             std::optional<std::string> psbt_parse(std::span<const uint8_t> buffer) const override;
             std::optional<std::string> addrv2_parse(std::span<const uint8_t> buffer) const override;


moisesPompilio:
Great, thanks for working on this! It’s looking good already!

moisesPompilio:
Currently, `deserialize_block` uses an `optional<std::vector<bool>>` to indicate the result:  
- `true` for success,  
- `false` for failure,  
- `null` if segwit is unsupported.

In this PR, since it uses `optional<std::string>`, part of the old logic is kept:  
- returns a hash string if deserialization succeeds,  
- returns an empty string if it fails,  
- returns null if segwit is unsupported.

This seems to be the direction the PR is taking, and honestly, it looks like a very solid approach.  
It’s a bit different from what I initially had in mind for issue #97 , but I think it makes a lot of sense.  
What do you think, @brunoerg  and @erickcestari ?

yuvicc:
Concept ACK

I think this is a better way to fuzz for `deserialize_block` target, instead of return boolean value we are returning hash string.

yuvicc:
Code review and Tested ACK e1a120f5895ac01d2742f7469015ea48eabe4944

Thanks.

modules/btcd/module.cpp
```diff
@@ -40,8 +40,10 @@ namespace bitcoinfuzz
             auto pointer{BTCDDesBlock(script_data)};
             std::string result(pointer);
             BTCDFreeString(pointer);
-            std::vector<bool> final_result{"true" == result};
-            return final_result;
+            if (result.empty() || result == "0") {
+                return std::nullopt;
+            }
+            return result;
```
erickcestari, line 46:
```suggestion
            return result;
```
We can return the result directly here

modules/rustbitcoin/module.cpp
```diff
@@ -17,16 +17,18 @@ namespace bitcoinfuzz
...
-        std::optional<std::vector<bool>> Rustbitcoin::deserialize_block(std::span<const uint8_t> buffer) const
+        std::optional<std::string> Rustbitcoin::deserialize_block(std::span<const uint8_t> buffer) const
         {
             auto pointer{rust_bitcoin_des_block(buffer.data(), buffer.size())};
             std::string result(pointer);
             free_c_string(pointer);
-            if (result == "unsupported segwit version") {
+            if (result == "unsupported segwit version" || result == "false" || result.empty()) {
                 return std::nullopt;
             }
```
erickcestari:
```suggestion
            if (result == "unsupported segwit version") {
                return std::nullopt;
            }
```

This is the only case that we will skip

erickcestari (changes requested):
I think it's mostly correct! Thanks!

We just need to be careful when returning `std::nullopt` or an empty string in a Module class, as it signals that we want to skip the result checks. This should only be done for very specific cases where we know the input would crash the fuzzer for a reason we already understand and acknowledge.

So in this case, instead of returning `std::nullopt`, we should return a `"0"` string. That’s because this is an error we do want to check across different implementations. If one implementation fails and another doesn't, we might have discovered a bug.

```cpp
void Driver::BlockDeserializationTarget(std::span<const uint8_t> buffer) const
    {
        std::optional<std::string> last_response{std::nullopt};
        for (auto& module : modules)
        {
            std::optional<std::string> res{module.second->deserialize_block(buffer)};
            if (!res.has_value() || res->empty()) continue;
            if (last_response.has_value()) assert(*res == *last_response);
            last_response = res.value();
        }
    }
```

erickcestari:
> Currently, `deserialize_block` uses an `optional<std::vector<bool>>` to indicate the result:
> 
> * `true` for success,
> * `false` for failure,
> * `null` if segwit is unsupported.
> 
> In this PR, since it uses `optional<std::string>`, part of the old logic is kept:
> 
> * returns a hash string if deserialization succeeds,
> * returns an empty string if it fails,
> * returns null if segwit is unsupported.
> 
> This seems to be the direction the PR is taking, and honestly, it looks like a very solid approach. It’s a bit different from what I initially had in mind for issue #97 , but I think it makes a lot of sense. What do you think, @brunoerg and @erickcestari ?

I like the idea, so we will need to change the way that we interpret the result at the `BlockDeserializationTarget`. Removing the `res->empty()` check.

```cpp
void Driver::BlockDeserializationTarget(std::span<const uint8_t> buffer) const
    {
        std::optional<std::string> last_response{std::nullopt};
        for (auto& module : modules)
        {
            std::optional<std::string> res{module.second->deserialize_block(buffer)};
            if (!res.has_value() || res->empty()) continue;
            if (last_response.has_value()) assert(*res == *last_response);
            last_response = res.value();
        }
    }
```

to

```cpp
void Driver::BlockDeserializationTarget(std::span<const uint8_t> buffer) const
    {
        std::optional<std::string> last_response{std::nullopt};
        for (auto& module : modules)
        {
            std::optional<std::string> res{module.second->deserialize_block(buffer)};
            if (!res.has_value()) continue;
            if (last_response.has_value()) assert(*res == *last_response);
            last_response = res.value();
        }
    }
```

bc1cindy:
Updated the code based on @erickcestari feedback.

brunoerg:
@bc1cindy did you run it on your machine?

bc1cindy:
> @bc1cindy did you run it on your machine?

@brunoerg  yes, I tested it locally and everything works! The build compiles and the fuzzer runs normally.

Is there a specific data corpus or additional test you recommend running to validate the changes?

modules/bitcoin/module.cpp
```diff
@@ -284,22 +284,22 @@ std::optional<bool> Bitcoin::miniscript_parse(std::string str) const
...
+std::optional<std::string> Bitcoin::deserialize_block(std::span<const uint8_t> buffer) const
 {
     DataStream ds{buffer};
     CBlock block;
-    std::vector<bool> res{};
     try {
         ds >> TX_WITH_WITNESS(block);
     } catch (const std::ios_base::failure&) {
-        return res;
+        return "0";
```
moisesPompilio: I think this should be a `std::nullopt` return, right?

modules/btcd/btcd_wrapper/wrapper.go
```diff
@@ -118,22 +118,22 @@ func BTCDDesBlock(scriptData C.ByteArray) *C.char {
...
 	err = blockchain.CheckBlockSanity(block, powLimit, blockchain.NewMedianTime())
 	if err != nil {
-		return C.CString("0")
+		return C.CString("")
 	}
 
 	err = blockchain.ValidateWitnessCommitment(block)
 	if err != nil {
-		return C.CString("0")
+		return C.CString("")
```
moisesPompilio: I think this return change isn't needed anymore, since empty strings become zero. So it should be fine to just return `C.CString("0") `here, right?

modules/rustbitcoin/module.cpp
```diff
@@ -17,16 +17,18 @@ namespace bitcoinfuzz
...
         {
             auto pointer{rust_bitcoin_des_block(buffer.data(), buffer.size())};
             std::string result(pointer);
             free_c_string(pointer);
             if (result == "unsupported segwit version") {
                 return std::nullopt;
             }
-            std::vector<bool> final_result{"true" == result};
-            return final_result;
+            if (result == "false" || result.empty() || result == "true") {
```
moisesPompilio: `result == "false"` and `result == "true"` shouldn't happen anymore, right?

modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
```diff
@@ -40,13 +40,17 @@ pub unsafe extern "C" fn rust_bitcoin_des_block(
     match res {
         Ok(res) => {
             let check = res.0.check_merkle_root() && res.0.check_witness_commitment();
-            return str_to_c_string(check.to_string().as_str());
+            if check {
+                return str_to_c_string(&res.0.block_hash().to_string());
+            } else {
+                return str_to_c_string("");
```
moisesPompilio:
```suggestion
                return str_to_c_string("0");
```
This way we can remove the empty check later, since in the end we just want "0" for comparison.

bc1cindy:
@moisesPompilio  @erickcestari thanks for the review! 

i've addressed all the requested changes:
  - changed empty strings to "0" in btcd wrapper and rust-bitcoin lib
  - removed the "false"/"true" checks in rustbitcoin module
  - all modules now return "0" for errors consistently
  - the build compiles and the fuzzer runs normally.

modules/btcd/module.cpp
```diff
@@ -40,8 +40,10 @@ namespace bitcoinfuzz
             auto pointer{BTCDDesBlock(script_data)};
             std::string result(pointer);
             BTCDFreeString(pointer);
-            std::vector<bool> final_result{"true" == result};
-            return final_result;
+            if (result.empty()) {
+                return "0";
+            }
```
moisesPompilio: At this point, shouldn't result always be either "0" or the block hash?

modules/rustbitcoin/module.cpp
```diff
@@ -17,16 +17,18 @@ namespace bitcoinfuzz
...
             std::string result(pointer);
             free_c_string(pointer);
             if (result == "unsupported segwit version") {
                 return std::nullopt;
             }
-            std::vector<bool> final_result{"true" == result};
-            return final_result;
+            if (result == "0" || result.empty()) {
+                return "0";
+            }
```
moisesPompilio: At this point, shouldn't result always be either "0" or the block hash? If so, this check may no longer be necessary.

modules/bitcoin/module.cpp
moisesPompilio: @bc1cindy  Shouldn't this return `std::nullopt` instead? Since this failure likely means segwit isn't supported, returning `"0"` here might be misleading.

bc1cindy:
thank you. 

updated the code based on @moisesPompilio feedback.
the build compiles and the fuzzer runs normally.

modules/bitcoin/module.cpp
moisesPompilio: Shouldn't this return `std::nullopt` instead? Since this failure likely means segwit isn't supported, returning "0" here might be misleading.

modules/btcd/btcd_wrapper/wrapper.go
```diff
@@ -107,7 +107,7 @@ func BTCDScriptAsm(scriptData C.ByteArray) *C.char {
 	script := C.GoBytes(unsafe.Pointer(scriptData.data), scriptData.length)
 	disasm, err := txscript.DisasmString(script)
 	if err != nil {
-		return C.CString("")
+		return C.CString("0")
```
moisesPompilio: It seems like this is modifying `BTCDScriptAsm` instead of `BTCDDesBlock`.

moisesPompilio:
While testing, I noticed that `btcd` was causing the fuzzer to stop very early during block deserialization.  
Looking into the error, it seems this happens because the current implementation applies stricter validation rules than `rust-bitcoin` and `bitcoin core`.

Additionally, it doesn’t handle the `"unsupported segwit version"` case like the other modules do.  
Since this PR already touches this part of the `btcd` code, I tried aligning its behavior with the rest.

The snippet below is what I tested, and it made `btcd` behave consistently with the other modules during fuzzing:

`/modules/btcd/btcd_wrapper/wrapper.go`
```go
//export BTCDDesBlock
func BTCDDesBlock(scriptData C.ByteArray) *C.char {
	buffer := C.GoBytes(unsafe.Pointer(scriptData.data), scriptData.length)

	var msgBlock wire.MsgBlock
	err := msgBlock.Deserialize(bytes.NewReader(buffer))
	if err != nil {
		if strings.Contains(strings.ToLower(err.Error()), "witness") {
			return C.CString("unsupported segwit version")
		}
		return C.CString("0")
	}

	block := btcutil.NewBlock(&msgBlock)

	if !validateMerkleRoot(block) {
		return C.CString("0")
	}

	err = blockchain.ValidateWitnessCommitment(block)
	if err != nil {
		if strings.Contains(strings.ToLower(err.Error()), "witness") {
			return C.CString("unsupported segwit version")
		}
		return C.CString("0")
	}

	return C.CString(block.Hash().String())
}

func validateMerkleRoot(block *btcutil.Block) bool {
	defer func() {
		if r := recover(); r != nil {
			// if there is a panic, we consider the block invalid
		}
	}()

	transactions := block.Transactions()
	if len(transactions) == 0 {
		return false
	}

	merkleStore := blockchain.BuildMerkleTreeStore(transactions, false)
	if len(merkleStore) == 0 {
		return false
	}

	calculatedMerkleRoot := merkleStore[len(merkleStore)-1]
	return block.MsgBlock().Header.MerkleRoot.IsEqual(calculatedMerkleRoot)
}
```
`/modules/btcd/module.cpp`
```c++
std::optional<std::string> Btcd::deserialize_block(std::span<const uint8_t> buffer) const {
    ByteArray script_data{
        .data = reinterpret_cast<char *>(const_cast<uint8_t *>(buffer.data())),
        .length = static_cast<int>(buffer.size())};

    auto pointer{BTCDDesBlock(script_data)};
    std::string result(pointer);
    BTCDFreeString(pointer);
    if (result == "unsupported segwit version") {
        return std::nullopt;
    }
    return result;
}
``` 
If it makes sense, maybe @bc1cindy  could push this as a separate commit — just to document and test the alignment of the btcd module with the others.

What do you think about this approach, @erickcestari  and @brunoerg ?

bc1cindy:
done @moisesPompilio 

everything worked out fine here.

modules/rustbitcoin/module.cpp
```diff
@@ -17,16 +17,18 @@ namespace bitcoinfuzz
...
             std::string result(pointer);
             free_c_string(pointer);
             if (result == "unsupported segwit version") {
                 return std::nullopt;
             }
-            std::vector<bool> final_result{"true" == result};
-            return final_result;
+            if (result == "0") {
+                return "0";
+            }
```
erickcestari:
```suggestion
```
We can remove this if

modules/btcd/btcd_wrapper/wrapper.go
```diff
@@ -237,4 +242,25 @@ func BTCDAddress(data C.ByteArray) *C.char {
 	return C.CString(prefix + addr.EncodeAddress())
 }
 
+func validateMerkleRoot(block *btcutil.Block) bool {
+	defer func() {
+		if r := recover(); r != nil {
+			// if there is a panic, we consider the block invalid
+		}
+	}()
```
erickcestari: I'm not sure about this. Why would this function panic? And if it does, wouldn't that mean we've found a bug? @moisesPompilio

erickcestari (changes requested):

brunoerg:
Needs rebase

modules/btcd/btcd_wrapper/wrapper.go
bc1cindy: waiting for @moisesPompilio in this

bc1cindy:
> Needs rebase

done @brunoerg

modules/btcd/btcd_wrapper/wrapper.go
moisesPompilio: Good question @erickcestari  — the panic was happening in `BuildMerkleTreeStore`, likely due to invalid transaction data (e.g., input like `CgAAAAAAAAAAAAAAAAAAAAAAIP2QDNl1wKI9GNIJWd+6YQL8lEr8l8LeaaIPhqpqgqUAAAAAAAAAAACAAAAAAAAAAAAAAAC9vb0AIAAAAAAA).  

The old code used `recover` to handle it, but I think a simple validation step might be enough to avoid the panic.  
This validation checks if the transactions are non-nil and have at least one input or output, which seems to be enough to prevent `BuildMerkleTreeStore` from panicking.


So this version below might be sufficient:

```go
func validateMerkleRoot(block *btcutil.Block) bool {
	transactions := block.Transactions()
	if len(transactions) == 0 {
		return false
	}
	// Check if the block has transactions valid
	for _, tx := range transactions {
		if tx == nil || tx.MsgTx() == nil {
			return false
		}
		// check if the transaction has at least 1 input and 1 output
		if len(tx.MsgTx().TxIn) == 0 && len(tx.MsgTx().TxOut) == 0 {
			return false
		}
	}

	merkleStore := blockchain.BuildMerkleTreeStore(transactions, false)
	if len(merkleStore) == 0 {
		return false
	}

	calculatedMerkleRoot := merkleStore[len(merkleStore)-1]

	return block.MsgBlock().Header.MerkleRoot.IsEqual(calculatedMerkleRoot)
}
```

brunoerg: What do you guys mean with `panic`? `panic` is likely that the application is crashing with the input.

bc1cindy:
implemented the new version and added @moisesPompilio as co-author.
the build compiles and the fuzzer runs normally.

modules/btcd/btcd_wrapper/wrapper.go
erickcestari: Yes. in Go, when panics (will crash), you can use `recover()` to catch the panic and prevent the entire program from crashing. That's why I was asking about using `recover` in the code. But it looks like he removed it now.

brunoerg: Well, if it's panicking we should properly report it and wait for a fix instead of simply bypassing it.

moisesPompilio: Good point — I looked deeper into the `btcd` code and noticed it has two functions to calculate the merkle root: `CalcMerkleRoot` and `BuildMerkleTreeStore`. I initially thought there was only one.

It turns out `CalcMerkleRoot` is the one used throughout most of the codebase. It’s simpler, doesn’t require any transaction validation, and doesn’t panic. On the other hand, I couldn’t find where `BuildMerkleTreeStore` is actually used, so it’s unclear why it requires fully valid transactions to avoid panics.

So by switching to CalcMerkleRoot, the code gets cleaner and we no longer need extra validation or a recover block. Here's the version I tested:

```go
func BTCDDesBlock(scriptData C.ByteArray) *C.char {
	buffer := C.GoBytes(unsafe.Pointer(scriptData.data), scriptData.length)

	var msgBlock wire.MsgBlock
	err := msgBlock.Deserialize(bytes.NewReader(buffer))
	if err != nil {
		if strings.Contains(strings.ToLower(err.Error()), "witness") {
			return C.CString("unsupported segwit version")
		}
		return C.CString("0")
	}

	block := btcutil.NewBlock(&msgBlock)

	transactions := block.Transactions()
	if len(transactions) == 0 {
		return C.CString("0")
	}
	calculatedMerkleRoot := blockchain.CalcMerkleRoot(transactions, false)
	if !block.MsgBlock().Header.MerkleRoot.IsEqual(&calculatedMerkleRoot) {
		return C.CString("0")
	}

	err = blockchain.ValidateWitnessCommitment(block)
	if err != nil {
		if strings.Contains(strings.ToLower(err.Error()), "witness") {
			return C.CString("unsupported segwit version")
		}
		return C.CString("0")
	}

	return C.CString(block.Hash().String())
}
```

bc1cindy:
updated switching to CalcMerkleRoot, so the code gets cleaner and we no longer need extra validation or a recover block, following @moisesPompilio feedback

brunoerg:
Still needs rebase.

bc1cindy:
> Still needs rebase.

@brunoerg  i've verified that this branch is already rebased on the latest v2 (commit eeded61), the merge-base shows it's up to date. 

do you want me to open a new PR?

bc1cindy:
> Still needs rebase.

rebased, and fixed rust_bitcoin_lib/src/lib.rs - API changed from check_merkle_root() to validate()
fuzzer compiled and ran successfully @brunoerg

moisesPompilio:
The code looks good now, but I think it should be split into two commits:  
- One for making all deserialization block functions return the hash (including the necessary change in `btcd` to keep the test working).  
- Another just to align `btcd`'s behavior with the other implementations.

bc1cindy:
> The code looks good now, but I think it should be split into two commits:
> 
> * One for making all deserialization block functions return the hash (including the necessary change in `btcd` to keep the test working).
> * Another just to align `btcd`'s behavior with the other implementations.

done @moisesPompilio

moisesPompilio:
ACK 67dcc873a83990f4bb8b5ffc4ff3523972a1d661

modules/btcd/btcd_wrapper/wrapper.go
```diff
@@ -116,24 +115,35 @@ func BTCDScriptAsm(scriptData C.ByteArray) *C.char {
...
 	if err != nil {
+		if strings.Contains(strings.ToLower(err.Error()), "witness") {
+			return C.CString("unsupported segwit version")
+		}
 		return C.CString("0")
 	}
 
-	return C.CString("true")
+	return C.CString(block.Hash().String())
 }
```
erickcestari: I don't think these specific changes are beneficial. We originally added the skip for the "unsupported segwit version" error in rust-bitcoin module because it was a known issue. With the new changes, the code skips on any error that contains "witness" in the name, and there are plenty of those.

Another concern is with `CheckBlockSanity`. I think it's better to keep it as it was. If this function performs additional checks compared to rust-bitcoin or Bitcoin Core, we should only skip specific errors with a clear rationale for each case.

My suggestion would be to drop the commit “Align btcd validation with other implementations” for now. We can revisit this in a follow-up PR, where we can go through the changes in more detail and continue the discussion. What do you think @moisesPompilio ?

moisesPompilio:
> My suggestion would be to drop the commit “Align btcd validation with other implementations” for now. We can revisit this in a follow-up PR, where we can go through the changes in more detail and continue the discussion. What do you think @moisesPompilio ?

Yes, my suggestion was before PR 23 was merged — at that point, Rust-Bitcoin and Bitcoin Core were still aligned, and only BTCD behaved differently. But after the merge, Rust-Bitcoin is no longer aligned with Bitcoin Core, so the change I suggested would now make BTCD align only with Bitcoin Core. We should open a separate issue to discuss what each implementation should validate before attempting to compare the block hash.

bc1cindy:
> > My suggestion would be to drop the commit “Align btcd validation with other implementations” for now. We can revisit this in a follow-up PR, where we can go through the changes in more detail and continue the discussion. What do you think @moisesPompilio ?
> 
> Yes, my suggestion was before PR 23 was merged — at that point, Rust-Bitcoin and Bitcoin Core were still aligned, and only BTCD behaved differently. But after the merge, Rust-Bitcoin is no longer aligned with Bitcoin Core, so the change I suggested would now make BTCD align only with Bitcoin Core. We should open a separate issue to discuss what each implementation should validate before attempting to compare the block hash.

removed @moisesPompilio @erickcestari

erickcestari (approved):
ACK a3240b15d12bf9dc43dff8970bdfc3aaab6c5027

moisesPompilio:
ACK a3240b15d12bf9dc43dff8970bdfc3aaab6c5027

modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
```diff
@@ -41,8 +41,13 @@ pub unsafe extern "C" fn rust_bitcoin_des_block(
 
     match res {
         Ok(res) => {
+            let block_hash = res.0.block_hash();
```
brunoerg, line 44: nit: `block_hash` could be defined into `if is_valid {`

PR #223, opened 2025-07-11
rust-bitcoin: use it based on git commit

brunoerg:
We usually fuzz rust-bitcoin based on a release (e.g. `0.32.6`). However, for the other modules we usually fuzz them based on master branch or recent code. This allows us to find bugs before they get released. That said, this PR changes the rust-bitcoin module to fetch it from the git repository. Note that many changes had to be done due to diverse changes made since the release of 0.32.6.


brunoerg:
note: we could update the commit hash every 1 week, for example.

yuvicc (commented):
lgtm ACK 241676122d80ebf6b9363aec0928a5e53f2c3d7b

Thanks for this!

PR #226, opened 2025-07-16
feat: add a new workaround in `rust_bitcoin_des_block`

erickcestari:
After updating rust-bitcoin to use the `master` branch, a new validation was introduced on `Amount` that checks if a transaction spends more than 21,000,000 bitcoins. We should skip this input, as we aren't performing it in other implementations.


brunoerg (approved):
code review ACK a85cae616a32ebed46a41176cc2640022d71d171

PR #127, opened 2025-04-03
Added target compactblocks_parse for Bitcoin Core and Rust Bitcoin

yuvicc:
### Added new target `compactblocks_parse` for Bitcoin Core and Rust Bitcoin

see commit message for more details.

PR #231, opened 2025-07-22
update: return length for compact blocks parse target

yuvicc:
This way we can keep the return type as `unsigned integer`, see [comment](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/127#discussion_r2193125151).

### DIFF
--- a/modules/bitcoin/module.cpp
+++ b/modules/bitcoin/module.cpp
@@ -1,3 +1,4 @@
+#include <cstdint>
 #include <optional>
 #include <span>
 
@@ -456,7 +457,7 @@ std::optional<std::string> Bitcoin::psbt_parse(std::span<const uint8_t> buffer) 
     return result;
 }
 
-std::optional<int> Bitcoin::cmpctblocks_parse(std::span<const uint8_t> buffer) const
+std::optional<uint32_t> Bitcoin::cmpctblocks_parse(std::span<const uint8_t> buffer) const
 {
     DataStream ds{buffer};
     CBlockHeaderAndShortTxIDs block_header_and_short_txids;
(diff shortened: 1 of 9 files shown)


modules/bitcoin/module.cpp
```diff
@@ -456,27 +456,25 @@ std::optional<std::string> Bitcoin::psbt_parse(std::span<const uint8_t> buffer)
...
 {
     DataStream ds{buffer};
     CBlockHeaderAndShortTxIDs block_header_and_short_txids;
 
     try {
         ds >> block_header_and_short_txids;
     } catch (const std::ios_base::failure& e) {
-        if (std::string(e.what()).find("Superflous witness record") != std::string::npos)
-            return -2;
+        // Return nullopt on any parsing error
```
brunoerg: Why? If we do it, we would miss many mismatches.

yuvicc: Correct, I think a fix would be to use UINT32_MAX and UINT32_MAX - 1 as the error code, since the parse bytes shouldn't exceed 80 as per BIP 152

brunoerg:
My suggestion was to simply change the type from `int` to `uint32_t`. However, I started reviewing this and noticed that there are more changes than it. I think it would be great to explain the other changes and split them in other commits.

yuvicc:
> I think it would be great to explain the other changes and split them in other commits.

Sure!

modules/bitcoin/module.cpp
```diff
@@ -456,27 +457,30 @@ std::optional<std::string> Bitcoin::psbt_parse(std::span<const uint8_t> buffer)
     return result;
 }
 
-std::optional<int> Bitcoin::cmpctblocks_parse(std::span<const uint8_t> buffer) const
+std::optional<uint32_t> Bitcoin::cmpctblocks_parse(std::span<const uint8_t> buffer) const
 {
+    if (buffer.empty()) {
```
brunoerg: If the goal is just changing from int to uint32_t. Why this check was added? Any reason?

brunoerg:
Needs rebase

modules/bitcoin/module.cpp
yuvicc: No, it wasn't there, I shall remove it!

yuvicc:
> Needs rebase

Done!

brunoerg (approved):
code review ACK 9cf612b60866707ec78e0c7ac535bed7b63d7d5a

edit: needs rebase

brunoerg:
Closing due lack of activity. Let me know if it should be reopened.

PR #233, opened 2025-07-28
feat: add new cases to skip in bolt11 rust-lightning

erickcestari:


morehouse:
I think it would be good to have issues for tracking when we can remove these workarounds, and also TODOs within the code itself.

PR #200, opened 2025-06-09
add: libbitcoin module

yuvicc:
# This is as Work In Progress for tracking libbitcoin module in bitcoinfuzz

## Checklist
- [x] add required module files
- [x] makefile for compiling
- [x] module.h/cpp for libbitcoin
- [x] include in workflow 
- [x] add some targets
     - [x] script_parse
     - [x] deserialize_block

addresses #144

PR #236, opened 2025-07-30
fix(eclair): use absolute path to get the eclair lib path

### LINKED ISSUE #234: eclair: No such file or directory ["./modules/eclair/lib"]
This relative path could be broken depending how we're running bitcoinfuzz (e.g. running the clean_corpus script from corpora).

```cpp
    try {
        for (const auto& entry : fs::directory_iterator("./modules/eclair/lib")) {
            auto ext = entry.path().extension();
            if (ext == ".jar" || ext == ".class") {
                if (!first) cp << ":";
                cp << entry.path().string();
                first = false;
            }
        }
    } catch (const fs::filesystem_error& e) {
        std::cerr << "Filesystem error: " << e.what() << std::endl;
    }
```

erickcestari:
Get rootdir at compile time and use it to find the correct path to the eclair lib.

Closes #234


yuvicc:
Similar CI failing in #200

brunoerg (approved):
code review ACK e72d64999988b7c72b649ac33a0f3962b5a2a205

PR #242, opened 2025-07-30
fix(eclair): avoid processing null input

erickcestari:
Should fix this error in ci:
https://github.com/bitcoinfuzz/bitcoinfuzz/actions/runs/16628877694/job/47056243873


erickcestari:
It didn't fix :/

PR #243, opened 2025-07-31
fix: setup java when running the target deserialize_invoice

erickcestari:


brunoerg (approved):
ACK 6ad1dbcf20a10d921bbf1a3ed983484a36937a42

Thank you!

PR #244, opened 2025-07-31
fix: build clightning with clang on macOS instead of cc

erickcestari:
Removes platform-specific build logic and enables sanitizers on macOS by overriding CWARNFLAGS to handle clang's stricter deprecation warnings.

### DIFF
--- a/modules/clightning/Makefile
+++ b/modules/clightning/Makefile
@@ -18,12 +18,7 @@ all: module.a
 
 $(LIGHTNING_DIR)/config.vars:
 	@echo "Running ./configure in $(LIGHTNING_DIR)..."
-	@if [ "$(shell uname)" = "Darwin" ]; then \
-		cd $(LIGHTNING_DIR) && ./configure --disable-compat --disable-valgrind CC=cc; \
-	else \
-		cd $(LIGHTNING_DIR) && ./configure --enable-address-sanitizer --disable-compat --enable-fuzzing --disable-valgrind --enable-ub-sanitizer CC=clang; \
-	fi
-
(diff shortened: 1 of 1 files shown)


modules/clightning/Makefile
```diff
@@ -18,12 +18,7 @@ all: module.a
 
 $(LIGHTNING_DIR)/config.vars:
 	@echo "Running ./configure in $(LIGHTNING_DIR)..."
-	@if [ "$(shell uname)" = "Darwin" ]; then \
-		cd $(LIGHTNING_DIR) && ./configure --disable-compat --disable-valgrind CC=cc; \
-	else \
-		cd $(LIGHTNING_DIR) && ./configure --enable-address-sanitizer --disable-compat --enable-fuzzing --disable-valgrind --enable-ub-sanitizer CC=clang; \
-	fi
-
+	cd $(LIGHTNING_DIR) && ./configure --enable-address-sanitizer --disable-compat --enable-fuzzing --disable-valgrind --enable-debugbuild --enable-ub-sanitizer CC=clang CWARNFLAGS="-Wall -Wundef -Wmissing-prototypes -Wmissing-declarations -Wstrict-prototypes -Wold-style-definition"; \
```
brunoerg: Do we need this `; \`?

```suggestion
	cd $(LIGHTNING_DIR) && ./configure --enable-address-sanitizer --disable-compat --enable-fuzzing --disable-valgrind --enable-debugbuild --enable-ub-sanitizer CC=clang CWARNFLAGS="-Wall -Wundef -Wmissing-prototypes -Wmissing-declarations -Wstrict-prototypes -Wold-style-definition"
```

brunoerg:
Rebase should fix CI issue.

modules/clightning/Makefile
erickcestari: We don't need. Removed.

brunoerg (approved):
ACK d85014d812394d975e018830857519fb49012a15

PR #254, opened 2025-08-08
fix(eclair): amount and short_channel_id format

erickcestari:


morehouse (approved):
ACK https://github.com/bitcoinfuzz/bitcoinfuzz/pull/254/commits/ca6183f1955ccbde4f5a32124081d030a33752b0

PR #258, opened 2025-08-13
feat(eclair): add support to deserialize_offer target

erickcestari:


modules/eclair/EclairWrapper.java
```diff
@@ -0,0 +1,252 @@
...
+      Option<String> issuer = offer.issuer();
+      if (issuer.isDefined()) {
+        sb.append(issuer.get());
+      }
+
+      sb.append(";QUANTITY=");
+      Option<Object> quantityMax = offer.quantityMax();
+      if (quantityMax.isDefined()) {
+        long qty = (Long) quantityMax.get();
+        if (qty != Long.MAX_VALUE) {
```
brunoerg: Why this check?

erickcestari: I removed this check! It was indeed useless. Furthermore in the code you can see a lot of those Object [casts/references], because when any JVM-based language compiles generic types to JVM bytecode, it loses the type information due to type erasure.

brunoerg (approved):
ACK 8c3707563deae1e27790f4cf66c0afe3afdf3a83

PR #261, opened 2025-08-18
doc: add new trophies

erickcestari:


brunoerg:
maybe we should add https://github.com/btcsuite/btcd/issues/2402?

erickcestari:
> maybe we should add [btcsuite/btcd#2402](https://github.com/btcsuite/btcd/issues/2402)?

Yes

brunoerg (approved):
ACK 1558f77eb6841d751f0f6b76116450a7d351d8b8

PR #263, opened 2025-08-26
ldk: use git commit instead of a fixed version

brunoerg:


brunoerg:
Trying to figure out the macOS build issue before fixing anything in the lib code related to the update.

PR #264, opened 2025-08-29
Parse p2p lightning message

erickcestari:


brunoerg (changes requested):
The error on macOS build is because HSMD objs are not being linked. When running `ar -t libcln.a | grep -E "(gossipd_wiregen|hsmd_wiregen)"` it only shows the presence of the gossipd_wiregen.o.

driver.cpp
```diff
@@ -372,6 +372,52 @@ namespace bitcoinfuzz
...
+        std::string last_module_name;
+        const char* mt_env = std::getenv("P2P_LIGHTNING_MESSAGE_TYPE");
+        std::string message_type_env = mt_env ? std::string(mt_env) : std::string();
+        std::optional<std::uint16_t> message_type{std::nullopt};
+
+        if (message_type_env == "PING") {
+            message_type = 18;
+        } else if (message_type_env == "PONG") {
+            message_type = 19;
+        }
```
morehouse: Eventually, what are all the p2p messages this target will handle?

Perhaps it makes sense to directly prepend a supported message type rather than doing "guess and check" here.  There's 64k possible message types, so we get 1000x efficiency gain by always choosing a supported message type.

modules/clightning/module.cpp
```diff
@@ -239,6 +242,45 @@ std::string clightning_des_offer(const std::string_view input) {
...
+    if (msg_type == WIRE_PING) {
+        u16 num_pong_bytes;
+	    u8 *ignored;
+        
+        if (!fromwire_ping(tmpctx, msg, &num_pong_bytes, &ignored)) {
+            clean_tmpctx();
+            return "";
+        }
+
+        if (num_pong_bytes >= 65532) {
```
morehouse: Do some implementations fail in this case, but not others?

modules/clightning/module.cpp
```diff
@@ -239,6 +242,45 @@ std::string clightning_des_offer(const std::string_view input) {
...
+            clean_tmpctx();
+            return "";
+        }
+
+        result << "MSG_TYPE=PING;NUM_PONG_BYTES=" << num_pong_bytes;
+        result << ";IGNORED=" << tal_bytelen(ignored);
```
morehouse, line 273: What is `IGNORED`?

modules/clightning/module.cpp
```diff
@@ -239,6 +242,45 @@ std::string clightning_des_offer(const std::string_view input) {
...
+
+    if (msg_type == WIRE_PONG) {
+	    u8 *ignored;
+
+        if (!fromwire_pong(tmpctx, msg, &ignored)) {
+            clean_tmpctx();
+            return "";
+        }
+
+        result << "MSG_TYPE=PONG;IGNORED=" << tal_bytelen(ignored);
```
morehouse: Also here, what is `IGNORED`?

morehouse (commented):
I'm not sure this target is worthwhile if it only supports ping/pong.  Those are pretty basic messages.

But if other messages will eventually be added, it could make sense to merge this.

driver.cpp
erickcestari: > Eventually, what are all the p2p messages this target will handle?
> 
> Perhaps it makes sense to directly prepend a supported message type rather than doing "guess and check" here. There's 64k possible message types, so we get 1000x efficiency gain by always choosing a supported message type.

Right, we should probably do that as a custom mutator? So the inputs saved on corpora will have the correct version type.

modules/clightning/module.cpp
erickcestari: The others fail too. It's a rule in BOLT1.

`check_ping_make_pong` I will use this function to check since it's the function used on clightning to parse.

```c

bool check_ping_make_pong(const tal_t *ctx, const u8 *ping, u8 **pong)
{
	u16 num_pong_bytes;
	u8 *ignored;

	if (!fromwire_ping(ctx, ping, &num_pong_bytes, &ignored))
		return false;
	tal_free(ignored);

	/* BOLT #1:
	 *
	 * A node receiving a `ping` message:
	 *  - if `num_pong_bytes` is less than 65532:
	 *    - MUST respond by sending a `pong` message, with `byteslen` equal
	 *      to `num_pong_bytes`.
	 *  - otherwise (`num_pong_bytes` is **not** less than 65532):
	 *    - MUST ignore the `ping`.
	 */
	if (num_pong_bytes < 65532) {
		/* BOLT #1:
		 *
		 * A node sending a `pong` message:
		 *   - SHOULD set `ignored` to 0s.
		 *   - MUST NOT set `ignored` to sensitive data such as secrets
		 *     or portions of initialized memory.
		 */
		ignored = tal_arrz(ctx, u8, num_pong_bytes);
		*pong = towire_pong(ctx, ignored);
		tal_free(ignored);
	} else
		*pong = NULL;

	return true;
}
```

modules/clightning/module.cpp
erickcestari: The `ignored` field in ping and pong messages allows arbitrary padding, which can be used to mask traffic patterns and improve network privacy.

https://github.com/lightning/bolts/blob/master/01-messaging.md#the-ping-and-pong-messages

erickcestari:
> I'm not sure this target is worthwhile if it only supports ping/pong. Those are pretty basic messages.
> 
> But if other messages will eventually be added, it could make sense to merge this.

Yes, this is a start point to them add more messages.

erickcestari:
@brunoerg Could you test what happens when building this branch with c-lightning and running the fuzzing on the deserialize_offer and deserialize_invoice targets? In CI, the error that causes execution to exit doesn’t appear.

https://github.com/bitcoinfuzz/bitcoinfuzz/actions/runs/17402174854/job/49398497555?pr=264#step:6:80

driver.cpp
morehouse: Good idea, yes let's use a custom mutator to generate the correct message types.  We could also do the environment variable check there.

modules/clightning/module.cpp
```diff
@@ -239,6 +243,46 @@ std::string clightning_des_offer(const std::string_view input) {
...
+    peer_wire msg_type = (enum peer_wire)fromwire_peektype(msg);
+    std::ostringstream result;
+    
+    if (msg_type == WIRE_PING) {
+        u16 num_pong_bytes;
+	    u8 *ignored;
+        u8 *pong;
+
+        if (!check_ping_make_pong(tmpctx, msg, &pong)) {
+            clean_tmpctx();
```
morehouse: Nit:  There's 4 different places we might do `clean_tmpctx` in this function.  Perhaps it would be worth using an RAII object to automatically do the cleanup on return?  We could reuse this pattern for the other CLN targets as well.

modules/ldk/ldk_lib/src/lib.rs
```diff
@@ -250,6 +255,53 @@ pub unsafe extern "C" fn ldk_des_offer(input: *const std::os::raw::c_char) -> *m
...
+        return str_to_c_string("");
+    }
+
+    let msg_type = u16::from_be_bytes([data[0], data[1]]);
+    let mut cursor = Cursor::new(&data[2..]);
+
+    if msg_type == 18 {
+        match msgs::Ping::read(&mut cursor) {
+            Ok(ping) => {
+                if ping.ponglen >= 65532 {
```
morehouse, line 271: Do we need to do this manual ponglen check for LDK?

modules/lnd/lnd_wrapper/wrapper.go
```diff
@@ -130,4 +137,30 @@ func LndDeserializeInvoice(cInvoiceStr *C.char) *C.char {
 	return C.CString(sb.String())
 }
 
+//export LndParseP2pLightningMessage
+func LndParseP2pLightningMessage(data C.ByteArray) *C.char {
+	buffer := C.GoBytes(unsafe.Pointer(data.data), data.length)
```
morehouse: We package the pointer and size into a `ByteArray`, then we extract the pointer and size here anyway.  It would be simpler to get rid of `ByteArray` entirely and pass the pointer and size directly to this function.

morehouse: Nit: If we can use `unsafe.Slice` here, the line gets slightly simpler.
```suggestion
func LndParseP2pLightningMessage(data C.ByteArray) *C.char {
	buffer := unsafe.Slice(data.data, data.length)
```

modules/ldk/ldk_lib/src/lib.rs
erickcestari: Yes, they do this check later.

https://github.com/lightningdevkit/rust-lightning/blob/52025d18ef2339be7fd739e578488e66d8aded04/lightning/src/ln/peer_handler.rs#L2435

modules/lnd/lnd_wrapper/wrapper.go
erickcestari: We can't really do that because the `C.GoBytes` is the function that will convert the c bytes to go bytes[]

brunoerg:
Can you please rebase?

main.cpp
```diff
@@ -242,6 +242,57 @@ size_t LLVMFuzzerCustomMutator(uint8_t *fuzz_data, size_t size, size_t max_size,
...
+
+    const char* mt_env = std::getenv("P2P_LIGHTNING_MESSAGE_TYPE");
+    std::string message_type_env = mt_env ? std::string(mt_env) : std::string();
+    std::optional<std::uint16_t> message_type{std::nullopt};
+
+    if (message_type_env == "PING") {
+        message_type = 18;
+    } else if (message_type_env == "PONG") {
+        message_type = 19;
+    }
```
morehouse: Nit: perhaps we should use lowercase like in BOLT 1.
```suggestion
    if (message_type_env == "ping") {
        message_type = 18;
    } else if (message_type_env == "pong") {
        message_type = 19;
    }
```

main.cpp
```diff
@@ -242,6 +242,57 @@ size_t LLVMFuzzerCustomMutator(uint8_t *fuzz_data, size_t size, size_t max_size,
...
+                                         unsigned int seed);
+
+size_t LLVMFuzzerCustomMutator(uint8_t *fuzz_data, size_t size, size_t max_size,
+                               unsigned int seed) {
+    // First, mutate the data using LibFuzzer's default mutator
+    size_t new_size = LLVMFuzzerMutate(fuzz_data, size, max_size);
+
+    const char* mt_env = std::getenv("P2P_LIGHTNING_MESSAGE_TYPE");
+    std::string message_type_env = mt_env ? std::string(mt_env) : std::string();
+    std::optional<std::uint16_t> message_type{std::nullopt};
```
morehouse: Can we remove the `std::` prefix for `uint16_t`?

main.cpp
```diff
@@ -242,6 +242,57 @@ size_t LLVMFuzzerCustomMutator(uint8_t *fuzz_data, size_t size, size_t max_size,
...
+
+    if (message_type_env == "PING") {
+        message_type = 18;
+    } else if (message_type_env == "PONG") {
+        message_type = 19;
+    }
+    if (!message_type.has_value()) return new_size;
```
morehouse: I think we should randomly pick a supported message type in this case for efficiency.  Otherwise each module will just return an empty string and the mutation will be wasted.

modules/clightning/module.cpp
```diff
@@ -235,7 +245,44 @@ std::string clightning_des_offer(const std::string_view input) {
...
     }
     
-    clean_tmpctx();
+    return result.str();
+}
+
+std::optional<std::string> clightning_parse_p2p_lightning_message(std::span<const uint8_t> buffer) {
+    CleanTmpCtxGuard _cleanup;
+
+    u8 *msg = (u8 *) tal_arr(tmpctx, u8, buffer.size());
```
morehouse: I think `tal_arr` already does this cast:

```suggestion
    u8 *msg = tal_arr(tmpctx, u8, buffer.size());
```

modules/clightning/module.cpp
```diff
@@ -235,7 +245,44 @@ std::string clightning_des_offer(const std::string_view input) {
...
+    memcpy(msg, buffer.data(), buffer.size());
+    peer_wire msg_type = (enum peer_wire)fromwire_peektype(msg);
```
morehouse: Nit:  we should prefer C++-style casts for the extra compiler checks

```suggestion
    peer_wire msg_type = static_cast<enum peer_wire>(fromwire_peektype(msg));
```

modules/clightning/module.cpp
```diff
@@ -235,7 +245,44 @@ std::string clightning_des_offer(const std::string_view input) {
...
+        if (!pong) {
+            return "";
+        }
+        fromwire_ping(tmpctx, msg, &num_pong_bytes, &ignored);
+
+        result << "MSG_TYPE=PING;NUM_PONG_BYTES=" << num_pong_bytes;
+        result << ";IGNORED=" << tal_bytelen(ignored);
+    }
+
+    if (msg_type == WIRE_PONG) {
```
morehouse:
```suggestion
    } else if (msg_type == WIRE_PONG) {
```

modules/clightning/module.cpp
```diff
@@ -235,7 +245,44 @@ std::string clightning_des_offer(const std::string_view input) {
...
+
+        if (!check_ping_make_pong(tmpctx, msg, &pong)) {
+            return "";
+        }
+        if (!pong) {
+            return "";
+        }
+        fromwire_ping(tmpctx, msg, &num_pong_bytes, &ignored);
+
+        result << "MSG_TYPE=PING;NUM_PONG_BYTES=" << num_pong_bytes;
```
morehouse: Nit:  maybe it would be slightly more readable if the message types were lowercase, so they're distinguished more from the field names

```suggestion
        result << "MSG_TYPE=ping;NUM_PONG_BYTES=" << num_pong_bytes;
```

modules/lnd/lnd_wrapper/wrapper.go
```diff
@@ -130,4 +132,30 @@ func LndDeserializeInvoice(cInvoiceStr *C.char) *C.char {
...
+		sb.WriteString("MSG_TYPE=PING;NUM_PONG_BYTES=")
+		sb.WriteString(fmt.Sprintf("%d", message.(*lnwire.Ping).NumPongBytes))
+		sb.WriteString(";IGNORED=")
+		sb.WriteString(fmt.Sprintf("%d", len(message.(*lnwire.Ping).PaddingBytes)))
+	}
+
+	if message.MsgType() == 19 {
+		sb.WriteString("MSG_TYPE=PONG;IGNORED=")
+		sb.WriteString(fmt.Sprintf("%d", len(message.(*lnwire.Pong).PongBytes)))
+	}
```
morehouse: Nit: the common idiom in Go is a switch:

```go
switch message.MsgType() {
case 18:
  ...
case 19:
  ...
}
```

modules/lnd/module.cpp
```diff
@@ -16,5 +16,16 @@ namespace bitcoinfuzz
             free(result);
             return result_str;
         }
+
+        std::optional<std::string> Lnd::parse_p2p_lightning_message(std::span<const uint8_t> buffer) const
+        {
+            auto result = LndParseP2pLightningMessage((char*)buffer.data(), buffer.size());
```
morehouse: Nit:

```suggestion
            auto result = LndParseP2pLightningMessage(reinterpret_cast<char *>(buffer.data()), buffer.size());
```

main.cpp
erickcestari: Updated!

main.cpp
erickcestari: Updated!

main.cpp
erickcestari: Updated!

modules/clightning/module.cpp
erickcestari: Updated!

modules/clightning/module.cpp
erickcestari: Updated!

modules/clightning/module.cpp
erickcestari: Updated!

modules/clightning/module.cpp
erickcestari: Updated!

modules/lnd/lnd_wrapper/wrapper.go
erickcestari: Updated!

modules/lnd/module.cpp
erickcestari: Updated!

morehouse (approved):
LGTM

PR #265, opened 2025-09-02
feat: add jvmloader to initialize the JVM only 1 time

erickcestari:
This PR doesn’t fully resolve the problem, since starting the JVM with both .JAR libraries can cause one library to shadow another.

Related to: https://github.com/bitcoinfuzz/bitcoinfuzz/issues/266


erickcestari:
Ready for review @brunoerg

brunoerg (approved):
lgtm for now

PR #267, opened 2025-09-03
ci: use smoke run when corpus is missing

erickcestari:


brunoerg:
I'm not sure about it, since it can cause a failure on our CI. Some targets could crash even with few executions.

erickcestari:
> I'm not sure about it, since it can cause a failure on our CI. Some targets could crash even with few executions.

Shouldn't that mean we should fix those targets? Why do they fail so easily?

brunoerg:
> > I'm not sure about it, since it can cause a failure on our CI. Some targets could crash even with few executions.
> 
> Shouldn't that mean we should fix those targets? Why do they fail so easily?

The problem might not be in the targets but in the implementations - See that the first bug we found with bitcoinfuzz was due to the following miniscript policy: `pk()()` which was found with few executions.

erickcestari:
> > > I'm not sure about it, since it can cause a failure on our CI. Some targets could crash even with few executions.
> > 
> > 
> > Shouldn't that mean we should fix those targets? Why do they fail so easily?
> 
> The problem might not be in the targets but in the implementations - See that the first bug we found with bitcoinfuzz was due to the following miniscript policy: `pk()()` which was found with few executions.

I updated the test to only run with two inputs. The main idea is to at least run the target to see if there aren’t any crashes at runtime.

morehouse:
I agree with Bruno -- in CI it's generally better to only test for regressions, and leave the fuzzing-for-bugs to dedicated machines instead.

I think there's not much reason to leave a fuzz target without a corpus for very long anyway?  Once the target is merged, a corpus can quickly be added.

erickcestari:
Fine, I will close this Pr.

PR #268, opened 2025-09-03
doc: add Fuzzing with AFL++

brunoerg:


.github/workflows/workflow.yml
```diff
@@ -227,6 +234,23 @@ jobs:
...
+      - name: Restore cached build artifacts
+        uses: actions/cache/restore@v4
+        with:
+          path: |
+            modules/
+            bitcoinfuzz
+            NBitcoin.CppBridge.so
+            NBitcoin.CppBridge.dylib
+            build.env
+          key: bitcoinfuzz-build-${{ matrix.os }}-${{ github.sha }}-${{ github.run_id }}
```
erickcestari: This should be in the setup-test-env action.yml where we already restore the cached build. Or we may remove the action.yml and add all those step here.

erickcestari (changes requested):

erickcestari:
I just realized that the custom mutators we implemented don’t work when running under AFL++. If we want to use custom mutators with AFL++, we’ll probably need to write AFL++-specific versions.

erickcestari:
Should we remove all the hardcoded CC and CXX flags from the Makefiles, or will compiling with `clang` and `afl-clang-fast` produce the same result?

brunoerg:
> I just realized that the custom mutators we implemented don’t work when running under AFL++. If we want to use custom mutators with AFL++, we’ll probably need to write AFL++-specific versions.

Yes, the custom mutators would not work. Perhaps we could mention it on the documentation as well.

.github/workflows/workflow.yml
brunoerg: Yes, that was a leftover from a test thing. Fixed, thanks.

brunoerg:
> Should we remove all the hardcoded CC and CXX flags from the Makefiles, or will compiling with `clang` and `afl-clang-fast` produce the same result?

We should let users set these flags.

erickcestari:
Shouldn't we remove the CXX flag in all modules?

```
grep -R -F "CXX = clang++" .
./modules/btcd/Makefile:CXX = clang++
./modules/clightning/Makefile:CXX = clang++
./modules/embit/Makefile:CXX = clang++
./modules/ldk/Makefile:CXX = clang++
./modules/lnd/Makefile:CXX = clang++
./modules/nbitcoin/Makefile:CXX = clang++
./modules/lightningkmp/Makefile:CXX = clang++
./modules/eclair/Makefile:CXX = clang++
```

There are also some CC hardcoded too.

```
rg "CC=" --glob '!external/'

modules/bitcoin/Makefile
32:             CC="clang" \

modules/clightning/Makefile
22:     cd $(LIGHTNING_DIR) && ./configure --enable-address-sanitizer --disable-compat --enable-fuzzing --disable-valgrind --enable-debugbuild --enable-ub-sanitizer CC=clang CWARNFLAGS="-Wall -Wundef -Wmissing-prototypes -Wmissing-declarations -Wstrict-prototypes -Wold-style-definition"
```

brunoerg:
> Shouldn't we remove the CXX flag in all modules?
> 
> ```
> grep -R -F "CXX = clang++" .
> ./modules/btcd/Makefile:CXX = clang++
> ./modules/clightning/Makefile:CXX = clang++
> ./modules/embit/Makefile:CXX = clang++
> ./modules/ldk/Makefile:CXX = clang++
> ./modules/lnd/Makefile:CXX = clang++
> ./modules/nbitcoin/Makefile:CXX = clang++
> ./modules/lightningkmp/Makefile:CXX = clang++
> ./modules/eclair/Makefile:CXX = clang++
> ```
> 
> There are also some CC hardcoded too.
> 
> ```
> rg "CC=" --glob '!external/'
> 
> modules/bitcoin/Makefile
> 32:             CC="clang" \
> 
> modules/clightning/Makefile
> 22:     cd $(LIGHTNING_DIR) && ./configure --enable-address-sanitizer --disable-compat --enable-fuzzing --disable-valgrind --enable-debugbuild --enable-ub-sanitizer CC=clang CWARNFLAGS="-Wall -Wundef -Wmissing-prototypes -Wmissing-declarations -Wstrict-prototypes -Wold-style-definition"
> ```

Done.

erickcestari:
https://github.com/AFLplusplus/AFLplusplus/blob/stable/docs/fuzzing_in_depth.md#a-selecting-the-best-afl-compiler-for-instrumenting-the-target

brunoerg:
> https://github.com/AFLplusplus/AFLplusplus/blob/stable/docs/fuzzing_in_depth.md#a-selecting-the-best-afl-compiler-for-instrumenting-the-target

Just added a mention to this in the documentation.

brunoerg:
Basically a documentation addition, we can proceed as is and improve it in a follow-up.

PR #271, opened 2025-09-09
Add new target for transaction parsing

brunoerg:
This PR adds a new target (initially for Bitcoin Core and btcd) that deserializes a transaction, checks consensus rules, and returns the witness hash. The idea in a follow-up is to return more stuff to be compared.


brunoerg:
Corpora to test it is available at: https://github.com/bitcoinfuzz/corpora/pull/26

PR #279, opened 2025-09-23
feat: parse new lightning message open_channel

erickcestari:


modules/lnd/lnd_wrapper/wrapper.go
```diff
@@ -162,6 +162,68 @@ func LndParseP2pLightningMessage(data *C.char, length C.int) *C.char {
 	case 19:
 		sb.WriteString("MSG_TYPE=pong;IGNORED=")
 		sb.WriteString(fmt.Sprintf("%d", len(message.(*lnwire.Pong).PongBytes)))
+	case 32:
+		openChannel := message.(*lnwire.OpenChannel)
+		// Both ChannelReserve, FundingAmount and DustLimit may overflow the int64 type
+		if openChannel.ChannelReserve < 0 || openChannel.FundingAmount < 0 || openChannel.DustLimit < 0 {
+			return nil
```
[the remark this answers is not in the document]
brunoerg: You're incorrect, copilot. 😆

```go
// Amount represents the base bitcoin monetary unit (colloquially referred
// to as a `Satoshi').  A single Amount is equal to 1e-8 of a bitcoin.
type Amount int64
```

modules/ldk/ldk_lib/src/lib.rs
```diff
@@ -316,6 +316,65 @@ pub unsafe extern "C" fn ldk_parse_p2p_lightning_message(
...
+                }
+                if let Some(channel_type) = &common.channel_type {
+                    s.push_str(";CHANNEL_TYPE=");
+                    let mut flags = channel_type.le_flags().to_vec();
+                    flags.reverse();
+                    s.push_str(&flags.to_lower_hex_string());
+                }
+                str_to_c_string(&s)
+            }
+            Err(DecodeError::UnknownRequiredFeature) => std::ptr::null_mut(),
```
brunoerg, line 395: Why does it return null when there is an unknown required feature? Some features might be supported by other implementations but not here?

erickcestari: This is related to TLV where LND has additional optional TLV for simple taproot channels in open_channel message. So when rust-lightning is parsing a TLV with a even type it will return the `UnknownRequiredFeature` error.

Btw, I noticed now that we aren't skipping this case for c-lightning. We might should skip this by checking if in the LND it parsed any of other optional TLVs, if it parsed we skip there instead of skipping in c-lightning and rust-lightning.

```
otherwise, if type is unknown:
if type is even:
MUST fail to parse the tlv_stream.
otherwise, if type is odd:
MUST discard the next length bytes.
```

erickcestari: It’s actually simpler to skip these cases in Core Lightning and Rust-Lightning.

morehouse (approved):
LGTM

brunoerg:
Needs rebase.

PR #280, opened 2025-09-24
feat: parse new lightning message funding_created

erickcestari:


modules/lnd/lnd_wrapper/wrapper.go
```diff
@@ -162,6 +162,31 @@ func LndParseP2pLightningMessage(data *C.char, length C.int) *C.char {
 	case 19:
 		sb.WriteString("MSG_TYPE=pong;IGNORED=")
 		sb.WriteString(fmt.Sprintf("%d", len(message.(*lnwire.Pong).PongBytes)))
+	case 34:
```
morehouse, line 165: There's a lot of casting to `*FundingCreated` below.  We could improve readability by doing the cast one time  here.
```suggestion
	case 34:
		fc := message.(*lnwire.FundingCreated)
```

modules/lnd/lnd_wrapper/wrapper.go
```diff
@@ -162,6 +162,31 @@ func LndParseP2pLightningMessage(data *C.char, length C.int) *C.char {
...
+		_, err := message.(*lnwire.FundingCreated).CommitSig.ToSignature()
+		if err != nil {
+			// Both rust-lightning and c-lightning verify the signature only
+			// when verifying against the public key.
```
morehouse: This comment is confusing to me.  Is it trying to say that LND does some validation that LDK and CLN don't?

erickcestari: Actually LND does not validate signatures during parsing. To avoid many crashes in the target, we can use `.toSignature()`, which performs stricter checks than `secp256k1_ecdsa_signature_parse_compact()` used by rust-lightning and Core Lightning, so we have to skip those additional validations that Core Lightning and rust-lightning don't do. They will validate if the R or S are zero only when verifying the signature against the pubkey and msg.

modules/lnd/lnd_wrapper/wrapper.go
erickcestari: Yes. Fixed

modules/lnd/lnd_wrapper/wrapper.go
```diff
@@ -162,9 +162,45 @@ func LndParseP2pLightningMessage(data *C.char, length C.int) *C.char {
...
 	}
 
 	return C.CString(sb.String())
 }
 
+func validateSignature(sig lnwire.Sig) error {
+	_, err := sig.ToSignature()
+	if err != nil {
+		// Both rust-lightning and c-lightning checks if R and S are 0 when
+		// verifying the signature against the public key.
```
morehouse:
```suggestion
		// rust-lightning and c-lightning only check if R or S are 0 later on,
		// when verifying the signature against the public key.  For
		// compatibility, ignore such errors here.
```

modules/lnd/lnd_wrapper/wrapper.go
```diff
@@ -162,9 +162,45 @@ func LndParseP2pLightningMessage(data *C.char, length C.int) *C.char {
...
+		if err.Error() == "invalid signature: R is 0" ||
+			err.Error() == "invalid signature: S is 0" {
+			return nil
```
morehouse: This ignores the errors for zero-value R or S.  But what if there are also other things wrong with the signature, but the zero-value R or S was detected first?  I'm afraid we may let the invalid signature through in that case, which could cause false positives when comparing to CLN or LDK.

Is it possible to *add* the full R/S validation to CLN and LDK instead?

erickcestari: True. Fixed

morehouse (approved):
LGTM

PR #283, opened 2025-09-26
Parse p2p lightning message funding_signed and channel_ready

erickcestari:


modules/clightning/module.cpp
```diff
@@ -299,6 +299,35 @@ std::optional<std::string> clightning_parse_p2p_lightning_message(std::span<cons
         }
 
         result << "MSG_TYPE=pong;IGNORED=" << tal_bytelen(ignored);
+    } else if (msg_type == WIRE_FUNDING_SIGNED) {
+        channel_id channel;
+        secp256k1_ecdsa_signature signature;
+
+        if (tal_bytelen(msg) > 98) {
+            return std::nullopt;
+        }
```
morehouse, line 348: We should have a comment explaining the reason for this check.

modules/lnd/lnd_wrapper/wrapper.go
```diff
@@ -162,6 +162,49 @@ func LndParseP2pLightningMessage(data *C.char, length C.int) *C.char {
...
+		_, err := message.(*lnwire.FundingSigned).CommitSig.ToSignature()
+		if err != nil {
+			// Both rust-lightning and c-lightning verify the signature only
+			// when verifying against the public key.
+			if err.Error() == "invalid signature: R is 0" ||
+				err.Error() == "invalid signature: S is 0" {
+				return nil
+			}
+			return C.CString("")
+		}
```
morehouse: Let's use the same approach as https://github.com/bitcoinfuzz/bitcoinfuzz/pull/280, adding the validation to LDK and CLN instead.

modules/lnd/lnd_wrapper/wrapper.go
```diff
@@ -162,6 +162,49 @@ func LndParseP2pLightningMessage(data *C.char, length C.int) *C.char {
 	case 19:
 		sb.WriteString("MSG_TYPE=pong;IGNORED=")
 		sb.WriteString(fmt.Sprintf("%d", len(message.(*lnwire.Pong).PongBytes)))
+	case 35:
+		_, err := message.(*lnwire.FundingSigned).CommitSig.ToSignature()
```
morehouse: Let's cast just once at the beginning here.

```suggestion
	case 35:
		fs := message.(*lnwire.FundingSigned)
		_, err := fs.CommitSig.ToSignature()
```

modules/lnd/lnd_wrapper/wrapper.go
```diff
@@ -162,6 +162,49 @@ func LndParseP2pLightningMessage(data *C.char, length C.int) *C.char {
...
+			return nil
+		}
+		// If the message has additional bytes in the TLV that is more than an Alias, skip it
+		// type_msg = 2 bytes
+		// channel_id = 32 bytes
+		// pubkey = 33 bytes
+		// TLV alias = 10 bytes
+		if len(buffer) > 2+32+33+10 && channelReadyMsg.AliasScid != nil {
+			return nil
+		}
```
morehouse: We should add to these comments and explain *why* we skip the extra TLVs.

modules/lnd/lnd_wrapper/wrapper.go
erickcestari: Fixed

modules/lnd/lnd_wrapper/wrapper.go
erickcestari: Updated!

modules/clightning/module.cpp
erickcestari: Added

modules/lnd/lnd_wrapper/wrapper.go
erickcestari: Added

morehouse (approved):
LGTM

Issue #272, opened 2025-09-11
Differential fuzzing of BIP32 key derivations


quapka:
Hi all,
thanks for the interesting project. During Spring, I've stumbled upon a serialization problem of extended keys of depth over 255 (thus not fitting into a single byte and overflowing). The problem occurred in a number of libraries, see https://github.com/bitcoin/bitcoin/issues/32201. Based on this finding I got interested in doing differential fuzzing of BIP32 derivations across wide-range of targets (various languages). Also, from what I've seen the key derivation API is broad-ish, leading to errors like https://github.com/MetacoSA/NBitcoin/issues/1259. This motivates closer look and fuzz test the different key derivation API options.

Currently, together with a student of mine, we are working on a proof of concept fuzzer, but I would like to hear your opinion on whether extending `bitcoinfuzz` would seem like a good fit for you. Our goal is to support wide range of target implementations (from various languages), not just the ones currently present in your project. Possibly, including also open-source hardware wallets firmwares, etc.

brunoerg:
Hi, @quapka. It looks interesting! 

I think we could have a target here in `bitcoinfuzz` for BIP32 key derivations. I don't think you would need to spend time creating another project just for this, seems fine and not hard to write this target here. I think the easiest path is to start by creating the target with the existing modules we have here in `bitcoinfuzz` (e.g. Bitcoin Core, NBitcoin, rust-bitcoin). Then, we could work on integrating other relevant projects here (this usually takes more time) - I'm really interested on having bitcoinj, libbitcoin (which there is an open PR integrating it), and others, here. If you or your student want to work on it here, I'm totally open to help and guide you. If not, I will probably do it but not in short-time since we have many things on our backlog and we're basically no more than 3 people working on this project.

quapka:
Indeed, it sounds better to contribute to an existing project.

I've checked the overall design - could you double check that I got it right?

- [basemodule.cpp](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/include/bitcoinfuzz/basemodule.cpp) is where would would supply new fuzzing entry points for the BIP32 keys, e.g., `extended_pub_key_deserialize`, etc. I can imagine there being several such functions - some specific to only certain libraries.
[...]

PR #287, opened 2025-10-02
Add BIP32 master key gen target for BITCOIN_CORE, RUST_BITCOIN and NBITCOIN

kuliq23:
Based on the discussion we had in [Issue](https://github.com/bitcoinfuzz/bitcoinfuzz/issues/272), I prepared a first basic BIP32 target for master key generation from seed/bytes generated by the fuzzer. This is mainly intended as an example and a place to discuss good/bad practices while contributing to your project. 

I already stumbled upon trouble - LeakSanitizer reports indirect leaks (3768 bytes) regarding NBITCOINs _ExtKey_ at the end of runtime (not per fuzzrun).  See it [here](https://gist.github.com/kuliq23/6de8effc57888ae02aae99a41c46fc17). I believe these are false positives as _ExtKey_ and _Key_  are managed objects and their memory is automatically reclaimed by the GC. Let me know your opinion on how to handle this, thx.

### DIFF
--- a/modules/bitcoin/module.cpp
+++ b/modules/bitcoin/module.cpp
@@ -2,6 +2,7 @@
 #include <span>
 #include <string>
 
+#include "base58.h"
 #include "blockencodings.h"
 #include "chainparams.h"
 #include "consensus/validation.h"
@@ -23,6 +24,7 @@
 #include "key_io.h"
 #include "psbt.h"
 #include "span.h"
+#include "key.h"
 
 namespace {
 class FuzzedSignatureChecker : public BaseSignatureChecker
@@ -505,6 +507,14 @@ std::optional<int> Bitcoin::cmpctblocks_parse(std::span<const uint8_t> buffer) c
         return std::nullopt;
     }
 
+}
+
+std::optional<std::string> Bitcoin::bip32_master_keygen(std::span<const uint8_t> seed) const
+{
+    SelectParams(ChainType::MAIN);
+    CExtKey master;
+    master.SetSeed(std::span{reinterpret_cast<const std::byte*>(seed.data()), seed.size()});
+    return EncodeExtKey(master);
 }
 
 } // namespace module
--- a/modules/nbitcoin/NBitcoin/NBitcoinBridge.cs
+++ b/modules/nbitcoin/NBitcoin/NBitcoinBridge.cs
@@ -93,4 +93,21 @@ public static class Bridge
             return false;
         }
     }
+
+    [UnmanagedCallersOnly(EntryPoint = "nbitcoin_bip32_master_keygen")]
+    public static IntPtr BIP32MasterKeygen(IntPtr dataPtr, UIntPtr len)
+    {
+        var seed = new byte[(int)len];
+        Marshal.Copy(dataPtr, seed, 0, (int)len);
+        ExtKey sk = ExtKey.CreateFromSeed(seed);
+        IntPtr strPtr = Marshal.StringToHGlobalAnsi(sk.GetWif(Network.Main).ToString());
+        return strPtr;
+        
+    }
+
+    [UnmanagedCallersOnly(EntryPoint = "nbitcoin_free_c_string")]
+    public static void FreeString(IntPtr ptr)
+    {
+        if (ptr != IntPtr.Zero) Marshal.FreeHGlobal(ptr);
+    }
 }
--- a/modules/nbitcoin/NBitcoin/nbitcoin_lib.h
+++ b/modules/nbitcoin/NBitcoin/nbitcoin_lib.h
@@ -5,3 +5,7 @@ extern "C" bool nbitcoin_miniscript_parse(const char* input);
 extern "C" bool nbitcoin_descriptor_parse(const char* input);
 
 extern "C" bool nbitcoin_script_eval(const uint8_t* input_data, int32_t input_data_length, uint32_t flags, uint32_t version);
+
+extern "C" char* nbitcoin_bip32_master_keygen(const uint8_t *data, size_t len);
+
+extern "C" void nbitcoin_free_c_string(void* ptr);
--- a/modules/nbitcoin/module.cpp
+++ b/modules/nbitcoin/module.cpp
@@ -19,5 +19,13 @@ namespace bitcoinfuzz
         {
             return nbitcoin_script_eval(input_data.data(), input_data.size(), flags, version);
         }
+        std::optional<std::string> NBitcoin::bip32_master_keygen(std::span<const uint8_t> buffer) const
+        {
+            char* p = nbitcoin_bip32_master_keygen(buffer.data(), buffer.size());
+            if (!p) return std::nullopt;
+            std::string s(p);
+            nbitcoin_free_c_string(p);   
+            return s;
+        }
     }
 }
--- a/modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
+++ b/modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
@@ -5,6 +5,8 @@ use bitcoin::block::BlockUncheckedExt;
 use bitcoin::consensus::{deserialize_partial, encode, serialize};
 use bitcoin::script::{ScriptBuf, ScriptExt};
 use bitcoin::Block;
+use bitcoin::bip32::Xpriv;
+use bitcoin::NetworkKind;
 use p2p::address::AddrV2;
 use p2p::message::{AddrV2Payload, RawNetworkMessage};
 use p2p::Magic;
@@ -249,6 +251,13 @@ pub unsafe extern "C" fn rust_bitcoin_cmpctblocks_parse(data: *const u8, len: us
             return -1;
         }
     }
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn rust_bitcoin_bip32_master_keygen(data: *const u8, len: usize) -> *mut c_char {
+    let seed = slice::from_raw_parts(data, len);
+    let sk = Xpriv::new_master(NetworkKind::Main, &seed);
+    str_to_c_string(&sk.to_string())
 }
 
 unsafe fn c_str_to_str<'a>(input: *const c_char) -> Result<&'a str, Utf8Error> {
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -231,6 +231,10 @@ jobs:
             target: "parse_p2p_lightning_message"
             cxxflags: "-DLDK -DCLIGHTNING -DLND -DCUSTOM_MUTATOR_P2P_LIGHTNING_MESSAGE"
             asan_options: "detect_leaks=0:detect_container_overflow=0"
+          - corpus: "bip32_master_keygen"
+            target: "bip32_master_keygen"
+            cxxflags: "-DNBITCOIN -DRUST_BITCOIN -DBITCOIN_CORE"
+            asan_options: "detect_leaks=0:detect_container_overflow=0"
 
     steps:
       - name: Checkout repo
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -160,3 +160,14 @@ services:
       FUZZ: parse_p2p_lightning_message
     volumes:
       - ./docker/parse_p2p_lightning_message:/app/data
+
+  bip32_master_keygen:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    environment:
+      <<: *common-env
+      CXXFLAGS: "-DRUST_BITCOIN -DNBITCOIN -DBITCOIN_CORE"
+      FUZZ: bip32_master_keygen
+    volumes:
+      - ./docker/bip32_master_keygen:/app/data
--- a/driver.cpp
+++ b/driver.cpp
@@ -417,6 +417,30 @@ namespace bitcoinfuzz
             last_module_name = module.first;
         }
     }
+        void Driver::Bip32MasterKeygenTarget(std::span<const uint8_t> buffer) const
+    {
+        std::optional<std::string> last_response{std::nullopt};
+        std::string last_module_name;
+
+        for (auto &module : modules)
+        {
+            std::optional<std::string> res{module.second->bip32_master_keygen(buffer)};
+            if (!res.has_value()) continue;
+            if (last_response.has_value()) {
+                if (*res != *last_response) {
+                    std::cout << "BIP32 master keygen failed" << std::endl;
+                    std::cout << "Module: " << module.first << std::endl;
+                    std::cout << "Result: " << *res << std::endl;
+                    std::cout << "Module: " << last_module_name << std::endl;
+                    std::cout << "Result: " << *last_response << std::endl;
+                }
+                assert(*res == *last_response);
+            }
+
+            last_response = res.value();
+            last_module_name = module.first;
+        }
+    }
 
     void Driver::Run(const uint8_t *data, const size_t size, const std::string &target) const
     {
@@ -451,6 +475,8 @@ namespace bitcoinfuzz
             this->ParseLightningP2pMessageTarget(buffer);
         } else if (target == "transaction_eval") {
             this->TransactionEvalTarget(buffer);
+        } else if (target == "bip32_master_keygen") {
+            this->Bip32MasterKeygenTarget(buffer);
         } else {
             std::cout << "Target not defined!" << std::endl;
             assert(false);
--- a/driver.h
+++ b/driver.h
@@ -38,5 +38,6 @@ namespace bitcoinfuzz
         void ParseP2PMessageTarget(std::span<const uint8_t> buffer) const;
         void ParseLightningP2pMessageTarget(std::span<const uint8_t> buffer) const;
         void TransactionEvalTarget(std::span<const uint8_t> buffer) const;
+        void Bip32MasterKeygenTarget(std::span<const uint8_t> buffer) const;
     };
 }
--- a/include/bitcoinfuzz/basemodule.cpp
+++ b/include/bitcoinfuzz/basemodule.cpp
@@ -82,4 +82,10 @@ namespace bitcoinfuzz
     {
         return std::nullopt;
     }
+    
+    std::optional<std::string> BaseModule::bip32_master_keygen(std::span<const uint8_t> buffer) const
+    {
+        return std::nullopt;
+    }
+
 }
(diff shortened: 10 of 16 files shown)


brunoerg (commented):
I took a quick look at this and looks great overall, it's a very simple target so it shouldn't be too complicated btw. I'll take a better look tomorrow.

modules/nbitcoin/NBitcoin/nbitcoin_lib.h
```diff
@@ -2,4 +2,8 @@
 
 extern "C" bool nbitcoin_miniscript_parse(const char* input);
 
-extern "C" bool nbitcoin_descriptor_parse(const char* input);
```
quapka: This supposedly removed line is possibly due to it being the last line of a file. AFAICT, there's only been an addition to `nbitcoin_lib.h`.

kuliq23: Confirmed, that is not a real deletion. thx for pointing that out

modules/nbitcoin/module.cpp
```diff
@@ -15,5 +15,13 @@ namespace bitcoinfuzz
         {
             return nbitcoin_descriptor_parse(str.c_str());
         }
+        std::optional<std::string> NBitcoin::bip32_master_keygen(std::span<const uint8_t> buffer) const
+        {
+            char* p = nbitcoin_bip32_master_keygen(buffer.data(), buffer.size());
+            if (p == nullptr) return std::nullopt;
+            std::string s(p);
+            nbitcoin_free_c_string((void*)p);
```
brunoerg: Why do we need this `(void*)` cast?

modules/nbitcoin/module.cpp
```diff
@@ -15,5 +15,13 @@ namespace bitcoinfuzz
         {
             return nbitcoin_descriptor_parse(str.c_str());
         }
+        std::optional<std::string> NBitcoin::bip32_master_keygen(std::span<const uint8_t> buffer) const
+        {
+            char* p = nbitcoin_bip32_master_keygen(buffer.data(), buffer.size());
+            if (p == nullptr) return std::nullopt;
```
brunoerg: nit:
```suggestion
            if (!p) return std::nullopt;
```

modules/nbitcoin/NBitcoin/NBitcoinBridge.cs
```diff
@@ -58,4 +58,21 @@ private static bool TryParseMiniscript(string miniscript, KeyType keyType)
             return false;
         }
     }
+
+    [UnmanagedCallersOnly(EntryPoint = "nbitcoin_bip32_master_keygen")]
+    public static IntPtr BIP32MasterKeygen(IntPtr dataPtr, UIntPtr len)
```
brunoerg, line 98: I haven't tested but maybe it could be something like:

```csharp
[UnmanagedCallersOnly(EntryPoint = "nbitcoin_bip32_master_keygen")]
public static IntPtr BIP32MasterKeygen(IntPtr dataPtr, UIntPtr len)
{
    var seed = new Span<byte>(dataPtr.ToPointer(), (int)len);
    ExtKey sk = ExtKey.CreateFromSeed(seed.ToArray());
    return Marshal.StringToHGlobalAnsi(sk.GetWif(Network.Main).ToString());
}
```

This way we would avoid the allocation.

brunoerg (changes requested):
You must put this new target into the CI/CD and the Docker setup, btw.

modules/nbitcoin/module.cpp
kuliq23: Good catch, removed

modules/nbitcoin/NBitcoin/NBitcoinBridge.cs
kuliq23: This sadly does not help. Playing around, i found out both C# libs have similar leakage already. Referencing this in an Issue.

erickcestari:
Hi, I tested running the target on my Arch Linux machine and couldn't reproduce the memory leaks you reported. This might be an environment issue?

My setup:

```
$ uname -r
6.17.1-arch1-1

$ dotnet --version
9.0.110

$ openssl version
OpenSSL 3.6.0 1 Oct 2025 (Library: OpenSSL 3.6.0 1 Oct 2025)

$ clang --version
clang version 20.1.8
Target: x86_64-pc-linux-gnu
```

> I already stumbled upon trouble - LeakSanitizer reports indirect leaks (3768 bytes) regarding NBITCOINs ExtKey at the end of runtime (not per fuzzrun). See it [here](https://gist.github.com/kuliq23/6de8effc57888ae02aae99a41c46fc17). I believe these are false positives as ExtKey and Key are managed objects and their memory is automatically reclaimed by the GC. Let me know your opinion on how to handle this, thx.

brunoerg:
Can you please rebase it? It should fix the CI failure.

kuliq23:
> Hi, I tested running the target on my Arch Linux machine and couldn't reproduce the memory leaks you reported. This might be an environment issue?
> 
> My setup:
> 
> ```
> $ uname -r
> 6.17.1-arch1-1
> 
> $ dotnet --version
> 9.0.110
> 
> $ openssl version
> OpenSSL 3.6.0 1 Oct 2025 (Library: OpenSSL 3.6.0 1 Oct 2025)
> 
> $ clang --version
> clang version 20.1.8
> Target: x86_64-pc-linux-gnu
> ```
> 
> > I already stumbled upon trouble - LeakSanitizer reports indirect leaks (3768 bytes) regarding NBITCOINs ExtKey at the end of runtime (not per fuzzrun). See it [here](https://gist.github.com/kuliq23/6de8effc57888ae02aae99a41c46fc17). I believe these are false positives as ExtKey and Key are managed objects and their memory is automatically reclaimed by the GC. Let me know your opinion on how to handle this, thx.

Hi, thanks for the comment. Interesting, that your system hasn't been able to reproduce the leaks.
My setup:
```
$ uname -r

6.11.0-24-generic


$ dotnet --version

9.0.110


$ openssl version

OpenSSL 3.0.13 30 Jan 2024 (Library: OpenSSL 3.0.13 30 Jan 2024)


$ clang --version

Ubuntu clang version 18.1.3 (1ubuntu1)
Target: x86_64-pc-linux-gnu


```

I'll try to play around with some version combinations later to figure this out.



I also just added docker support for the new target and the leaks show up in Docker too... I see Dockerimage e.g. uses clang18 compared to your clang20. 

One interesting note: The leaks go from 3768 bytes to just 88 bytes (remains just the first reported [here](https://gist.github.com/kuliq23/6de8effc57888ae02aae99a41c46fc17)) depending on the number of Fuzzs : 
* 1-903 runs - 3768B in 41 allocs
* 904+ - 88B in 1 alloc

My guess is the runtime does not have enough time to clean up after itself when number of runs is limited.

erickcestari:
> OpenSSL 3.6.0 1 Oct 2025 (Library: OpenSSL 3.6.0 1 Oct 2025)

These leaks appear to be coming from OpenSSL (libcrypto.so). Updating the library might help.

```
Indirect leak of 112 byte(s) in 1 object(s) allocated from:
    #0 0x6474845667b3 in malloc (/home/ondrej/bitcoinfuzz/bitcoinfuzz+0x2e57b3) (BuildId: 293f26ae807fce3db777af8a1de629200015bf61)
    #1 0x79406da237c4 in CRYPTO_zalloc (/lib/x86_64-linux-gnu/libcrypto.so.3+0x2237c4) (BuildId: 0698e1ff610cb3c6993dccbd82c1281b1b4c5ade)
    #2 0x79406d9e4127  (/lib/x86_64-linux-gnu/libcrypto.so.3+0x1e4127) (BuildId: 0698e1ff610cb3c6993dccbd82c1281b1b4c5ade)
```

modules/nbitcoin/NBitcoin/NBitcoinBridge.cs
brunoerg: Sorry, this would be just an improvement, not supposed to fix that leak.

kuliq23:
```
$ openssl version
OpenSSL 3.6.0 1 Oct 2025 (Library: OpenSSL 3.6.0 1 Oct 2025)
```

Updated OpenSSL, it uses the new 3.6 cryptomodules but still leaks (the same). Atleast the report is simpler:) See it [here](https://gist.github.com/kuliq23/053b09257895424dd7d45d2df9b0a54b)

Anyways,
I'd like to move my work forward and would hate to be stuck on this rather annoying issue. I see 3 possible options for this particular PR and would love to hear your takes.

* 1. I get rid of (atleast for now) NBitcoin and replace it with another module for this target (perhaps bitcoincore) and update the PR.

* 2. The PR gets accepted as is and this leaking issue is resolved in a separate Issue/PR (as leaks like this happen in other C# targets too. Example [here](https://gist.github.com/kuliq23/31e93610e77dbb313468d224ae04c884) ). 


* 3. This PR remains unresolved until the leak issue is solved and I move on to another target, perhaps with a different module also (been playing with custom mutators too).

Please let me know which option you think is the best.
PS: Thanks for fast and thoughtful cooperation, I appreciate it!

brunoerg:
@kuliq23 Keep it as-is and we can investigate this leak issue on another issue/PR. Don't worry on it, we can move on having this target.

modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
```diff
@@ -251,6 +253,16 @@ pub unsafe extern "C" fn rust_bitcoin_cmpctblocks_parse(data: *const u8, len: us
     }
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn rust_bitcoin_bip32_master_keygen(data: *const u8, len: usize) -> *mut c_char {
+    let seed = slice::from_raw_parts(data, len);
+    let sk = Xpriv::new_master(NetworkKind::Main, &seed);
+    if sk.to_string().is_empty() {
+        return std::ptr::null_mut();
+    }
```
erickcestari: Should we keep the check for empty strings ("")? We could probably remove this conditional. We should minimize skipping results whenever possible.

erickcestari (changes requested):

modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
quapka: Agreed, I also don't see a reason why to differentiate.

kuliq23: That is a good point. Check removed.

kuliq23:
Added a third module (bitcoin core) in the meantime. Is the CICD and Docker integration ok as is?

modules/bitcoin/module.cpp
```diff
@@ -514,5 +516,17 @@ std::optional<int> Bitcoin::cmpctblocks_parse(std::span<const uint8_t> buffer) c
...
 
+std::optional<std::string> Bitcoin::bip32_master_keygen(std::span<const uint8_t> seed) const
+{
+    try {
+        SelectParams(ChainType::MAIN);
+        CExtKey master;
+        master.SetSeed(std::span{reinterpret_cast<const std::byte*>(seed.data()), seed.size()});
+        return EncodeExtKey(master);
+    } catch (const std::exception&) {
+        return std::nullopt;
```
erickcestari: Should we skip this case? I think we can just return an empty string instead, so we can identify discrepancies where other implementations generate the master key but Bitcoin Core doesn't.

erickcestari (changes requested):

modules/bitcoin/module.cpp
quapka: That's similar to what you've pointed out in lib.rs, right? Feel free to correct me, but I guess these patterns stem from misunderstanding of how the driver.cpp interprets the results.

If you return a `null` pointer, @kuliq23, then [this line](https://github.com/kuliq23/bitcoinfuzz/blob/bip32masterkeygen/driver.cpp#L423) in driver.cpp _skips_ the module (for the keygen target) completely and does not compare the module result to the rest, aka does not check differences with other modules. I would use the `null` pointer when, for example, for good reasons you:
- cannot construct valid inputs for the target function within a given module, though most should be accounted for in the mutators,
- you intentionally wanna create an output, e.g., due to compatibility and known discrepancies, [see](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/cb11ee224357ad16c54d854805921a4ed7687700/modules/ldk/ldk_lib/src/lib.rs#L140-L145).

In general, as Erick mentioned, you want to "force" the modules to return _a value_ as much as possible. Only then you have a value to compare against the rest of the values from other modules.

Returning an empty value is somewhat arbitrary and can be a limitation. E.g., imagine that two modules do two kinds of checks within the target function and one module fails on the first check, the other module on the second check. If you return an empty value in all these checks, it would seem that the modules behave the same (the driver would not report any discrepancies), but it would be incorrect. I am thinking that in such a case I would return a bit more descriptive error, e.g., `FAILED_CHECK_A` and `FAILED_CHECK_B`.

erickcestari: > That's similar to what you've pointed out in lib.rs, right? Feel free to correct me, but I guess these patterns stem from misunderstanding of how the driver.cpp interprets the results.
> 
> If you return a `null` pointer, @kuliq23, then [this line](https://github.com/kuliq23/bitcoinfuzz/blob/bip32masterkeygen/driver.cpp#L423) in driver.cpp _skips_ the module (for the keygen target) completely and does not compare the module result to the rest, aka does not check differences with other modules. I would use the `null` pointer when, for example, for good reasons you:
> 
> * cannot construct valid inputs for the target function within a given module, though most should be accounted for in the mutators,
> * you intentionally wanna creating an output, e.g., due to compatibility and known discrepancies, [see](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/cb11ee224357ad16c54d854805921a4ed7687700/modules/ldk/ldk_lib/src/lib.rs#L140-L145).
> 
> In general, as Erick mentioned, you want to "force" the modules to return _a value_ as much as possible. Only then you have a value to compare against the rest of the values from other modules.
> 
> Returning an empty value is somewhat arbitrary and can be a limitation. E.g., imagine that two modules do two kinds of checks within the target function and one module fails on the first check, the other module on the second check. If you return an empty value in all these checks, it would seem that the modules behave the same (the driver would not report any discrepancies), but it would be incorrect. I am thinking that in such a case I would return a bit more descriptive error, e.g., `FAILED_CHECK_A` and `FAILED_CHECK_B`.

Yes, you understood correctly. The idea of returning specific error strings might work for small, less complex targets. However, because different implementations have different validations processes, it's difficult to find an implementation that follows the same validation sequence in the same way. That's why we generally just return an empty string.

quapka: Yes, different error messages are nuanced, possibly more fragile. For one, it needs to be verified that an empty string is _not_ a valid output for the target in question. Anyways, one step at a time.

Anyways, looking at bitcoin [test cases](https://github.com/bitcoin/bitcoin/blob/7d27af98c7cf858b5ab5a02e64f89a857cc53172/src/test/bip32_tests.cpp#L125-L141), does there need to be a try/catch at all? If yes, then I would limit it only to the relevant part. I suppose the libFuzzer can handle any unexpected crashes. But I might be mistaken, not having worked with it prior.

kuliq23: I didn't realize i left the check in the new module also, oops... But, based on @quapka's comments and on the nature of this target - i reckon there won't be much crashing with this one - I am inclined to remove the catch block completely as it would provide us with the most information.  (Just ran it without it for 5 minutes and no problem) But if sending a "" is more reasonable in your eyes, then I'll do it that way:)

erickcestari: > I didn't realize i left the check in the new module also, oops... But, based on @quapka's comments and on the nature of this target - i reckon there won't be much crashing with this one - I am inclined to remove the catch block completely as it would provide us with the most information. (Just ran it without it for 5 minutes and no problem) But if sending a "" is more reasonable in your eyes, then I'll do it that way:)

I think removing the catch block is completely fine if the function isn't designed to throw exceptions. If an exception does occur, it's better to let the fuzzer crash so we can properly analyze and report the issue.

erickcestari:
> Added a third module (bitcoin core) in the meantime. Is the CICD and Docker integration ok as is?

Yes, it's looking good!

brunoerg:
Can you please squash the commits?

erickcestari:
Can you rebase? @kuliq23

erickcestari (approved):
Very good! Thanks.

ACK 5e6048b325d8db046171febd7d940ad727f22dc8

brunoerg (approved):
ACK 5e6048b325d8db046171febd7d940ad727f22dc8

brunoerg:
Please consider adding a corpus for this - https://github.com/bitcoinfuzz/corpora

kuliq23:
> Please consider adding a corpus for this - https://github.com/bitcoinfuzz/corpora

Did I get the workflow of adding a corpus right?

1. Fuzz the target like so in Docker 
`docker run -e FUZZ=<target_name> -e CXXFLAGS="-DLDK -DLND ..." -v $(pwd)/corpus:/app/data bitcoinfuzz`
2. Feed the generated corpus to the [clean_corpus.py script](https://github.com/bitcoinfuzz/corpora/blob/main/clean_corpus.py)
3. Open a PR to add the filtered corpus to https://github.com/bitcoinfuzz/corpora

erickcestari:
> > Please consider adding a corpus for this - https://github.com/bitcoinfuzz/corpora
> 
> Did I get the workflow of adding a corpus right?
> 
> 1. Fuzz the target like so in Docker
>    `docker run -e FUZZ=<target_name> -e CXXFLAGS="-DLDK -DLND ..." -v $(pwd)/corpus:/app/data bitcoinfuzz`
> 2. Feed the generated corpus to the [clean_corpus.py script](https://github.com/bitcoinfuzz/corpora/blob/main/clean_corpus.py)
> 3. Open a PR to add the filtered corpus to https://github.com/bitcoinfuzz/corpora

The 2nd step isn't really necessary. We use `clean_corpus.py` to remove inputs that make the differential fuzzing crash (find discrepancies), so when we run our CI it won't crash due to known bugs. That's why we separate the corpus with a suffix (_clean) - the cleaned corpus won't make bitcoinfuzz crash.

You can see some known bugs here: https://github.com/bitcoinfuzz/bitcoinfuzz/wiki/Modules,-targets-and-known-bugs

PR #305, opened 2025-10-21
ldk: use git commit instead of a fixed version

erickcestari:
When building with AddressSanitizer and coverage instrumentation flags in RUSTFLAGS, Cargo attempts to apply these flags to all crates including procedural macros like `lightning-macros`. This causes build failures because procedural macros run at compile-time and cannot be instrumented with runtime sanitizers.

As documented in the Rust Unstable Book:
"Use of sanitizers together with build scripts and procedural macros is technically possible, but in almost all cases it would be best avoided... In more practical terms when using cargo always remember to pass `--target` flag, so that rustflags will not be applied to build scripts and procedural macros."

By specifying an explicit --target (even when it matches the host architecture), Cargo creates a separation:
- Host artifacts (proc macros): Built without instrumentation
- Target artifacts (library): Built WITH ASAN + coverage instrumentation

The Makefile now detects the platform and passes the appropriate --target flag, fixing "can't find crate for `lightning_macros`" errors while maintaining the fuzzing instrumentation needed for differential fuzzing.

Reference: https://doc.rust-lang.org/stable/unstable-book/compiler-flags/sanitizer.html

### DIFF
--- a/modules/ldk/Makefile
+++ b/modules/ldk/Makefile
@@ -1,10 +1,13 @@
 all: module.a
 
 CXXFLAGS += -Wall -Wextra -fsanitize=address,fuzzer -std=c++20 -I ../../include
-LDK_LIB_PATH := ./ldk_lib/target/release/libldk_lib.a
+
+TARGET := $(shell rustc --print host-tuple)
+
+LDK_LIB_PATH := ./ldk_lib/target/$(TARGET)/release/libldk_lib.a
 
 cargo:
-	cd ldk_lib && cargo +nightly build --release
+	cd ldk_lib && cargo +nightly build --release --target $(TARGET)
 
 module.a: module.o $(LDK_LIB_PATH)
 	bash ../../merge.sh module.a $(LDK_LIB_PATH) module.o
--- a/modules/ldk/ldk_lib/Cargo.lock
+++ b/modules/ldk/ldk_lib/Cargo.lock
@@ -26,9 +26,9 @@ checksum = "d965446196e3b7decd44aa7ee49e31d630118f90ef12f97900f262eb915c951d"
 
 [[package]]
 name = "bitcoin"
-version = "0.32.5"
+version = "0.32.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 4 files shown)


morehouse:
A couple questions:

- Why didn't we hit this issue before?
- Why does using a specific commit instead of a released version help?

erickcestari:
> A couple questions:
> 
> * Why didn't we hit this issue before?

This error of lightning-macros only appears after this commit:

https://github.com/lightningdevkit/rust-lightning/commit/16119967df14b47680fae4a338c79168d83d78e5#diff-3d3d4a27081502c9f48188d52039253857c9e960dbe5ddb99ad7737c9e8a7a7e

So none of the releases until beta 0.2.0 will have this issue.

We had actually encountered this issue previously, but fixing it wasn't as simple as changing one line. When we tried compiling with instrumentation, lightning-macros started displaying errors/warnings. https://github.com/bitcoinfuzz/bitcoinfuzz/pull/263

> * Why does using a specific commit instead of a released version help?

Using specific commits allows us to test unreleased features and potentially discover bugs before they reach production. Per example, rust-lightning didn't have a beta 0.2.0 release package available 8 days ago, so referencing a git commit was the only way to test the new version.

modules/ldk/Makefile
```diff
@@ -1,10 +1,26 @@
...
+    TARGET := x86_64-unknown-linux-gnu
+endif
+
+ifeq ($(UNAME_S),Darwin)
+    ifeq ($(UNAME_M),arm64)
+        TARGET := aarch64-apple-darwin
+    else
+        TARGET := x86_64-apple-darwin
+    endif
+endif
```
morehouse: There are infinitely more targets than this, and this approach will automatically misclassify any Linux system that isn't x86_64.

Can we use `rustc --version --verbose` to directly determine the host target instead?

erickcestari: True, fixed using the `rustc --print host-tuple` command which gives the host target directly.

morehouse (approved):
LGTM

brunoerg:
ACK cc048bd5e3a7864c4fce60b1c17ebc1100e30920

PR #307, opened 2025-10-24
BIP32: Add Deserialize extended key target

kuliq23:
### NEW TARGET: Deserialization of a BIP32 extended key
This target accepts a Bip32 extended key in the base58check-encoded form (string of b58 chars), deserializes it, then returns it as a string (return may change to bool, or different parts of the key).
In order to make it even remotely possible to send a valid extkey, formatting of the random bytes is needed.

TLDR bip32 ExtKey serialization from [https://en.bitcoin.it/wiki/BIP_0032](https://en.bitcoin.it/wiki/BIP_0032)
> Extended public and private keys are serialized as follows:

> * 4 byte: version bytes (mainnet: 0x0488B21E public, 0x0488ADE4 private; testnet: 0x043587CF public, 0x04358394 private)
> * 1 byte: depth: 0x00 for master nodes, 0x01 for level-1 derived keys, ....
> * 4 bytes: the fingerprint of the parent's key (0x00000000 if master key)
> * 4 bytes: child number. This is ser32(i) for i in xi = xpar/i, with xi the key being serialized. (0x00000000 if master key)
> * 32 bytes: the chain code
> * 33 bytes: the public key or private key data (serP(K) for public keys, 0x00 || ser256(k) for private keys)

> This 78 byte structure can be encoded like other Bitcoin data in Base58, by first adding 32 checksum bits (derived from the double SHA-256 checksum), and then converting to the Base58 representation. This results in a Base58-encoded string of up to 112 characters. Because of the choice of the version bytes, the Base58 representation will start with "xprv" or  "xpub" on mainnet, "tprv" or "tpub" on testnet.

My idea was to create multiple custom mutators, each "helping the target" a bit more, to maximize the range tested.
(e.g. Mut1 - encode random bytes to b58, Mut2 - hardcode the first 4 version bytes + Mut1, Mut3 - fill to 78 bytes + Mut2, Mut4 - append correct checksum + Mut3, ...)

#### 1. Base58 utils access
For these mutations, base58 functions are needed, so far i only added a placeholder mutator that creates seemingly b58 looking, 78B long data.  I see 3 solutions:
* Custom implementation in main.cpp (similarly to bitcoin_double_sha256() in main.cpp)
* "Borrow" it from Core "modules/bitcoin/base58.h"
* Add the base58 utils to "modules/custommutator/" (similarly to how bech32 and sha256 are there) <- optimal?

#### 2. How mutators work
Please correct me if i am wrong:

As mutators are selected in the building process, running the fuzzer means that only an explicit set of them can be active for that running instance at a time....
Does that mean that if we e.g. want to fuzz "long-term", multiple instances of bitcoinfuzz need to be running to cover all targets? (each with a specified mutator?)
Is this something to avoid (shall I reavaluate the multi-mutator solution and instead e.g. implement multiple targets that begin with mutating the input?

Thanks.

### DIFF
--- a/custommutator/mutators/extended_key.h
+++ b/custommutator/mutators/extended_key.h
@@ -1,0 +1,82 @@
+#ifndef BITCOINFUZZ_CUSTOMMUTATOR_EXTENDED_KEY_H
+#define BITCOINFUZZ_CUSTOMMUTATOR_EXTENDED_KEY_H
+#include <array>
+#include <cstdint>
+#include <cstdlib>
+#include <cstring>
+#include <custommutator/utils/base58.h>
+#include <span>
+#include <string>
+#include <string_view>
+#include <vector>
+extern "C" size_t LLVMFuzzerMutate(uint8_t *Data, size_t Size, size_t MaxSize);
+constexpr size_t VERSION_PREFIX_LEN = 4;
+constexpr size_t BIP32_PAYLOAD_LEN = 78;
+struct VersionBytes {
+  uint8_t data[VERSION_PREFIX_LEN];
+};
+constexpr std::array<VersionBytes, 4> VERSION_PREFIXES = {{
+    {{0x04, 0x88, 0xB2, 0x1E}}, // xpub
+    {{0x04, 0x88, 0xAD, 0xE4}}, // xprv
+    {{0x04, 0x35, 0x87, 0xCF}}, // tpub
+    {{0x04, 0x35, 0x83, 0x94}}, // tprv
+}};
+constexpr size_t IDX_XPUB = 0;
+constexpr size_t IDX_XPRV = 1;
+constexpr size_t IDX_TPUB = 2;
+constexpr size_t IDX_TPRV = 3;
+const VersionBytes &GetVersionBytes(std::string_view key_type) {
+  if (key_type == "xprv")
+    return VERSION_PREFIXES[IDX_XPRV];
+  if (key_type == "tpub")
+    return VERSION_PREFIXES[IDX_TPUB];
+  if (key_type == "tprv")
+    return VERSION_PREFIXES[IDX_TPRV];
+  return VERSION_PREFIXES[IDX_XPUB];
+}
+const VersionBytes &GetRandomVersion(unsigned int seed) {
+  return VERSION_PREFIXES[seed % VERSION_PREFIXES.size()];
+}
+extern "C" size_t LLVMFuzzerCustomMutator(uint8_t *fuzz_data, size_t size,
+                                          size_t max_size, unsigned int seed) {
+
+  // Decode input from base58check
+  std::string_view input{reinterpret_cast<const char *>(fuzz_data), size};
+  std::vector<unsigned char> payload(BIP32_PAYLOAD_LEN, 0x00);
+
+  std::vector<unsigned char> decoded;
+  if (DecodeBase58Check(std::string{input}, decoded,
+                        static_cast<int>(max_size))) {
+    const size_t copy_len = std::min(decoded.size(), BIP32_PAYLOAD_LEN);
+    std::memcpy(payload.data(), decoded.data(), copy_len);
+  } else {
+    const size_t copy_len = std::min(size, BIP32_PAYLOAD_LEN);
+    std::memcpy(payload.data(), fuzz_data, copy_len);
+  }
+
+  // Mutate the payload
+  size_t mutated_size =
+      LLVMFuzzerMutate(payload.data(), BIP32_PAYLOAD_LEN, BIP32_PAYLOAD_LEN);
+
+  // Ensure we have exactly BIP32_PAYLOAD_LEN bytes
+  if (mutated_size < BIP32_PAYLOAD_LEN) {
+    std::memset(payload.data() + mutated_size, 0,
+                BIP32_PAYLOAD_LEN - mutated_size);
+  }
+  payload.resize(BIP32_PAYLOAD_LEN);
+
+  // Apply version prefix: use env if set, otherwise random
+  const char *env = std::getenv("EXTKEYTYPE");
+  const auto &version = env ? GetVersionBytes(env) : GetRandomVersion(seed);
+  std::memcpy(payload.data(), version.data, VERSION_PREFIX_LEN);
+
+  // Encode back to base58check
+  std::string encoded =
+      EncodeBase58Check(std::span<const unsigned char>{payload});
+
+  // Copy result to output buffer
+  const size_t output_len = std::min(encoded.size(), max_size);
+  std::memcpy(fuzz_data, encoded.data(), output_len);
+  return output_len;
+}
+#endif // BITCOINFUZZ_CUSTOMMUTATOR_EXTENDED_KEY_H
--- a/modules/nbitcoin/NBitcoin/NBitcoinBridge.cs
+++ b/modules/nbitcoin/NBitcoin/NBitcoinBridge.cs
@@ -3,6 +3,7 @@ using System.Text;
 using NBitcoin.Scripting;
 using NBitcoin.WalletPolicies;
 
+
 namespace NBitcoin.CppBridge;
 
 public static class Bridge
@@ -170,6 +171,58 @@ public static class Bridge
         ExtKey sk = ExtKey.CreateFromSeed(seed);
         IntPtr strPtr = Marshal.StringToHGlobalAnsi(sk.GetWif(Network.Main).ToString());
         return strPtr;
+    }
+
+    [UnmanagedCallersOnly(EntryPoint = "nbitcoin_bip32_deserialize_extended_key")]
+    public static IntPtr BIP32DeserializeExtendedKeyTarget(IntPtr inputPtr)
+    {
+        if (inputPtr == IntPtr.Zero) return Marshal.StringToCoTaskMemUTF8("INVALID");
+
+        string input = Marshal.PtrToStringUTF8(inputPtr) ?? "";
+        string result;
+
+        if (TryParseXprv(input, Network.Main, out result) ||
+            TryParseXprv(input, Network.TestNet, out result) ||
+            TryParseXpub(input, Network.Main, out result) ||
+            TryParseXpub(input, Network.TestNet, out result))
+        {
+            return Marshal.StringToCoTaskMemUTF8(result);
+        }
+
+        return Marshal.StringToCoTaskMemUTF8("INVALID");
+    }
+
+    // Helpers
+    private static string Hex(byte[] data) => Convert.ToHexString(data).ToLower();
+
+    private static bool TryParseXprv(string input, Network network, out string result)
+    {
+        result = null;
+        try
+        {
+            var ext = NBitcoin.ExtKey.Parse(input, network);
+            result = $"depth={ext.Depth:x2};fp={Hex(ext.ParentFingerprint.ToBytes())};child={ext.Child:x8};chaincode={Hex(ext.ChainCode)};key={ext.PrivateKey.ToHex().ToLower()}";
+            return true;
+        }
+        catch
+        {
+            return false;
+        }
+    }
+
+    private static bool TryParseXpub(string input, Network network, out string result)
+    {
+        result = null;
+        try
+        {
+            var ext = NBitcoin.ExtPubKey.Parse(input, network);
+            result = $"depth={ext.Depth:x2};fp={Hex(ext.ParentFingerprint.ToBytes())};child={ext.Child:x8};chaincode={Hex(ext.ChainCode)};key={ext.PubKey.ToHex().ToLower()}";
+            return true;
+        }
+        catch
+        {
+            return false;
+        }
     }
 
     [UnmanagedCallersOnly(EntryPoint = "nbitcoin_free_c_string")]
--- a/modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
+++ b/modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
@@ -1,7 +1,11 @@
 use bitcoin::absolute::Decodable;
 use bitcoin::address::Address;
 use bitcoin::bip152::HeaderAndShortIds;
+use bitcoin::bip32::ChainCode;
+use bitcoin::bip32::ChildNumber;
+use bitcoin::bip32::Fingerprint;
 use bitcoin::bip32::Xpriv;
+use bitcoin::bip32::Xpub;
 use bitcoin::block::BlockUncheckedExt;
 use bitcoin::consensus::{deserialize_partial, encode, serialize};
 use bitcoin::script::ScriptExt;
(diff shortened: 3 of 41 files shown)


erickcestari:
> Add the base58 utils to "modules/custommutator/" (similarly to how bech32 and sha256 are there) <- optimal?

Yes, it's the best way to go.

> Does that mean that if we e.g. want to fuzz "long-term", multiple instances of bitcoinfuzz need to be running to cover all targets? (each with a specified mutator?)

Yes, it's very common to have different binaries for fuzzing, since when fuzzing we normally compile it with the intent of running a specific target. Per example, if I am fuzzing the target `deserialize_invoice` doesn't make too much sense to compile the final binary with the rust-bitcoin and bitcoin core.

> Is this something to avoid (shall I reavaluate the multi-mutator solution and instead e.g. implement multiple targets that begin with mutating the input?

In this case that we can reuse the custom mutator for more than one target (xpub and xpriv). We could have the same approach as the:

https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/main.cpp#L245

Where we basically change the custom mutator behavior based in a runtime flag. 
`const char* mt_env = std::getenv("P2P_LIGHTNING_MESSAGE_TYPE");` .

I think the custom mutator for those extended key targets should have these steps:

1. Read environment flag to decide xpub vs xpriv
2. Decode base58 (remove checksum) -> raw 78 bytes
3. Mutate the 78-byte structure with LibFuzzer
4. Force the correct version bytes (first 4 bytes) based on key type (after mutation ensure that it will have the correct prefix)
5. Encode back to base58
6. Recalculate and add checksum (double SHA-256, first 4 bytes)

kuliq23:
Excellent, I'll get to work!

kuliq23:
> 2. Decode base58 (remove checksum) -> raw 78 bytes

Does this mean that the initial `fuzz_data` that the mutator receives is a (valid) key encoded in base58? I thought the mutator receives random bytes and works with them... Is this to do with the "corpus"? (Does the corpus provide an initial set of correct values?)

> 5. Encode back to base58
> 6. Recalculate and add checksum (double SHA-256, first 4 bytes)

nit: I believe these steps should be switched to comply with the scheme.

erickcestari:
> Does this mean that the initial `fuzz_data` that the mutator receives is a (valid) key encoded in base58? I thought the mutator receives random bytes and works with them... Is this to do with the "corpus"? (Does the corpus provide an initial set of correct values?)

The `fuzz_data` depends on the fuzzer's state: initially when there's no corpus, the fuzzer generates random bytes, but once a corpus exists, it uses inputs from the corpus which are valid BIP32-formatted keys (base58-encoded with checksums, since our custom mutator ensures this format). In both cases, the mutator attempts to decode the input to raw bytes, and if decoding fails (e.g., for random initial bytes), we either return early or fall back to a minimal valid input (see the [code here](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/aa89a7fbe41d21b1771b7c099f0c6d125b9d6af6/main.cpp#L373)). Once we have raw bytes, the mutator modifies them and then re-serializes the result into proper BIP32 format with a valid checksum, ensuring that even mutations of random data eventually produce structurally valid keys for the target function to process.

> 
> > 5. Encode back to base58
> > 6. Recalculate and add checksum (double SHA-256, first 4 bytes)
> 
> nit: I believe these steps should be switched to comply with the scheme.

Yes, true

quapka:
nit: Would it be better to follow the _...extended_key..._ naming instead of `xprv` and `xpub`? I guess also `tprv` and `tpub` could be tested. So, e.g.:
- `BIP32DeserializeKeyXpubXprvTarget` -> `BIP32DeserializeExtendedKeyTarget`
- `bip32_deserialize_key_xpub_xprv` ->  `bip32_deserialize_extended_key`

quapka:
@kuliq23, please, keep an eye on https://github.com/bitcoinfuzz/bitcoinfuzz/pull/336. If that gets merged before this one, you'll likely need to move your mutator-related files around as well.

kuliq23:
## Ready for review

- Updated so that deserialization target does not depend on the serialization functionality of each module and returns a string of this format:
```
$"depth={depthHex};fp={fingerprintHex};child={childHex};chaincode={chainHex};key={keyHex}"
----
depth=00;fp=00000000;child=00000000;chaincode=0000000000000000000000000000000000000000000000000000000000000000;key=0000000000000000000000000000000000000000000000000000000000000000
```
- ~~Removed BTCD, as [NewExtendedKey()](https://github.com/btcsuite/btcd/blob/9ff0780da683ec1fa88ce9b9f2f4fd39d5592221/btcutil/hdkeychain/extendedkey.go#L124) performes no checks and should not be used.~~
EDIT: The function responsible for validation is actually [`NewKeyFromString`](https://github.com/btcsuite/btcd/blob/c794769dc23fef28ed99b1fa3a1be92582559452/btcutil/hdkeychain/extendedkey.go#L683), BTCD was added in #557 
- Moved CUSTOMMUTATOR_EXTENDED_KEY to custommutator/mutators/extended_key.h

- FIND:
NBitcoin deserializes extkeys that RustBitcoin and BitcoinCore refuse, namely e.g.:
```
tpubD6NzVbkrk32yV5x4sxuLgohVBbYnknoUqt4KZPF6s2LJpAJA1e3gTbYp4pzHs1QWWsuUdWiAYzmvMPE9YohUybXvF3ur46sW2quurdpqgS5

xpub662eSZPVhg7X1bDRQ1W3nSCz9e3qsT2RDMKMAh5izNfwMPqquLs2mLMjw4tw5ZuUMUNfdwzF4S9fpP857tQnBCxXTRoTcFnyNTunw7kdqTS

tprv8ZgxMBicTyKp8ejpC3WEY9jnMR8ddW8igEyibLiqdvYDMrC7Sp2D7enWsSFhV3s9X1nxtPsXR3V2fXfe9NfmRRrMs7WMER5Trx3PKs4Ekas

xprv9s21bJh6f2qNHs99bFx5N2guoY8kKZaeUfpddvVBG1fwdRQNTRzEr4ABu9TKHGgZ8YXD7CCQPp8YSBMjDRBstw2dLtuN9sPRdZMh7ReMcCv
```
I will investigate this further.

- Bitcoin Core target could be refactored into a more polished version, I'm having trouble with includes in module.h, this does not block the merge

brunoerg:
I'm going to review it soon. Have you investigate your find? Also, be cautious on publicly posting any findings due to security concerns. 

The conflicts that must be resolved is because we added a code formatter check recently, sorry for the noise.

custommutator/mutators/extended_key.h
```diff
@@ -0,0 +1,101 @@
...
+    size_t returned_size = LLVMFuzzerMutate(
+        scratch.data(),
+        BIP32_PAYLOAD_LEN,
+        max_size
+    );
+    if (returned_size > BIP32_PAYLOAD_LEN) returned_size = BIP32_PAYLOAD_LEN;
+
+    // pad remaining bytes if needed
+    if (returned_size < BIP32_PAYLOAD_LEN) {
+        std::memset(scratch.data() + returned_size, 0, BIP32_PAYLOAD_LEN - returned_size);
```
erickcestari: After adding the pad to the scratch, the returned size should be set to BIP32_PAYLOAD_LEN.


```suggestion
    if (returned_size < BIP32_PAYLOAD_LEN) {
        std::memset(scratch.data() + returned_size, 0, BIP32_PAYLOAD_LEN - returned_size);
        returned_size = BIP32_PAYLOAD_LEN;
```

custommutator/mutators/extended_key.h
```diff
@@ -0,0 +1,101 @@
+#ifndef BITCOINFUZZ_CUSTOMMUTATOR_EXTENDED_KEY_H
```
erickcestari, line 1: Nice custom mutator! I've made a few changes, including setting the default to a random prefix instead of the xpub fallback.

```suggestion
#ifndef BITCOINFUZZ_CUSTOMMUTATOR_EXTENDED_KEY_H
#define BITCOINFUZZ_CUSTOMMUTATOR_EXTENDED_KEY_H

#include <custommutator/utils/base58.h>
#include <cstdint>
#include <cstdlib>
#include <cstring>
#include <span>
#include <string>
#include <string_view>
#include <vector>
#include <array>

extern "C" size_t LLVMFuzzerMutate(uint8_t *Data, size_t Size, size_t MaxSize);

constexpr size_t VERSION_PREFIX_LEN = 4;
constexpr size_t BIP32_PAYLOAD_LEN = 78;

struct VersionBytes {
    uint8_t data[VERSION_PREFIX_LEN];
};

constexpr std::array<VersionBytes, 4> VERSION_PREFIXES = {{
    {{0x04, 0x88, 0xB2, 0x1E}},  // xpub
    {{0x04, 0x88, 0xAD, 0xE4}},  // xprv
    {{0x04, 0x35, 0x87, 0xCF}},  // tpub
    {{0x04, 0x35, 0x83, 0x94}},  // tprv
}};

constexpr size_t IDX_XPUB = 0;
constexpr size_t IDX_XPRV = 1;
constexpr size_t IDX_TPUB = 2;
constexpr size_t IDX_TPRV = 3;

const VersionBytes& GetVersionBytes(std::string_view key_type) {
    if (key_type == "xprv") return VERSION_PREFIXES[IDX_XPRV];
    if (key_type == "tpub") return VERSION_PREFIXES[IDX_TPUB];
    if (key_type == "tprv") return VERSION_PREFIXES[IDX_TPRV];
    return VERSION_PREFIXES[IDX_XPUB];
}

const VersionBytes& GetRandomVersion(unsigned int seed) {
    return VERSION_PREFIXES[seed % VERSION_PREFIXES.size()];
}

extern "C" size_t LLVMFuzzerCustomMutator(uint8_t *fuzz_data, size_t size, 
                                          size_t max_size, unsigned int seed) {
    
    // Decode input from base58check
    std::string_view input{reinterpret_cast<const char*>(fuzz_data), size};
    std::vector<unsigned char> payload(BIP32_PAYLOAD_LEN, 0x00);
    
    std::vector<unsigned char> decoded;
    if (DecodeBase58Check(std::string{input}, decoded, static_cast<int>(max_size))) {
        const size_t copy_len = std::min(decoded.size(), BIP32_PAYLOAD_LEN);
        std::memcpy(payload.data(), decoded.data(), copy_len);
    } else {
        const size_t copy_len = std::min(size, BIP32_PAYLOAD_LEN);
        std::memcpy(payload.data(), fuzz_data, copy_len);
    }
    
    // Mutate the payload
    size_t mutated_size = LLVMFuzzerMutate(payload.data(), BIP32_PAYLOAD_LEN, BIP32_PAYLOAD_LEN);
    
    // Ensure we have exactly BIP32_PAYLOAD_LEN bytes
    if (mutated_size < BIP32_PAYLOAD_LEN) {
        std::memset(payload.data() + mutated_size, 0, BIP32_PAYLOAD_LEN - mutated_size);
    }
    payload.resize(BIP32_PAYLOAD_LEN);
    
    // Apply version prefix: use env if set, otherwise random
    const char* env = std::getenv("EXTKEYTYPE");
    const auto& version = env ? GetVersionBytes(env) : GetRandomVersion(seed);
    std::memcpy(payload.data(), version.data, VERSION_PREFIX_LEN);
    
    // Encode back to base58check
    std::string encoded = EncodeBase58Check(std::span<const unsigned char>{payload});
    
    // Copy result to output buffer
    const size_t output_len = std::min(encoded.size(), max_size);
    std::memcpy(fuzz_data, encoded.data(), output_len);

    return output_len;
}

#endif //BITCOINFUZZ_CUSTOMMUTATOR_EXTENDED_KEY_H
```

modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
```diff
@@ -268,3 +273,74 @@ pub unsafe extern "C" fn rust_bitcoin_bip32_master_keygen(
...
+                &xpub.public_key.serialize(), // as bytes
+            );
+            //return formatted string
+            str_to_c_string(&result)
+        } else {
+            str_to_c_string("INVALID")
+        }
+    }
+
+}
```
erickcestari, line 336: We shouldn't retrieve the environment here. Instead, we should try decoding the input using the bip32 key types. We should also do this for the other modules.

```suggestion
#[no_mangle]
pub unsafe extern "C" fn rust_bitcoin_bip32_deserialize_extended_key(
    data: *const u8,
    len: usize,
) -> *mut c_char {
    let data_slice = slice::from_raw_parts(data, len);
    let ext_str = match std::str::from_utf8(data_slice) {
        Ok(s) => s,
        Err(_) => return str_to_c_string("INVALID"),
    };

    if let Ok(xprv) = Xpriv::from_str(&ext_str) {
        //format the result to string
        let result = format_ext_key_common(
            xprv.depth,
            xprv.parent_fingerprint,
            xprv.child_number,
            xprv.chain_code,
            &xprv.private_key.secret_bytes(), // as bytes
        );
        //return formatted string
        str_to_c_string(&result)
    } else {
        if let Ok(xpub) = Xpub::from_str(&ext_str) {
            //format the result to string
            let result = format_ext_key_common(
                xpub.depth,
                xpub.parent_fingerprint,
                xpub.child_number,
                xpub.chain_code,
                &xpub.public_key.serialize(), // as bytes
            );
            //return formatted string
            str_to_c_string(&result)
        } else {
            str_to_c_string("INVALID")
        }
    }
}
```

brunoerg:
Needs rebase

brunoerg:
Any update on it? It's been a month without any change here.

kuliq23:
> Any update on it? It's been a month without any change here.

No update here yet. Im gonna get to it soon, hopefully next week, lots on my plate rn.

custommutator/mutators/extended_key.h
kuliq23: thx

custommutator/mutators/extended_key.h
kuliq23: good catch!

kuliq23:
**Incorporated suggests and rebased.**

- Edit/fix deserialize modules and mutator

- modules no longer use ENV to get extkeytype pre-hand

First symbol of extkey used for testnet vs mainnet recognition in btccore as changing the Network Parameters mid function call causes issues.

---

As mentioned in [issue](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/307#issuecomment-3598251317) , NBitcoin accepts keys of all 4 types that throw "non-zero child number in master key" error in rust_bitcoin and result in empty key-bytes in bitcoin-core, this makes me think that NBitcoin does not simply check the correctness as thoroughly as rust-bitcoin and bitcoin-core and this isn't a groundbreaking issue.

---

Glad to see a formatter! I thought it is incorporated in the CI pipeline but the fromat check seems to fail, any hints on how I avoid that?

erickcestari:
> Glad to see a formatter! I thought it is incorporated in the CI pipeline but the fromat check seems to fail, any hints on how I avoid that?

You can run `make format-all` from the project root directory. Alternatively, run `make format` inside each module.

You can also check that the formatting is correct using the commands `make check-format-all` or `make check-format`, respectively.

The CI is displaying that the NBitcoin, basemodule, bitcoin core and rust_bitcoin modules are not correctly formatted.

erickcestari:
> As mentioned in https://github.com/bitcoinfuzz/bitcoinfuzz/pull/307#issuecomment-3598251317 , NBitcoin accepts keys of all 4 types that throw "non-zero child number in master key" error in rust_bitcoin and result in empty key-bytes in bitcoin-core, this makes me think that NBitcoin does not simply check the correctness as thoroughly as rust-bitcoin and bitcoin-core and this isn't a groundbreaking issue.

It's not a groundbreaking issue, but it's the type of issue that differential fuzzing is great at catching. I created a PR in NBitcoin to fix it. https://github.com/MetacoSA/NBitcoin/pull/1294.

Nice work!

modules/nbitcoin/NBitcoin/NBitcoinBridge.cs
```diff
@@ -172,6 +173,58 @@ public static IntPtr BIP32MasterKeygen(IntPtr dataPtr, UIntPtr len)
         return strPtr;
     }
 
+    [UnmanagedCallersOnly(EntryPoint = "nbitcoin_bip32_deserialize_extended_key")]
+    public static IntPtr BIP32DeserializeExtendedKeyTarget(IntPtr inputPtr)
+    {
+        if (inputPtr == IntPtr.Zero) return Marshal.StringToCoTaskMemUTF8("INVALID");
+
+        string input = Marshal.PtrToStringUTF8(inputPtr);
```
erickcestari: nit:

```suggestion
        string input = Marshal.PtrToStringUTF8(inputPtr) ?? "";
```

custommutator/mutators/extended_key.h
```diff
@@ -0,0 +1,87 @@
...
+      EncodeBase58Check(std::span<const unsigned char>{payload});
+
+  // Copy result to output buffer
+  const size_t output_len = std::min(encoded.size(), max_size);
+  std::memcpy(fuzz_data, encoded.data(), output_len);
+  // use hardcoded
+  // xprv9s21ZrQLyYpt8mej728d1sv48AxGUreFVUg2Ag6rdB8p6sHE7X7o4XxuokQTr9ucc2Wc6q5obDdTJBQFkpDTLJm7tqd1naA19inytoE1nEN
+  // std::memcpy(fuzz_data,
+  // "xprv9s21ZrQLyYpt8mej728d1sv48AxGUreFVUg2Ag6rdB8p6sHE7X7o4XxuokQTr9ucc2Wc6q5obDdTJBQFkpDTLJm7tqd1naA19inytoE1nEN",
+  // output_len);
```
erickcestari: nit: We can remove this hardcoded extended key comment. 
```suggestion

```

erickcestari (changes requested):
Just a few nits. Then it should be ready to merge.

custommutator/mutators/extended_key.h
kuliq23: yes, i overlooked that, ty

modules/nbitcoin/NBitcoin/NBitcoinBridge.cs
kuliq23: good idea, thanks

kuliq23:
Thanks for the review and the mention in PR. Also thanks for guiding me. Hopefully ready to merge now. :)

erickcestari (approved):
Thanks! Nice work!

ACK bd8572e05480d9306f3ad59f02ccde838c6cc061

PR #308, opened 2025-10-27
fix: handle offers without explicit chain in c-lightning and LDK modules

erickcestari:
We were returning the default chain when decoding an offer without
containing a chain in core lightning, which caused crashes between ldk
and core lightning, so now we return in both to mantain compatibility
with Eclair that does that by default.

### DIFF
--- a/modules/clightning/module.cpp
+++ b/modules/clightning/module.cpp
@@ -186,7 +186,8 @@ std::string clightning_des_offer(const std::string_view input) {
             result << hex_encode(offer->offer_chains[i].shad.sha.u.u8, 32);
         }
     } else {
-        // If no chains are specified, Clightning defaults to bitcoin
+        // If no chains are specified, we fallback to the bitcoin chain (this is
+        // is necessary for compatibility with Eclair).
         struct bitcoin_blkid chain = chainparams_for_network("bitcoin")->genesis_blockhash;
         result << hex_encode(chain.shad.sha.u.u8, 32);
     }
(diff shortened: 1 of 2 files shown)


brunoerg (approved):
ACK 89b86fa783262d395ca62185a76b28e6bc4b88c3

erickcestari:
@brunoerg, can you review it again? I changed the logic, now ldk also returns the default chain to be compatible with eclair.

modules/clightning/module.cpp
```diff
@@ -186,7 +186,8 @@ std::string clightning_des_offer(const std::string_view input) {
             result << hex_encode(offer->offer_chains[i].shad.sha.u.u8, 32);
         }
     } else {
-        // If no chains are specified, Clightning defaults to bitcoin
+        // If no chains are specified, we fallback to the bitcoin chain (this is
+        // is necessary for compatibility with Eclair).
```
brunoerg, line 190: Any issue that tracks/explain this compatibility issue with Eclair?

brunoerg (approved):
reACK 32ef3f132b99564c3584ea74ab52719d4d878219

modules/clightning/module.cpp
erickcestari: Idk if we need to create an issue for this. It's more an API choice than an actual bug or compatibility problem, so a comment should work. The BOLT12 spec says when there's no chains field in the offer, it just defaults to Bitcoin mainnet.

Issue #291, opened 2025-10-13
parse_p2p_lightning_message(open_channel): LND trims leading zeros in channel_type


erickcestari:
```
Lightning P2P message parsing failed
Module: Lnd
[...]

PR #312, opened 2025-10-30
feat(open_channel): strip leading zeros to maintain compatibility with LND

erickcestari:
Avoid crashing all time because of leading zeros of channel_type.

More info: #291


brunoerg (approved):
code review ACK 67f6396587a6651b7c2a2a7f7b2a207557629896

PR #318, opened 2025-11-08
BIP32: Add BTCD module to masterkeygen target

kuliq23:
BTCD [master keygen func](https://github.com/btcsuite/btcd/blob/9ff0780da683ec1fa88ce9b9f2f4fd39d5592221/btcutil/hdkeychain/extendedkey.go#L652) differs from other modules by accepting only 16-64 bytes long sequence as input. Bitcoin Core, NBitcoin and Rustbitcoin don't check this. This is why the fuzzing fails now.

[BIP32 docs](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki) MK Generation step in question:

* Generate a seed byte sequence S of a chosen length (between 128 and 512 bits; 256 bits is advised) from a (P)RNG. 

**I suppose correct solution would be to feed only byte sequences of this specified length to this target?**

I will try to investigate possible problems and implications of not checking the length in mentioned modules as well as possible mishaps in handling bits (in spec) vs bytes.

### DIFF
--- a/modules/btcd/btcd_wrapper/wrapper.go
+++ b/modules/btcd/btcd_wrapper/wrapper.go
@@ -22,6 +22,7 @@ import (
 	"github.com/btcsuite/btcd/addrmgr"
 	"github.com/btcsuite/btcd/blockchain"
 	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/btcutil/hdkeychain"
 	"github.com/btcsuite/btcd/btcutil/psbt"
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/txscript"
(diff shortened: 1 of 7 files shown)


modules/btcd/btcd_wrapper/wrapper.go
```diff
@@ -281,4 +282,14 @@ func BTCDAddress(data C.ByteArray) *C.char {
...
+//export BTCDBip32MasterKeygen
+func BTCDBip32MasterKeygen(data C.ByteArray) *C.char {
+	seed := C.GoBytes(unsafe.Pointer(data.data), data.length)
+	masterKey, err := hdkeychain.NewMaster(seed, &chaincfg.MainNetParams)
+	if err != nil {
+		return C.CString(fmt.Sprintf("ERROR: %v", err))
+	}
+	return C.CString(masterKey.String())
+}
+
```
erickcestari: We should skip this case because other implementations allow it. Even though btcd validates seed length (128-512 bits), we want to test that inputs accepted by other implementations execute correctly.

```suggestion
//export BTCDBip32MasterKeygen
func BTCDBip32MasterKeygen(data C.ByteArray) *C.char {
	seed := C.GoBytes(unsafe.Pointer(data.data), data.length)
	masterKey, err := hdkeychain.NewMaster(seed, &chaincfg.MainNetParams)
	if err != nil {
		// Skip seed length validation errors (128-512 bits requirement) as other
		// implementations don't enforce this constraint.
		if err.Error() == "seed length must be between 128 and 512 bits" {
			return nil
		}
		return C.CString("")
	}
	return C.CString(masterKey.String())
}
```

erickcestari (changes requested):

kuliq23:
Thanks! Will rebase and remove draft in the morning.

erickcestari:
ACK 17dde5d70586622a5e8057c9d30c0138fd48c2a9

Nice work!!

erickcestari (approved):

PR #322, opened 2025-11-12
feat: add new lightning message init

erickcestari:


modules/clightning/module.cpp
```diff
@@ -279,6 +280,54 @@ std::optional<std::string> clightning_parse_p2p_lightning_message(std::span<cons
...
+
+        u8 *features_combined = featurebits_or(tmpctx, globalfeatures, features);
+
+        result << "MSG_TYPE=init";
+        result << ";FEATURES=" << tal_hex(tmpctx, features_combined);
+
+        if (tlvs->networks) {
+            result << ";NETWORKS=";
+            for (size_t i = 0; i < tal_count(tlvs->networks); ++i) {
+                result << hex_encode(tlvs->networks[i].shad.sha.u.u8, 32) << ";";
```
morehouse: Since all the networks apply to the `NETWORKS=`, it would be more readable to separate them with `,` instead of `;`.
```suggestion
                result << hex_encode(tlvs->networks[i].shad.sha.u.u8, 32) << ",";
```

modules/clightning/module.cpp
```diff
@@ -279,6 +280,54 @@ std::optional<std::string> clightning_parse_p2p_lightning_message(std::span<cons
...
+            }
+        }
+
+        if (tlvs->remote_addr) {
```
morehouse: Let's combine the two `if (tlvs->remote_addr)` blocks.

modules/clightning/module.cpp
```diff
@@ -279,6 +280,54 @@ std::optional<std::string> clightning_parse_p2p_lightning_message(std::span<cons
...
+            // C-lightning doens't verify that the remote address is a valid DNS name with ASCII characters in decoding phase
+            if (remote_addr->type == ADDR_TYPE_DNS) {
+                return std::nullopt;
+            }
+            if (remote_addr->type == ADDR_TYPE_IPV6) {
+                // Use hex representation to avoid differences between c-lightning and rust-lightning formats for IPV6
+                result << ";REMOTE_NETWORK_ADDRESS=" << hex_encode(remote_addr->addr, remote_addr->addrlen) << ":" << remote_addr->port;
+            } else {
+                result << ";REMOTE_NETWORK_ADDRESS=" << fmt_wireaddr(tmpctx, remote_addr);
+            }
```
morehouse, line 328:
```suggestion
            result << ";REMOTE_NETWORK_ADDRESS=";
            if (remote_addr->type == ADDR_TYPE_IPV6) {
                // Use hex representation to avoid differences between c-lightning and rust-lightning formats for IPV6
                result << hex_encode(remote_addr->addr, remote_addr->addrlen) << ":" << remote_addr->port;
            } else {
                result << fmt_wireaddr(tmpctx, remote_addr);
            }
```

modules/ldk/ldk_lib/src/lib.rs
```diff
@@ -286,6 +286,46 @@ pub unsafe extern "C" fn ldk_parse_p2p_lightning_message(
...
+                        SocketAddress::TcpIpV6 { addr, port } => {
+                            // Use hex representation to avoid differences between c-lightning and rust-lightning formats for IPV6
+                            result.push_str(&format!(
+                                "{}:{}",
+                                addr.to_hex_string(Case::Lower),
+                                port
+                            ));
+                        }
+                        SocketAddress::OnionV2(_) => {
+                            // Skip OnionV2 since is deprecated
```
morehouse:
```suggestion
                            // Skip OnionV2 since it's deprecated and LDK doesn't fully parse it
```

erickcestari: Done!

modules/clightning/module.cpp
erickcestari: Done!

modules/clightning/module.cpp
erickcestari: Updated!

modules/clightning/module.cpp
erickcestari: Updated!

morehouse (approved):
LGTM

PR #328, opened 2025-11-15
fix: strip features and channel type leading zeros

erickcestari:
Should fix this CI error: https://github.com/bitcoinfuzz/bitcoinfuzz/actions/runs/19376093989/job/55444927272


brunoerg (approved):
ACK 20ca3043165e6093b686dff085e804b3dd077e7d

PR #332, opened 2025-11-19
nbitcoin: add psbt_parse target support

brunoerg:


modules/nbitcoin/NBitcoin/NBitcoinBridge.cs
```diff
@@ -1,11 +1,96 @@
...
+            catch
+            {
+                return Marshal.StringToHGlobalAnsi("");
+            }
+        }
+        catch
+        {
+            return Marshal.StringToHGlobalAnsi("");
+        }
+    }
```
erickcestari, line 75: nit: Since all try-catch blocks return the same value, we could only have one block.

driver.cpp
```diff
@@ -228,6 +228,11 @@ namespace bitcoinfuzz
             std::optional<std::string> res{module.second->psbt_parse(buffer)};
             if (!res.has_value()) continue;
 
+            #ifdef NBITCOIN
+            if (res->find("in=0") != std::string::npos ||
+                res->find("out=0") != std::string::npos) return;
+            #endif
```
erickcestari, line 234: Why do we need this? Shouldn't we skip it by returning a null pointer in the NBitcoin module instead of skipping it here in driver.cpp?

erickcestari (commented):
The NBitcoin needs to be added to the docker-compose.yml for psbt_parse.

modules/nbitcoin/NBitcoin/NBitcoinBridge.cs
brunoerg: Done.

brunoerg:
> The NBitcoin needs to be added to the docker-compose.yml for psbt_parse.

Just added it.

driver.cpp
brunoerg: Sometimes another implementation returns in=0;out=0; and NBitcoin returns an empty string, sometimes NBitcoin returns in=0;out=0; and another implementation an empty string. So it's just a false signal, not a bug and not helpful for us, better to skip this case when running with NBitcoin.

erickcestari: Got it!

erickcestari (approved):
ACK 9bb24b7cde88f980ac3a2e6a2b2cd0221192b212

Issue #331, opened 2025-11-19
Make differential fuzzing between secp256k1 libraries


erickcestari:
I think the main libraries we should have here are those written in different programming languages. For example, we will not add Rust-secp256k1, an FFI of secp256k1 from Bitcoin Core.

Suggestions:

- [x] [secp256k1](https://github.com/bitcoin-core/secp256k1) (C)
- [x] [secp256k1-go](https://github.com/decred/dcrd/tree/master/dcrec/secp256k1) (go) - LND uses this one
- [ ] [noble-secp256k1](https://github.com/paulmillr/noble-secp256k1) (js)
- [x] [NBitcoin-secp256k1](https://github.com/MetacoSA/NBitcoin/tree/master/NBitcoin.Secp256k1) (c#)
- [ ] [secp256kfun](https://github.com/LLFourn/secp256kfun) (rust)
- [x] [k256](https://crates.io/crates/k256) (rust)
- [ ] [fia-crypto](https://github.com/mit-plv/fiat-crypto)

We can get some ideas from cryptofuzz, since they have some targets for secp256k1: https://github.com/MozillaSecurity/cryptofuzz

brunoerg:
Since cryptofuzz is not alive anymore and I don't see constant maintaince on this Mozilla fork, I think it makes sense to have it here.

NishantBansal2003:
Was thinking of doing differential fuzzing between libsecp256k1 and Decred’s secp256k1. If this hasn’t been started yet, can I pick it up?

erickcestari:
> Was thinking of doing differential fuzzing between libsecp256k1 and Decred’s secp256k1. If this hasn’t been started yet, can I pick it up?

Yes, feel free to take this on.

luisschwab:
There's also [`secp256kfun`](https://github.com/LLFourn/secp256kfun) in Rust (you'll probably find many bugs there).

luisschwab:
And also [`k256`](https://crates.io/crates/k256). This one is a production-grade secp256k1 library.

erickcestari:
Good suggestions! Those are definitely worth having.

erickcestari:
@NishantBansal2003 Are you planning on working on adding the other modules?

NishantBansal2003:
> @NishantBansal2003 Are you planning on working on adding the other modules?


[...]

PR #338, opened 2025-11-26
Add differential fuzzing for libsecp256k1 and decred secp256k1

NishantBansal2003:
ref: #331 

Add differential fuzzing between the libsecp256k1 (C) and Decred secp256k1 (Go) libraries. This initial implementation covers commonly used functions such as private to public key operations, signing (compact and DER formats), and ECDH operations.
The approach can be extended further to include additional functions and other libraries. To keep this PR manageable for review, only these functions and the two libraries are included so far.


NishantBansal2003:
Reason for opening in draft is that the functions  `private_to_public`  and  `ecdh`  are working correctly, but the functions  `sign_compact`  and  `sign_der`  have mismatched signatures between modules. I’m not sure if this is an actual bug or my implementation issue. I have checked that both functions follow the lower-S form and deterministic RFC6979. I want to know your thoughts before concluding anything.

erickcestari:
Nice work! It's looking good.

I'm not sure if we should use the same Secp256k1 library that is inside in the Bitcoin module. Having a Git submodule would make it easy to update the secp256k1 library without affecting the bitcoin module. What do you think? @brunoerg

driver.cpp
```diff
@@ -482,6 +600,18 @@ namespace bitcoinfuzz
...
+            this->PrivateToPublicTarget(buffer);
+        } else if (target == "sign_compact")
+        {
+            this->SignCompactTarget(buffer);
+        } else if (target == "sign_der")
+        {
+            this->SignDerTarget(buffer);
+        } else if (target == "ecdh")
+        {
+            this->ECDHTarget(buffer);
```
erickcestari: nit: let's follow the same pattern of else if

```suggestion
        } else if (target == "private_to_public") {
            this->PrivateToPublicTarget(buffer);
        } else if (target == "sign_compact") {
            this->SignCompactTarget(buffer);
        } else if (target == "sign_der") {
            this->SignDerTarget(buffer);
        } else if (target == "ecdh") {
            this->ECDHTarget(buffer);
```

modules/secp256k1/module.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+secp256k1_context *secp256k1_ctx;
+
+struct CleanSecp256k1CtxGuard
+{
+	CleanSecp256k1CtxGuard() noexcept = default;
+	~CleanSecp256k1CtxGuard() noexcept { secp256k1_context_destroy(secp256k1_ctx); }
+	CleanSecp256k1CtxGuard(const CleanSecp256k1CtxGuard &) = delete;
+	CleanSecp256k1CtxGuard &operator=(const CleanSecp256k1CtxGuard &) = delete;
+};
```
erickcestari: When I tried to run the target `private_to_public` I got some memory leak errors. Because the fuzzer creates Secp256k1 objects which call `init()` to allocate a 208-byte context, but doesn't always call the functions that contain `CleanSecp256k1CtxGuard`. When objects are destroyed without their functions being called, the allocated contexts are never freed because the cleanup guard was never executed. The fix is to move `secp256k1_context_destroy()` to the Secp256k1 destructor so cleanup happens at object destruction, not function execution.

```
diff --git a/modules/secp256k1/module.cpp b/modules/secp256k1/module.cpp
index 5c4d331..384be45 100644
--- a/modules/secp256k1/module.cpp
+++ b/modules/secp256k1/module.cpp
@@ -20,14 +20,6 @@ extern "C"
 
 secp256k1_context *secp256k1_ctx;
 
-struct CleanSecp256k1CtxGuard
-{
-	CleanSecp256k1CtxGuard() noexcept = default;
-	~CleanSecp256k1CtxGuard() noexcept { secp256k1_context_destroy(secp256k1_ctx); }
-	CleanSecp256k1CtxGuard(const CleanSecp256k1CtxGuard &) = delete;
-	CleanSecp256k1CtxGuard &operator=(const CleanSecp256k1CtxGuard &) = delete;
-};
-
 void init(int *argc, char ***argv)
 {
 	secp256k1_ctx = secp256k1_context_create(SECP256K1_CONTEXT_NONE);
@@ -67,7 +59,6 @@ int create_signature_from_hash(const uint8_t *privkey, const uint8_t *hash, secp
 
 std::optional<std::string> secp256k1_private_to_public(std::span<const uint8_t> buffer)
 {
-	CleanSecp256k1CtxGuard _cleanup;
 	int ret = 0;
 	secp256k1_pubkey pubkey;
 	size_t pubkey_len = SECP256K1_PUBKEY_COMPRESSED_LEN;
@@ -84,7 +75,6 @@ std::optional<std::string> secp256k1_private_to_public(std::span<const uint8_t>
 
 std::optional<std::string> secp256k1_sign_compact(std::span<const uint8_t> buffer, std::span<const uint8_t> hash)
 {
-	CleanSecp256k1CtxGuard _cleanup;
 	int ret = 0;
 	secp256k1_ecdsa_signature signature;
 	std::vector<uint8_t> signature_compact(SECP256K1_SIGNATURE_COMPACT_LEN);
@@ -100,7 +90,6 @@ std::optional<std::string> secp256k1_sign_compact(std::span<const uint8_t> buffe
 
 std::optional<std::string> secp256k1_sign_der(std::span<const uint8_t> buffer, std::span<const uint8_t> hash)
 {
-	CleanSecp256k1CtxGuard _cleanup;
 	int ret = 0;
 	secp256k1_ecdsa_signature signature;
 	size_t signature_der_len = SECP256K1_SIGNATURE_DER_MAX_LEN;
@@ -117,7 +106,6 @@ std::optional<std::string> secp256k1_sign_der(std::span<const uint8_t> buffer, s
 
 std::optional<std::string> secp256k1_ecdh(std::span<const uint8_t> buffer)
 {
-	CleanSecp256k1CtxGuard _cleanup;
 	int ret = 0;
 	secp256k1_pubkey pubkey;
 	std::vector<uint8_t> shared_secret(SECP256K1_SHARED_SECRET_LEN);
@@ -160,5 +148,12 @@ namespace bitcoinfuzz
 			return secp256k1_ecdh(buffer);
 		}
 
+		Secp256k1::~Secp256k1()
+		{
+			if (secp256k1_ctx) {
+				secp256k1_context_destroy(secp256k1_ctx);
+				secp256k1_ctx = nullptr;
+			}
+		}
 	}
 }
diff --git a/modules/secp256k1/module.h b/modules/secp256k1/module.h
index efabdc3..82e5b06 100644
--- a/modules/secp256k1/module.h
+++ b/modules/secp256k1/module.h
@@ -15,7 +15,7 @@ namespace bitcoinfuzz
 			std::optional<std::string> sign_compact(std::span<const uint8_t> buffer, std::span<const uint8_t> hash) const override;
 			std::optional<std::string> sign_der(std::span<const uint8_t> buffer, std::span<const uint8_t> hash) const override;
 			std::optional<std::string> ecdh(std::span<const uint8_t> buffer) const override;
-			~Secp256k1() noexcept override = default;
+			~Secp256k1();
 		};
 	}
 }
\ No newline at end of file
```

erickcestari (changes requested):
I didn't review the `sign_compact` and `sign_der` target yet

modules/secp256k1/module.cpp
NishantBansal2003: I initially went with similar approach but found similar context create and cleanup behavior in the clightning module, so I followed that pattern. So my question is if there are memory leak issues here, shouldn’t the clightning module exhibit memory leak cases as well?
Also how do you run bitcoinfuzz with memory leak checks?

driver.cpp
NishantBansal2003: Was about to ask what C++ formatter you are using? I can see there are a lot of places in my code where the formatting got changed when I format the document with my formatter. In other places in my code, the formatting is a little different as well then usual.

modules/secp256k1/module.cpp
erickcestari: > I initially went with similar approach but found similar context create and cleanup behavior in the clightning module, so I followed that pattern. So my question is if there are memory leak issues here, shouldn’t the clightning module exhibit memory leak cases as well?

The clightning's `tmpctx` pattern keeps the global context alive and only frees its **children** on each call, allowing the context to be reused across multiple operations. The `secp256k1` module code destroys and recreates the entire global context with every function call, causing a leak when no functions are called. Actually I think the best way to go is to create the secp256k1 context once in init(), keep it alive throughout the program lifetime, and only destroy it at program exit.

This diff can solve that:

* Remove the `CleanSecp256k1CtxGuard`
* Update the init to check if the `secp256k1_ctx` was already created
* Don't call `secp256k1_context_destroy`, let the OS reclaim memory on exit

```
diff --git a/modules/secp256k1/module.cpp b/modules/secp256k1/module.cpp
index 5c4d331..0d60085 100644
--- a/modules/secp256k1/module.cpp
+++ b/modules/secp256k1/module.cpp
@@ -20,18 +20,12 @@ extern "C"
 
 secp256k1_context *secp256k1_ctx;
 
-struct CleanSecp256k1CtxGuard
-{
-	CleanSecp256k1CtxGuard() noexcept = default;
-	~CleanSecp256k1CtxGuard() noexcept { secp256k1_context_destroy(secp256k1_ctx); }
-	CleanSecp256k1CtxGuard(const CleanSecp256k1CtxGuard &) = delete;
-	CleanSecp256k1CtxGuard &operator=(const CleanSecp256k1CtxGuard &) = delete;
-};
-
 void init(int *argc, char ***argv)
 {
-	secp256k1_ctx = secp256k1_context_create(SECP256K1_CONTEXT_NONE);
-	assert(secp256k1_ctx);
+    if (!secp256k1_ctx) {
+        secp256k1_ctx = secp256k1_context_create(SECP256K1_CONTEXT_NONE);
+        assert(secp256k1_ctx);
+    }
 }
 
 std::string hex_encode(const unsigned char *data, size_t len)
@@ -67,7 +61,6 @@ int create_signature_from_hash(const uint8_t *privkey, const uint8_t *hash, secp
 
 std::optional<std::string> secp256k1_private_to_public(std::span<const uint8_t> buffer)
 {
-	CleanSecp256k1CtxGuard _cleanup;
 	int ret = 0;
 	secp256k1_pubkey pubkey;
 	size_t pubkey_len = SECP256K1_PUBKEY_COMPRESSED_LEN;
@@ -84,7 +77,6 @@ std::optional<std::string> secp256k1_private_to_public(std::span<const uint8_t>
 
 std::optional<std::string> secp256k1_sign_compact(std::span<const uint8_t> buffer, std::span<const uint8_t> hash)
 {
-	CleanSecp256k1CtxGuard _cleanup;
 	int ret = 0;
 	secp256k1_ecdsa_signature signature;
 	std::vector<uint8_t> signature_compact(SECP256K1_SIGNATURE_COMPACT_LEN);
@@ -100,7 +92,6 @@ std::optional<std::string> secp256k1_sign_compact(std::span<const uint8_t> buffe
 
 std::optional<std::string> secp256k1_sign_der(std::span<const uint8_t> buffer, std::span<const uint8_t> hash)
 {
-	CleanSecp256k1CtxGuard _cleanup;
 	int ret = 0;
 	secp256k1_ecdsa_signature signature;
 	size_t signature_der_len = SECP256K1_SIGNATURE_DER_MAX_LEN;
@@ -117,7 +108,6 @@ std::optional<std::string> secp256k1_sign_der(std::span<const uint8_t> buffer, s
 
 std::optional<std::string> secp256k1_ecdh(std::span<const uint8_t> buffer)
 {
-	CleanSecp256k1CtxGuard _cleanup;
 	int ret = 0;
 	secp256k1_pubkey pubkey;
 	std::vector<uint8_t> shared_secret(SECP256K1_SHARED_SECRET_LEN);
```

> Also how do you run bitcoinfuzz with memory leak checks?

If you are using the makefiles to compile with clang and clang++  AddressSanitizer should already be enabled. AddressSanitizer is a compiler instrumentation that detects memory errors, such as leaks, use-after-free, and undefined behavior.
https://clang.llvm.org/docs/AddressSanitizer.html#introduction

driver.cpp
erickcestari: Yes, we haven't actually been using a formatter. We should establish one to make formatting easier and avoid having to do it manually. Since we have many languages in this project, it would probably be best to establish a formatter for each language. I will open an issue to discuss that further.

erickcestari:
> Reason for opening in draft is that the functions  `private_to_public`  and  `ecdh`  are working correctly, but the functions  `sign_compact`  and  `sign_der`  have mismatched signatures between modules. I’m not sure if this is an actual bug or my implementation issue. I have checked that both functions follow the lower-S form and deterministic RFC6979. I want to know your thoughts before concluding anything.

The issue is that when signing with decred with always normalize the signature to lower-S form. We should call the `secp256k1_ecdsa_signature_normalize` function in the secp256k1 module after signing to have the same behavior.


```cpp
int create_signature_from_hash(const uint8_t *privkey, const uint8_t *hash, secp256k1_ecdsa_signature& signature)
{
	int ret = 0;

	ret = secp256k1_ec_seckey_verify(secp256k1_ctx, privkey);
	ret = ret && secp256k1_ecdsa_sign(secp256k1_ctx, &signature, hash, privkey, nullptr, nullptr);
	ret = ret && secp256k1_ecdsa_signature_normalize(secp256k1_ctx, &signature, &signature); // add this line

	return ret;
}
```

NishantBansal2003:
Thanks for looking into this. Yes, we could do that and match the implementation of decred secp256k1 with libsecp256k1. However, the reason I did not add this in the first place is that I don’t see CLN using the normalize function at all, as seen here: https://github.com/ElementsProject/lightning/blob/534e3ae8f8694ff6ddf4083e9f43f9fabfa34668/bitcoin/signature.c#L97-L118. I also verified this on Bitcoin StackExchange: https://bitcoin.stackexchange.com/questions/129270/why-do-different-implementations-produce-different-der-signatures-for-the-same-p and it turns out there can be different valid signatures that are all accepted by the Bitcoin network.
So, for this target, we can use the normalize function, but a very good target to have would be one that generates a signature using the private key and hash, then provides that signature to other implementations for verification using the public key and hash. I’m not sure(yet) how this could be implemented in bitcoinfuzz, but I think it would be a very useful target.

driver.cpp
```diff
@@ -447,6 +447,124 @@ namespace bitcoinfuzz
...
+
+            last_response = res.value();
+            last_module_name = module.first;
+        }
+    }
+
+    void Driver::SignDerTarget(std::span<const uint8_t> buffer) const
+    {
+        FuzzedDataProvider provider(buffer.data(), buffer.size());
+        if (buffer.size() < 64) { return; }
```
brunoerg: nit
```suggestion
        if (buffer.size() < 64) return;
```

driver.cpp
```diff
@@ -447,6 +447,124 @@ namespace bitcoinfuzz
...
+
+        std::vector<uint8_t> privKeyBuffer = provider.ConsumeBytes<uint8_t>(32);
```
brunoerg: Please, use snake case for new vars.

```suggestion
        std::vector<uint8_t> privkey_buffer{provider.ConsumeBytes<uint8_t>(32)};
```

modules/decredsecp256k1/decredsecp256k1_wrapper/go.mod
```diff
@@ -0,0 +1,5 @@
+module decredsecp256k1_wrapper
+
+go 1.25.4
+
+require github.com/decred/dcrd/dcrec/secp256k1/v4 v4.0.1
```
brunoerg: We're now fuzzing implementations based on master branch instead of a release version. Can you change this to be commit based?

brunoerg:
> I'm not sure if we should use the same Secp256k1 library that is inside in the Bitcoin module. Having a Git submodule would make it easy to update the secp256k1 library without affecting the bitcoin module. What do you think? @brunoerg

I prefer to have a git submodule, better to manage at this point and we don't become dependant on Bitcoin Core. But maybe we could leave it as-is and update it to use submodule in a follow-up?

NishantBansal2003:
> I prefer to have a git submodule, better to manage at this point and we don't become dependant on Bitcoin Core. But maybe we could leave it as-is and update it to use submodule in a follow-up?

There are only a couple of changes needed for this, so let’s include them in this PR itself

erickcestari:
> Thanks for looking into this. Yes, we could do that and match the implementation of decred secp256k1 with libsecp256k1. However, the reason I did not add this in the first place is that I don’t see CLN using the normalize function at all, as seen here: https://github.com/ElementsProject/lightning/blob/534e3ae8f8694ff6ddf4083e9f43f9fabfa34668/bitcoin/signature.c#L97-L118. I also verified this on Bitcoin StackExchange: https://bitcoin.stackexchange.com/questions/129270/why-do-different-implementations-produce-different-der-signatures-for-the-same-p and it turns out there can be different valid signatures that are all accepted by the Bitcoin network. So, for this target, we can use the normalize function, but a very good target to have would be one that generates a signature using the private key and hash, then provides that signature to other implementations for verification using the public key and hash. I’m not sure(yet) how this could be implemented in bitcoinfuzz, but I think it would be a very useful target.

If you examine the Core Lightning code closely, you will see that it creates a loop until a signature with a low S is generated. (Bitcoin Core also does this: https://github.com/bitcoin/bitcoin/blob/bf69442b3f5004dc3df5a1b1d752114ba68fa5f4/src/key.cpp#L221)
```c
/* Grind for low R */
	do {
		ok = secp256k1_ecdsa_sign(secp256k1_ctx,
					  s,
					  h->sha.u.u8,
					  privkey->secret.data, NULL,
					  dev_no_signature_grind ? NULL
					  : extra_entropy);
		((u32 *)extra_entropy)[0]++;
		if (dev_no_signature_grind)
			break;
	} while (!sig_has_low_r(s));
```

Ultimately, all implementations using libsecp256k1 (https://github.com/bitcoin-core/secp256k1) must first normalize and then verify signatures.

At the decred secp256k1 implementation they mention that it's in accordance with the RFC6979 and BIP0062 (https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki#low-s-values-in-signatures): 
```
// The produced signature is deterministic (same message and same
// key yield the same signature) and canonical in accordance with RFC6979 and
// BIP0062.
```

This means that the sign function will always be normalized, i.e., have a low-S signature. That's why normalization is necessary. I think it's fine, since they should all give the same result. Probably all the other secp256k1 libraries will have this functionality to test.

https://github.com/bitcoin/bitcoin/blob/4b24bfeab9d6732aae3e69efd33105792ef1198f/src/pubkey.cpp#L283

> a very good target to have would be one that generates a signature using the private key and hash, then provides that signature to other implementations for verification using the public key and hash. I’m not sure(yet) how this could be implemented in bitcoinfuzz, but I think it would be a very useful target.

We could create a custom mutator that would allow us to do this.

erickcestari (approved):
Thanks!!

ACK 72e1c1086d55cb53d1baf3552f3b02144cb4d9e8

.github/workflows/workflow.yml
```diff
@@ -170,6 +170,16 @@ jobs:
...
+      
+      - name: Build - decredsecp256k1
+        timeout-minutes: 5
+        run: |
+          cd modules/decredsecp256k1 && make
+      
+      - name: Build - secp256k1
+        timeout-minutes: 5
+        run: |
+          cd modules/secp256k1 && make
```
brunoerg, line 194: Now that it uses a submodule for the secp, you should add here that initializing the submodule is needed (similar to what we say for clightning:  `git submodule update --init --recursive external/lightning`).

driver.cpp
```diff
@@ -482,6 +600,14 @@ namespace bitcoinfuzz
             this->TransactionEvalTarget(buffer);
         } else if (target == "bip32_master_keygen") {
             this->Bip32MasterKeygenTarget(buffer);
+        } else if (target == "private_to_public") {
```
brunoerg: I don't like this name `private_to_public`. I think that `private_to_public_key` would be better.

modules/secp256k1/module.cpp
```diff
@@ -0,0 +1,160 @@
...
+{
+	int ret = 0;
+	secp256k1_ecdsa_signature signature;
+	std::vector<uint8_t> signature_compact(SECP256K1_SIGNATURE_COMPACT_LEN);
+
+	ret = create_signature_from_hash(buffer.data(), hash.data(), signature);
+	ret = ret && secp256k1_ecdsa_signature_serialize_compact(secp256k1_ctx, signature_compact.data(), &signature);
+	if (!ret)
+	{
+		return std::nullopt;
```
brunoerg: Why it's returning `std::nullopt` here? In case it was not able to create the signature, for example, we shouldn't ignore it, right?

brunoerg: I don't see a similar behaviour for the Decred module.

modules/secp256k1/module.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+	ret = create_pubkey_from_privkey(buffer.data(), pubkey);
+	ret = ret && secp256k1_ec_pubkey_serialize(secp256k1_ctx, pubkey_compressed.data(), &pubkey_len, &pubkey, SECP256K1_EC_COMPRESSED);
+	if (!ret)
+	{
+		return std::nullopt;
+	}
+
+	return hex_encode(pubkey_compressed.data(), pubkey_len);
+}
```
erickcestari: This aligns with what Bruno commented. We should only skip when parsing the private key since decred secp256k1 is much more lenient. In this context, if secp256k1 can parse the private key, then it should be able to create a public key from it. If not, we will be able to catch this bug by returning an empty string. We should only return a null pointer in very specific cases where we want to skip the string comparison in driver.cpp.

```suggestion
std::optional<std::string> secp256k1_private_to_public(std::span<const uint8_t> buffer)
{
	secp256k1_pubkey pubkey;
	size_t pubkey_len = SECP256K1_PUBKEY_COMPRESSED_LEN;
	std::vector<uint8_t> pubkey_compressed(pubkey_len);
	const uint8_t * privkey = buffer.data();

	if (!secp256k1_ec_seckey_verify(secp256k1_ctx, privkey)) {
		return std::nullopt;
	}
	if (!secp256k1_ec_pubkey_create(secp256k1_ctx, &pubkey, privkey)) {
		return "";
	}

	if (!secp256k1_ec_pubkey_serialize(secp256k1_ctx, pubkey_compressed.data(), &pubkey_len, &pubkey, SECP256K1_EC_COMPRESSED)) {
		return "";
	}

	return hex_encode(pubkey_compressed.data(), pubkey_len);
}
```

erickcestari (changes requested):

NishantBansal2003:
~~Okay, suppose secp256k1 is unable to produce a signature at any point in the code. How about returning an empty string in all such cases? That way, we align with the two possible outcomes from Decred’s code:~~
- ~~If Decred returns an empty string ("") or a nil char pointer (which we ignore), everything stays consistent -> we either compare two empty results or simply skip the Decred one.~~
- ~~If Decred returns a non empty value, comparing it against the empty secp256k1 output will clearly fail, which is the expected behaviour. (assuming secp256k1 is unable to produce the signature, then Decred should also not produce a signature for it, hope this assumption is safe to rely on?)~~

~~I also realized this approach can be applied to all other cases as well (ECDH, PrivateToPublicKey), since Decred is very lenient in each of these scenarios.~~

~~WDYT?~~

EDIT: Never mind, it is mentioned in `secp.PrivateKeyFromBytes` that: `This function primarily exists to provide a mechanism for converting serialized private keys that are already known to be good.`

NishantBansal2003:
Sorry for not pointing this out earlier, but looking at this again:

> The issue is that when signing with decred with always normalize the signature to lower-S form. We should call the `secp256k1_ecdsa_signature_normalize` function in the secp256k1 module after signing to have the same behavior.
> 
> ```c++
> int create_signature_from_hash(const uint8_t *privkey, const uint8_t *hash, secp256k1_ecdsa_signature& signature)
> {
> 	int ret = 0;
> 
> 	ret = secp256k1_ec_seckey_verify(secp256k1_ctx, privkey);
> 	ret = ret && secp256k1_ecdsa_sign(secp256k1_ctx, &signature, hash, privkey, nullptr, nullptr);
> 	ret = ret && secp256k1_ecdsa_signature_normalize(secp256k1_ctx, &signature, &signature); // add this line
> 
> 	return ret;
> }
> ```


`secp256k1_ecdsa_signature_normalize` returns `1 if sigin was not normalized, 0 if it already was` (https://github.com/bitcoin-core/secp256k1/blob/e7f7083b530a55c83ce9089a7244d2d9d67ac8b2/include/secp256k1.h#L589)
So with this line:
`ret = ret && secp256k1_ecdsa_signature_normalize(secp256k1_ctx, &signature, &signature);` we are essentially skipping the case where the signature is already normalized

Another thing is (https://github.com/bitcoin-core/secp256k1/blob/e7f7083b530a55c83ce9089a7244d2d9d67ac8b2/include/secp256k1.h#L658-L661), That means we are currently skipping all valid signature test cases from secp256k1, If both implementations follow low-S normalization and RFC6979, the signatures should match every time, so this is likely a bug.

WDYT?

> If you examine the Core Lightning code closely, you will see that it creates a loop until a signature with a low S is generated. (Bitcoin Core also does this: https://github.com/bitcoin/bitcoin/blob/bf69442b3f5004dc3df5a1b1d752114ba68fa5f4/src/key.cpp#L221)
> 
> ```c
> /* Grind for low R */
> 	do {
> 		ok = secp256k1_ecdsa_sign(secp256k1_ctx,
> 					  s,
> 					  h->sha.u.u8,
> 					  privkey->secret.data, NULL,
> 					  dev_no_signature_grind ? NULL
> 					  : extra_entropy);
> 		((u32 *)extra_entropy)[0]++;
> 		if (dev_no_signature_grind)
> 			break;
> 	} while (!sig_has_low_r(s));
> ```

Here it loops until it finds a low R and not low S. From what I found, this is done for size optimization, so that the tx size is reduced (and thus the tx fee).

erickcestari:
> Here it loops until it finds a low R and not low S. From what I found, this is done for size optimization, so that the tx size is reduced (and thus the tx fee).

True, my mistake. I read it quickly and completely forgot that it was talking about the R.  😅

> secp256k1_ecdsa_signature_normalize returns 1 if sigin was not normalized, 0 if it already was (https://github.com/bitcoin-core/secp256k1/blob/e7f7083b530a55c83ce9089a7244d2d9d67ac8b2/include/secp256k1.h#L589)
So with this line:
ret = ret && secp256k1_ecdsa_signature_normalize(secp256k1_ctx, &signature, &signature); we are essentially skipping the case where the signature is already normalized

> Another thing is (https://github.com/bitcoin-core/secp256k1/blob/e7f7083b530a55c83ce9089a7244d2d9d67ac8b2/include/secp256k1.h#L658-L661), That means we are currently skipping all valid signature test cases from secp256k1, If both implementations follow low-S normalization and RFC6979, the signatures should match every time, so this is likely a bug.

> WDYT?

Thanks for looking into this more deeply. I should have done the same. As you said, if the signatures are different, it's a bug, so we don't need to call `secp256k1_ecdsa_signature_normalize`.

NishantBansal2003:
> As you said, if the signatures are different, it's a bug, so we don't need to call `secp256k1_ecdsa_signature_normalize`.

Okay, I’m not sure about the details or the exact reason for the difference, but should I open an issue in Decred to discuss this further? Or should we wait considering the discussion in (https://bitcoin.stackexchange.com/questions/129270/why-do-different-implementations-produce-different-der-signatures-for-the-same-p ) and only open an issue if there’s actually a problem during signature verification across implementations?

NishantBansal2003:
I verified (with a small set of inputs) that even different signatures are being verified correctly, so let’s drop the `sign_compact` and `sign_der` functions, and I will add a `sign_verify` function using a custom mutator.

Marking this PR as draft for now.

erickcestari:
> > As you said, if the signatures are different, it's a bug, so we don't need to call `secp256k1_ecdsa_signature_normalize`.
> 
> Okay, I’m not sure about the details or the exact reason for the difference, but should I open an issue in Decred to discuss this further? Or should we wait considering the discussion in (https://bitcoin.stackexchange.com/questions/129270/why-do-different-implementations-produce-different-der-signatures-for-the-same-p ) and only open an issue if there’s actually a problem during signature verification across implementations?

I think it would be worth opening an issue to figure out what's going on.

It looks like Cryptofuzz has already found two issues similar to those we found in our fuzzing.

https://github.com/MozillaSecurity/cryptofuzz/blob/3d2377257129fc5da6effb92b0736e31db147dee/modules/decred/cryptofuzz.go#L238-L243
 
1. They skip if the private key is zero. If the hash and private keys are zero, the sign function will enter an infinite loop.
2. They skip if the private key is larger than the scalar 115792089237316195423570985008687907852837564279074904382605163141518161494337 

> I verified (with a small set of inputs) that even different signatures are being verified correctly, so let’s drop the `sign_compact` and `sign_der` functions, and I will add a `sign_verify` function using a custom mutator.
> 
> Marking this PR as draft for now.

I think this target is still worthwhile. The different signatures are a real bug. While that probably isn't dangerous, it means that one of the implementations isn't properly following RFC 6979 and low-S normalization.

brunoerg:
Has this been addressed? https://github.com/decred/dcrd/issues/3580#issuecomment-3619932956

modules/decredsecp256k1/decredsecp256k1_wrapper/wrapper.go
```diff
@@ -0,0 +1,115 @@
...
+//
+// While signatures produced from unreduced hashes are still cryptographically
+// valid and compliant with RFC 6979 section 3.6, this normalization ensures
+// deterministic signature matching across libsecp256k1 implementation.
+func NormalizeHashToCurveOrder(hashBytes []byte) []byte {
+	var scalar secp256k1.ModNScalar
+	scalar.SetByteSlice(hashBytes)
+	hash := scalar.Bytes()
+	return hash[:]
+}
```
NishantBansal2003, line 69: > Has this been addressed? [decred/dcrd#3580 (comment)](https://github.com/decred/dcrd/issues/3580#issuecomment-3619932956)

Yes, addressed here

davecgh:
> It looks like Cryptofuzz has already found two issues similar to those we found in our fuzzing.
> 
> https://github.com/MozillaSecurity/cryptofuzz/blob/3d2377257129fc5da6effb92b0736e31db147dee/modules/decred/cryptofuzz.go#L238-L243
> 
>     1. They skip if the private key is zero. If the hash and private keys are zero, the sign function will enter an infinite loop.
>     2. They skip if the private key is larger than the scalar 115792089237316195423570985008687907852837564279074904382605163141518161494337

To provide some additional details here, these are both well documented requirements to use the [PrivKeyFromBytes](https://pkg.go.dev/github.com/decred/dcrd/dcrec/secp256k1/v4#PrivKeyFromBytes) method which fuzzing code often does.

Its documentation explicitly states:

> ...
> WARNING: This means passing a slice with more than 32 bytes is truncated and that truncated value is reduced modulo N. Further, 0 is not a valid private key. It is up to the caller to provide a value in the appropriate range of [1, N-1]. Failure to do so will either result in an invalid private key or potentially weak private keys that have bias that could be exploited.
> This function primarily exists to provide a mechanism for converting serialized private keys that are already known to be good.
>
> Typically callers should make use of GeneratePrivateKey or GeneratePrivateKeyFromRand when creating private keys since they properly handle generation of appropriate values.

So, they aren't issues, rather documented preconditions and Cryptofuzz is doing the correct thing by ensuring the preconditions are satisfied.

> ... The different signatures are a real bug. While that probably isn't dangerous, it means that one of the implementations isn't properly following RFC 6979 and low-S normalization.

Already discussed in the [issue](https://github.com/decred/dcrd/issues/3580) NishanBansal2003 opened, but for the sake of having the referenced information all in the same place, this isn't a bug and it's not due to low-S normalization.

It is because RFC6979 does not guarantee identical signatures between implementations because it explicitly permits the use of different variants in [section 3.6](https://www.rfc-editor.org/rfc/rfc6979#section-3.6) and those variants will result in different signatures.  So, it only guarantees deterministic signatures given the same set of inputs with the same implementation (or, more precisely, implementations using the exact same variants). 

In other words, two different implementations producing different, but still independently deterministic, signatures is not a bug or an issue in terms of security.

The specific difference that is being observed here is that both implementations use different variants the specification explicitly calls out as valid.  In particular, Decred's code uses this variant in [section 3.6](https://www.rfc-editor.org/rfc/rfc6979#section-3.6):

> - It is possible to use H(m) directly, instead of bits2octets(H(m)), as part of the HMAC input.  As explained in [Section 3.5](https://www.rfc-editor.org/rfc/rfc6979#section-3.5), we use bits2octets(H(m)) in order to ease integration into systems that already use an (EC)DSA signature engine by sending it an already- truncated hash value.  **Using the whole H(m) does not introduce any vulnerability.**

Emphasis mine.

On the other hand, libsecp256k1 uses the ` bit2octets(H(m))` variant.

In other words, libsecp256k1 reduces the hash modulo the secp256k1 curve order prior to feeding it into the HMAC whereas Decred's secp256k1 does not.

In short, both implementations are compliant with the spec and both are producing cryptographically valid signatures.

Given all of that, along with the fact that ECDSA requires the hash to be reduced modulo the secp256k1 curve order already prior to signing, the deterministic signatures can be made to match by ensuring they both use the same variant for the deterministic nonce generation which can be achieved by normalizing the hash by reducing it modulo the secp256k1 curve order prior to calling `Sign`.

modules/decredsecp256k1/decredsecp256k1_wrapper/wrapper.go
brunoerg: Perfect. Nice comment explaining it as well.

modules/decredsecp256k1/Makefile
```diff
@@ -0,0 +1,16 @@
+CXXFLAGS += -Wall -Wextra -std=c++20 -I. -I ../../include -I./decredsecp256k1_wrapper
```
brunoerg, line 1: You probably wrote this Makefile based on how we build `btcd`, which is fine, but I just changed that one and would be nice to have this updated accordingly. See: https://github.com/bitcoinfuzz/bitcoinfuzz/pull/352/files

modules/decredsecp256k1/decredsecp256k1_wrapper/wrapper.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/decred/dcrd/dcrec/secp256k1/v4"
+	secp "github.com/decred/dcrd/dcrec/secp256k1/v4"
+	"github.com/decred/dcrd/dcrec/secp256k1/v4/ecdsa"
+)
+
+//export DecredPrivateToPublicKey
+func DecredPrivateToPublicKey(privKeyData C.ByteArray) *C.char {
+	privKeyBytes := C.GoBytes(unsafe.Pointer(privKeyData.data), privKeyData.length)
+
+	privKey := secp.PrivKeyFromBytes(privKeyBytes)
```
erickcestari: Same here

```suggestion
func DecredPrivateToPublicKey(privKeyData C.ByteArray) *C.char {
	privKeyBytes := C.GoBytes(unsafe.Pointer(privKeyData.data), privKeyData.length)
	
		if isAllZeros(privKeyBytes) {
		return nil
	}

	privKey := secp.PrivKeyFromBytes(privKeyBytes)
```

modules/decredsecp256k1/decredsecp256k1_wrapper/wrapper.go
```diff
@@ -0,0 +1,115 @@
...
+
+//export DecredECDH
+func DecredECDH(privKeyData C.ByteArray) *C.char {
+	var (
+		pubJacobian secp.JacobianPoint
+		s           secp.JacobianPoint
+	)
+
+	privKeyBytes := C.GoBytes(unsafe.Pointer(privKeyData.data), privKeyData.length)
+
```
erickcestari, line 137: Same here

```suggestion
func DecredECDH(privKeyData C.ByteArray) *C.char {
	var (
		pubJacobian secp.JacobianPoint
		s           secp.JacobianPoint
	)

	privKeyBytes := C.GoBytes(unsafe.Pointer(privKeyData.data), privKeyData.length)
	
		if isAllZeros(privKeyBytes) {
		return nil
	}

```

modules/decredsecp256k1/decredsecp256k1_wrapper/wrapper.go
```diff
@@ -0,0 +1,115 @@
...
+	signStr := hex.EncodeToString(sign)
+
+	return C.CString(signStr)
+}
+
+//export DecredSignVerify
+func DecredSignVerify(privKeyData C.ByteArray, hashData C.ByteArray, signData C.ByteArray) C.int {
+	privKeyBytes := C.GoBytes(unsafe.Pointer(privKeyData.data), privKeyData.length)
+	hashBytes := C.GoBytes(unsafe.Pointer(hashData.data), hashData.length)
+	signBytes := C.GoBytes(unsafe.Pointer(signData.data), signData.length)
```
erickcestari, line 109: Same here

```suggestion
func DecredSignVerify(privKeyData C.ByteArray, hashData C.ByteArray, signData C.ByteArray) C.int {
	privKeyBytes := C.GoBytes(unsafe.Pointer(privKeyData.data), privKeyData.length)
	hashBytes := C.GoBytes(unsafe.Pointer(hashData.data), hashData.length)
	signBytes := C.GoBytes(unsafe.Pointer(signData.data), signData.length)
	
	
		if isAllZeros(privKeyBytes) {
		return nil
	}
```

modules/decredsecp256k1/decredsecp256k1_wrapper/wrapper.go
```diff
@@ -0,0 +1,115 @@
...
+	hashBytesMod := NormalizeHashToCurveOrder(hashBytes)
+	sign := ecdsa.SignCompact(privKey, hashBytesMod, false)
+	signStr := hex.EncodeToString(sign[1:])
+
+	return C.CString(signStr)
+}
+
+//export DecredSignDER
+func DecredSignDER(privKeyData C.ByteArray, hashData C.ByteArray) *C.char {
+	privKeyBytes := C.GoBytes(unsafe.Pointer(privKeyData.data), privKeyData.length)
```
erickcestari, line 90: Same here

```suggestion
func DecredSignDER(privKeyData C.ByteArray, hashData C.ByteArray) *C.char {
	privKeyBytes := C.GoBytes(unsafe.Pointer(privKeyData.data), privKeyData.length)
	
	if isAllZeros(privKeyBytes) {
		return nil
	}
```

modules/decredsecp256k1/decredsecp256k1_wrapper/wrapper.go
```diff
@@ -0,0 +1,115 @@
...
+	scalar.SetByteSlice(hashBytes)
+	hash := scalar.Bytes()
+	return hash[:]
+}
+
+//export DecredSignCompact
+func DecredSignCompact(privKeyData C.ByteArray, hashData C.ByteArray) *C.char {
+	privKeyBytes := C.GoBytes(unsafe.Pointer(privKeyData.data), privKeyData.length)
+	hashBytes := C.GoBytes(unsafe.Pointer(hashData.data), hashData.length)
+
```
erickcestari, line 75: Same here

```suggestion
	privKeyBytes := C.GoBytes(unsafe.Pointer(privKeyData.data), privKeyData.length)
	hashBytes := C.GoBytes(unsafe.Pointer(hashData.data), hashData.length)

if isAllZeros(privKeyBytes) {
		return nil
	}
```

erickcestari (changes requested):
> WARNING: This means passing a slice with more than 32 bytes is truncated and that truncated value is reduced modulo N. Further, 0 is not a valid private key.

Let's satisfy the precondition of not providing an invalid private key 0.

```go
func isAllZeros(b []byte) bool {
	for _, v := range b {
		if v != 0 {
			return false
		}
	}
	return true
}
```

.github/workflows/workflow.yml
NishantBansal2003: I’m assuming you meant to add `git submodule update --init --recursive external/secp256k1` to the README.md? Because I don’t see this in the workflow.yml for CLN

NishantBansal2003:
> Let's satisfy the precondition of not providing an invalid private key 0.
> 
> ```go
> func isAllZeros(b []byte) bool {
> 	for _, v := range b {
> 		if v != 0 {
> 			return false
> 		}
> 	}
> 	return true
> }
> ```

 I don’t think this is required, since we are already doing this:
```c
if (!secp256k1_ec_seckey_verify(secp256k1_ctx, privkey))
{
	return std::nullopt;
}
```

With this, we skip comparing cases where the private key is invalid

erickcestari:
> I don’t think this is required, since we are already doing this:
> 
> ```c
> if (!secp256k1_ec_seckey_verify(secp256k1_ctx, privkey))
> {
> 	return std::nullopt;
> }
> ```
> 
> With this, we skip comparing cases where the private key is invalid

The main reason is that we should use the Decred secp256k1 library correctly. An infinite loop occurs when the library tries to sign with a zero private key and a zero hash. If we don't follow the preconditions, we'll encounter fake bugs.

davecgh:
In general, I'd suggest something like the following wrapper code any time you are trying to use raw bytes that are not already known to be a valid private key, which obviously applies to the case with fuzzing input.  That way, all instances that are attempting to convert arbitrary bytes can use the same func and be assured they're always dealing with a valid private key.

```Go
func ParsePrivKey(privKeyBytes []byte) (*secp256k1.PrivateKey, bool) {
	if len(privKeyBytes) > 32 {
		return nil, false
	}
	var key secp256k1.ModNScalar
	overflows := key.SetByteSlice(privKeyBytes)
	if overflows || key.IsZero() {
		return nil, false
	}
	return secp256k1.NewPrivateKey(&key), true
}

....
privKeyBytes := C.GoBytes(unsafe.Pointer(privKeyData.data), privKeyData.length)
privKey, valid := ParsePrivKey(privKeyBytes)
if !valid {
	return nil
}
```

EDIT: As a bit of an aside, I think it is probably be worthwhile to consider changing the API in the Decred Go lib in a future version to provide a constant time implementation for working directly with raw bytes which avoids the potential footguns altogether by refusing to create a private key when the raw bytes are invalid rather than solely documenting the preconditions and putting the onus on the caller.

brunoerg:
Needs rebase. I hope this is the last one, will do a final review here, sorry for the noise.

.github/workflows/workflow.yml
brunoerg: Yes, README.md, sorry I commented on the incorrect place.

custommutator/utils/signature.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+	if (!ret)
+	{
+		return size;
+	}
+
+	// Append signature after first 64 bytes
+	memcpy(data + MIN_BUFFER_SIZE, signature_der, signature_der_len);
+	return MIN_BUFFER_SIZE + signature_der_len;
+}
```
erickcestari: We could put this inside `secp256k1_signature.h`, since it probably won't be reused in other custom mutators.

modules/secp256k1/module.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+  ret = ret = secp256k1_ec_pubkey_create(secp256k1_ctx, &pubkey, privkey);
+  ret = ret && secp256k1_ecdh(secp256k1_ctx, shared_secret.data(), &pubkey,
+                              privkey, nullptr, nullptr);
+  if (!ret) {
+    return "";
+  }
+
+  return hex_encode(shared_secret.data(), SECP256K1_SHARED_SECRET_LEN);
+}
```
erickcestari, line 150: This target could be improved by performing an ECDH between a private key and a public key provided by the fuzzer, rather than deriving the public key from the private key.

driver.cpp
```diff
@@ -468,6 +470,159 @@ void Driver::Bip32MasterKeygenTarget(std::span<const uint8_t> buffer) const {
...
+    last_module_name = module.first;
+  }
+}
+
+void Driver::ECDHTarget(std::span<const uint8_t> buffer) const {
+  FuzzedDataProvider provider(buffer.data(), buffer.size());
+  if (buffer.size() < 32)
+    return;
+
+  std::vector<uint8_t> privkey_buffer = provider.ConsumeBytes<uint8_t>(32);
```
erickcestari, line 602: This target would need at least 65 bytes (32  private key + 33 compressed public key)

erickcestari (changes requested):
It's looking good! Just a few more changes, and it'll be good to go!

Note: This one is already close to finishing, so splitting it now isn’t necessary. For future work, breaking similar changes into separate PRs for each target can make the process smoother.

modules/secp256k1/module.cpp
brunoerg: > This target could be improved by performing an ECDH between a private key and a public key provided by the fuzzer, rather than deriving the public key from the private key.

Hm, it would be hard (near impossible) to the fuzzer to generate matching public and private keys. Wouldn't it?

NishantBansal2003: Nice enhancement...

> Hm, it would be hard (near impossible) to the fuzzer to generate matching public and private keys. Wouldn't it?


I don’t think we need a matching private key for that public key. We can simply use the fuzzer generated public key and call `secp.ParsePubKey()`(If the key is invalid, this function will return an error). This also lets us test how both handles invalid public keys.

erickcestari: > > This target could be improved by performing an ECDH between a private key and a public key provided by the fuzzer, rather than deriving the public key from the private key.
> 
> Hm, it would be hard (near impossible) to the fuzzer to generate matching public and private keys. Wouldn't it?

It doesn't need to generate matching keys. This is a target for ECDH, not signature verification.
In ECDH, two parties each have their own key pair.  Both arrive at the same shared secret due to the properties of elliptic curve multiplication: (a × G) × b = (b × G) × a
So for fuzzing ECDH, we just need any valid private key and any valid public key (from different key pairs). The operation will produce a shared secret regardless of whether the keys "match."

custommutator/utils/signature.cpp
NishantBansal2003: The reason for not doing this is that we would need to add `-I external/secp256k1/include` to the main Makefile, otherwise it will complain secp256k1.h is not found. I don’t think we should add this include to the main Makefile since it may not be needed in most cases.

But I have no strong preference -- if everyone agrees, I can change it as well

erickcestari: If we include the direct path to the `.h` file, `#include <external/secp256k1/include/secp256k1.h>`, we will not need the `-I external/secp256k1/include`.

You can remove the `signature.cpp` and `signature.h` and paste this inside the `secp256k1_signature.h`
```cpp
#ifndef BITCOINFUZZ_CUSTOMMUTATOR_SECP256K1_SIGNATURE_H
#define BITCOINFUZZ_CUSTOMMUTATOR_SECP256K1_SIGNATURE_H

#define template c_template

extern "C" {
#include <external/secp256k1/include/secp256k1.h>
}

#undef template

#include <cstddef>
#include <cstring>
#include <span>

static constexpr size_t MIN_BUFFER_SIZE = 64;
static constexpr size_t MAX_SIGNATURE_DER_SIZE = 74;

static secp256k1_context *g_secp256k1_ctx = nullptr;

static secp256k1_context *get_secp256k1_context() {
  if (!g_secp256k1_ctx) {
    g_secp256k1_ctx = secp256k1_context_create(SECP256K1_CONTEXT_SIGN |
                                               SECP256K1_CONTEXT_VERIFY);
  }
  return g_secp256k1_ctx;
}

extern "C" size_t LLVMFuzzerMutate(uint8_t *Data, size_t Size, size_t MaxSize);

extern "C" size_t LLVMFuzzerCustomMutator(uint8_t *fuzz_data, size_t size,
                                          size_t max_size,
                                          [[maybe_unused]] unsigned int seed) {
  size_t new_size = LLVMFuzzerMutate(fuzz_data, size, max_size);

  if (new_size < MIN_BUFFER_SIZE) {
    return new_size;
  }

  secp256k1_context *ctx = get_secp256k1_context();
  if (!ctx) {
    return new_size;
  }

  secp256k1_ecdsa_signature signature;
  size_t signature_der_len = MAX_SIGNATURE_DER_SIZE;
  unsigned char signature_der[MAX_SIGNATURE_DER_SIZE];

  std::span<const uint8_t> buffer(fuzz_data, MIN_BUFFER_SIZE);
  std::span<const uint8_t> privkey = buffer.subspan(0, 32);
  std::span<const uint8_t> hash = buffer.subspan(32, 32);

  int ret = secp256k1_ec_seckey_verify(ctx, privkey.data());
  ret = ret && secp256k1_ecdsa_sign(ctx, &signature, hash.data(),
                                    privkey.data(), nullptr, nullptr);
  ret = ret && secp256k1_ecdsa_signature_serialize_der(
                   ctx, signature_der, &signature_der_len, &signature);

  if (ret && (MIN_BUFFER_SIZE + signature_der_len <= max_size)) {
    memcpy(fuzz_data + MIN_BUFFER_SIZE, signature_der, signature_der_len);
    return MIN_BUFFER_SIZE + signature_der_len;
  }

  return new_size;
}

#endif
```

modules/secp256k1/module.cpp
brunoerg: Yes, you're right, didn't notice that was the ECDH target. So yes, better to address @erickcestari's suggestion. Let the fuzzer create both keys.

erickcestari (approved):
Thanks!

ACK d5aeca581fba35ee12586312b3e24b83105b9312

PR #344, opened 2025-12-02
feat: add nix flake

GustavoStingelin:


quapka:
Great that you've started working on Nix inclusion. I've [mentioned it briefly](https://github.com/bitcoinfuzz/bitcoinfuzz/issues/272#issuecomment-3292610739), but unfortunately did not have much time to work on it.

I was thinking to do something similar to what I did in [ECTester](https://github.com/crocs-muni/ECTester/blob/master/flake.nix). The setup is exploiting a bit how Nix Flakes work to allow building almost arbitrary (pre-fetched) version of certain libs (see the JSON files [here](https://github.com/crocs-muni/ECTester/tree/master/nix) that reference the versions). Something similar would allow running `bitcoinfuzz` on a particular version, instead of only a single one.

I am happy to help/discuss the approach.

.github/workflows/nix.yml
```diff
@@ -0,0 +1,46 @@
...
+            ${{ runner.os }}-cargo-
+
+      - name: nix flake check
+        run: nix flake check --show-trace
+
+      - name: nix develop
+        run: nix develop --command bash -c "echo devshell OK"
+
+      - name: auto_build.sh
+        run: nix develop --command bash -lc 'CXXFLAGS="-DLDK -DLND" ./auto_build.sh'
```
brunoerg: Why only LDK and LND?

brunoerg:
I personally don't use Nix, so I'd leave this PR to be reviewed by any Nix user. Glad @quapka may help with this.

brunoerg:
Also, I expect you guys to commit to keep it working since we frequently do changes that affect building.

.github/workflows/nix.yml
GustavoStingelin: MB, will add others

joaozinhom:
tACK https://github.com/bitcoinfuzz/bitcoinfuzz/pull/344/commits/3bc330b035eb257e9ff06da957fe4f54875f81a4 tested on my NixOS(25.05.813260.6c8f0cca8451 (Warbler) x86_64):
AMD Ryzen 7 7730U
AMD ATI Barcelo
16gb RAM

erickcestari:
Should we also test whether it can run on the GitHub Runner MacOS machine? I've tried to add nix sometime ago, but it always failed in macos for some reason.

brunoerg:
> Should we also test whether it can run on the GitHub Runner MacOS machine? I've tried to add nix sometime ago, but it always failed in macos for some reason.

No, and we should drop support for macOS at all, but this is discussion for another issue.

erickcestari:
> Great that you've started working on Nix inclusion. I've [mentioned it briefly](https://github.com/bitcoinfuzz/bitcoinfuzz/issues/272#issuecomment-3292610739), but unfortunately did not have much time to work on it.
> 
> I was thinking to do something similar to what I did in [ECTester](https://github.com/crocs-muni/ECTester/blob/master/flake.nix). The setup is exploiting a bit how Nix Flakes work to allow building almost arbitrary (pre-fetched) version of certain libs (see the JSON files [here](https://github.com/crocs-muni/ECTester/tree/master/nix) that reference the versions). Something similar would allow running `bitcoinfuzz` on a particular version, instead of only a single one.
> 
> I am happy to help/discuss the approach.

I'm also not a Nix user. Could you explain how we could make this possible in BitcoinFuzz?

Some of our modules are Git submodules, some are copies of files (like Bitcoin Core), and some use the default package manager of the programming language.

jaoleal:
> Also, I expect you guys to commit to keep it working since we frequently do changes that affect building.

Sorry to be nosy about this Pull Request but,

My experience maintaining Nix on floresta tells me that might be not a good idea to have `nix flake check` in CI... Or even nix at all in bitcoinfuzz, I say this specially because bitcoinfuzz doesnt have any maintainer thats willing to stick with nix that much correct ?

Nix is hard to maintain, documentation laks and the language is hard to get used to because is soooooo counterintuitive. Its frequently a problem and often i get yelled at floresta because of it... (Im working on a PR to remove some nix from it https://github.com/vinteumorg/Floresta/pull/718).

I am not satisfied with the current approach, is too simple and heavy to load because its importing EVERYTHING that one MIGTH need... IMHO the ideal approach would be too hard to maintain, just by taking a quick look at the derivation that @quapka mentioned i can already see the hours being spended to make something that **IDEALLY** fits this case.

I see that this specific project may have a lot of complications and may be impossible to use it on a NixOs machine as it is, or in a nix sustained environment but, does it need to ? 

AFAIK this project isnt a product that will have users and yes runners... Probably the maintainers that will have a proper setup to run it... [Theyre even dropping support to MacOs](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/344#issuecomment-3607765011) and i imagine that this is because of the complexitity involving dependencies.

Nix is for development and packaging, only development is needed in this case right ?... I did a quick looking and i expect that the DockerFile might be well maintained and nix does have a great support for containers... If one NEEDS nix supports im pretty convinced that might have an approach that does not include inserting Nix in this codebase as well avoiding choking the project because of this, trust me it happens a lot and itll be a mess for bitcoinfuzz.

This is pretty much a NACK for me.

jaoleal:
Also, nix is not for every project. 

A solution for everyone is a solution for no one, just see how python and js stinks

GustavoStingelin:
@jaoleal so, are you saying that using only Docker would be enough? Johnny is building a new Docker version.

GustavoStingelin:
FYI @johnnyasantoss

GustavoStingelin:
#348

quapka:
On one hand, as @jaoleal point out, Nix is not a free ride, no hiding here. On the other hand, my point was not to just build via Nix, but add _runtime_, i.e., no source-code changes or Git-fu required, configuration of which version (_read_ commit) of a module is used for fuzzing. (Another thing, still doable in principal, would be to test the same module in two and more versions against each other.)

> I'm also not a Nix user. Could you explain how we could make this possible in BitcoinFuzz?
> Some of our modules are Git submodules, some are copies of files (like Bitcoin Core), and some use the default package manager of the programming language.

Think of Nix as a _stricter_ language-agnostic build derivation environment. You define your dependencies (that are built in similar _strict_ manners themselves), sources and then build steps. Finally, executing this definition gives you outputs. If you are familiar with lock files in various languages, thing of Nix (Flakes) as language-agnostic lock files (locking many more dependencies than the language-specific ones).

Compared to this PR and [_all in one_](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/344/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0) kind of structure, where @GustavoStingelin creates a shell (like literal Bash shell) with the dependencies for _all_ (in principal) modules together, I tried to separate the build into custom "shells" per module (or mutator and also one for bitcoinfuzz itself). Both have pros and cons.

Taking bitcoincore as an example (in my branch) in [this derivation definition](https://github.com/quapka/bitcoinfuzz/blob/build-with-nix/modules/bitcoin/default.nix), [this](https://github.com/quapka/bitcoinfuzz/blob/build-with-nix/modules/bitcoin/default.nix#L1-L9) defines the dependencies for the build. [Here](https://github.com/quapka/bitcoinfuzz/blob/build-with-nix/modules/bitcoin/default.nix#L23-L28) the sources are constructed (filtering any irrelevant files [here](https://github.com/quapka/bitcoinfuzz/blob/build-with-nix/modules/bitcoin/default.nix#L23-L28)). The actual `Makefile` is reused, and the existing `make` is called [here](https://github.com/quapka/bitcoinfuzz/blob/build-with-nix/modules/bitcoin/default.nix#L42-L46) (the `-j$(nproc)` is left there from my testing...). Finally, the `module.a` [is copied](https://github.com/quapka/bitcoinfuzz/blob/build-with-nix/modules/bitcoin/default.nix#L50-L53) into a special `$out` directory, which is the entry point for all _artifacts_ from this derivation. This bitcoin-core derivation is called [here](https://github.com/quapka/bitcoinfuzz/blob/build-with-nix/flake.nix#L26) and the result is  [made available](https://github.com/quapka/bitcoinfuzz/blob/build-with-nix/flake.nix#L23) to `bicoinfuzz`, by copying the result of the `bitcoin-core` derivation, i.e., its `module.a`, where `bitcoinfuzz` expects it (again, hooking to the existing Makefile). Bitcoinfuzz itself [is built](https://github.com/quapka/bitcoinfuzz/blob/build-with-nix/default.nix) similarly in it's _own_ derivation.

Now, with Nix installed (and Nix Flakes feature enabled), the `bitcoinfuzz` can be called with (currently, there is lack of the module picking via `CXXFLAGS`):
```bash
FUZZ=bip32_master_keygen nix run
```

Taking a step back, the [`bitcoin-core` derivation](https://github.com/quapka/bitcoinfuzz/blob/build-with-nix/modules/bitcoin/default.nix) can be changed to not rely on the local bitcoin-core-related  files, but make it dependent on [`bitcoin` from Nixpkgs](https://github.com/NixOS/nixpkgs/blob/nixos-25.11/pkgs/applications/blockchains/bitcoin/default.nix#L198) (probably the most logical way), in such a way that the `version` and `src` can be overriden by a parameter.

Due to the added complexity of not choosing only a module, but also it's revision, I would expect the "final product", including different versions per lib, to read a `config.json` (for the sake of sanity) and execute all of these:
```json
[
  {
    "module": "bitcoin",
    "rev": "<bitcoin-commit>",
  },
  {
    "module": "bitcoin",
    "rev": "<differnt-bitcoin-commit>",
  },
  {
    "module": "rustbitcoin",
    "rev": "<rust-commit>",
  }
]
```

Come to think of it, running different versions of the same module maybe be tricky due to the same symbols. Anyways, this would work for different version of different modules.

quapka:
Testing _this_ PR on (bit outdated) system:
```bash
$ nixos-version
24.11.718912.1f426f65ac4e (Vicuna)
$ hwinfo | grep Model
  Model: 25.116.1 "AMD Ryzen 7 PRO 7840U w/ Radeon 780M Graphics"
```

I was to `./auto_build.sh` with `CXXFLAGS="-DBITCOIN_CORE -DRUST_BITCOIN`, but `-DNBITCOIN` failed with the following:

```text
Executing in modules/nbitcoin: make
/nix/store/sqlnjj8c3n3si3sjnadhdbcwgrk97g2w-clang-wrapper-21.1.2/bin/clang++ -DBITCOIN_CORE -DRUST_BITCOIN -DNBITCOIN -Wall -Wextra -fsanitize=address,fuzzer -std=c++20 -I ../../include -I . -fPIC -c module.cpp -o module.o
dotnet publish -c Release -r linux-x64 --self-contained
  Determining projects to restore...
  Restored REDACTED/projects/bitcoinfuzz/modules/nbitcoin/NBitcoin.CppBridge.csproj (in 7.56 sec).
  NBitcoin.CppBridge -> REDACTED/projects/bitcoinfuzz/modules/nbitcoin/bin/Release/net9.0/linux-x64/NBitcoin.CppBridge.dll
  Generating native code
  Could not start dynamically linked executable: REDACTED/.nuget/packages/runtime.linux-x64.microsoft.dotnet.ilcompiler/9.0.11/tools/ilc
  NixOS cannot run dynamically linked executables intended for generic
  linux environments out of the box. For more information, see:
  https://nix.dev/permalink/stub-ld
REDACTED/.nuget/packages/microsoft.dotnet.ilcompiler/9.0.11/build/Microsoft.NETCore.Native.targets(317,5): error MSB3073: The command ""REDACTED/.nuget/packages/runtime.linux-x64.microsoft.dotnet.ilcompiler/9.0.11/tools/ilc" @"obj/Release/net9.0/linux-x64/native/NBitcoin.CppBridge.ilc.rsp"" exited with code 127. [REDACTED/projects/bitcoinfuzz/modules/nbitcoin/NBitcoin.CppBridge.csproj]
make: *** [Makefile:16: bin/Release/net9.0/linux-x64/native/NBitcoin.CppBridge.so] Error 1
Error: Failed to execute in modules/nbitcoin
```

GustavoStingelin:
@quapka `programs.nix-ld.enable = true;` [ref](https://github.com/GustavoStingelin/nix-machinary/blob/98f794f82a8f9cb58ebc14726b5d39e244e25898/hosts/reapermobile/default.nix#L80) should solve the `NixOS cannot run dynamically linked executables intended for generic`, maybe we would need to add [FHS](https://www.alexghr.me/blog/til-nix-flake-fhs/) to the dev shell

quapka:
> @quapka `programs.nix-ld.enable = true;` [ref](https://github.com/GustavoStingelin/nix-machinary/blob/98f794f82a8f9cb58ebc14726b5d39e244e25898/hosts/reapermobile/default.nix#L80) should solve the `NixOS cannot run dynamically linked executables intended for generic`, maybe we would need to add [FHS](https://www.alexghr.me/blog/til-nix-flake-fhs/) to the dev shell

Thanks, I might be overly on the Nix-side, but isn't `nix-ld.enable` nad FHS actually going against the benefits Nix gives?

Anyways, I have added also [LND, LDK, BTCD](https://github.com/bitcoinfuzz/bitcoinfuzz/compare/v2...quapka:bitcoinfuzz:build-with-nix) and [NBitcoin](https://github.com/quapka/bitcoinfuzz/blob/build-with-nix/modules/nbitcoin/default.nix) (without the `nix-ld` or FHS).

[UPDATE] The NBitcoin currently fails due to the `NBitcoin.CppShared.so` expected on the relative `./NBitcoin.CppShared.so` path. This should be [patch-able](https://nixos.wiki/wiki/Packaging/Binaries).

GustavoStingelin:
Replaced by #348.

Sorry to everyone who prefers Nix, and thanks a lot for all the comments and feedback. For now, Docker won the build decision because it is simpler to maintain, so I am closing this. If there is interest, we can open a separate discussion about Nix versus Docker and go deeper into the tradeoffs.

PR #349, opened 2025-12-04
style: add automated code formatters

### LINKED ISSUE #339: Add Automated Code Formatters
We currently don't use automated code formatters, leading to inconsistent formatting and manual formatting effort.

Should we establish one for each language? Or establish one just for the bitcoinfuzz framework files (cpp) for now?

erickcestari:
I've used the black formatter to format python code, since pylint doesn't work as a formatter.

Formatters used:

clang-format: for C code, header files and C++ code.
black: python
ktlint: kotlin
google-java-format: java
dotnet format: C#
rustfmt: rust
gofmt: go

Closes: #339


modules/bitcoin/module.cpp
```diff
@@ -87,7 +85,8 @@ struct TestData {
     std::map<std::vector<unsigned char>, std::vector<unsigned char>> hash160_preimages;
 
     //! Set the precomputed data.
-    void Init() {
+    void Init()
+    {
```
morehouse: Why does `clang-format` put the curly brace on a new line for functions in this file but not others?

erickcestari: > Why does `clang-format` put the curly brace on a new line for functions in this file but not others?

Because of the file [.clang-format](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/modules/bitcoin/.clang-format) file inside the bitcoin module. The others are being formatted with the 'LLVM" default style 

Should we force to be in the LLVM style? Using the command `clang-format --style="LLVM" -i ./module.cpp ./module.h`?

morehouse: No, I don't think so.  Most of these files are copied from Bitcoin Core, and we don't want to have to reformat every time they're updated.

morehouse (approved):
LGTM

modules/bitcoin/module.cpp
erickcestari: > No, I don't think so. Most of these files are copied from Bitcoin Core, and we don't want to have to reformat every time they're updated.

The bitcoin core will not be formatted. It'll only format bitcoinfuzz code, in case of bitcoin core module it will format module.cpp and module.h

erickcestari: I've removed the clang-format file and formatted with the default clang-format. If you want, I can undo.

morehouse (approved):
ACK 7f9000564a84b8af4076d23b789d902727719d1f

brunoerg (approved):
code review ACK 7f9000564a84b8af4076d23b789d902727719d1f

brunoerg:
CI failure is unrelated.

PR #351, opened 2025-12-05
Add new lightning messages to the target parse_p2p_lightning_message

erickcestari:


modules/clightning/module.cpp
```diff
@@ -643,6 +643,25 @@ std::optional<std::string> clightning_parse_p2p_lightning_message(std::span<cons
...
             result << ";BLINDED_PATH=" << fmt_pubkey(tmpctx, update_add_htlc->blinded_path);
         }
+    } else if (msg_type == WIRE_UPDATE_FULFILL_HTLC) {
+        channel_id channel;
+        u64 id;
+        preimage payment_preimage;
+
+        // Skip messages that are bigger than 74 bytes (the maximum size of a
+        // update_fulfill_htlc). Since rust-lightning returns an error for
+        // messages that are too big.
```
morehouse: Ever since [attributable failures](https://github.com/lightning/bolts/pull/1044) landed, I don't think this is true anymore.

Also applies to other comments below.

modules/clightning/module.cpp
```diff
@@ -643,6 +643,66 @@ std::optional<std::string> clightning_parse_p2p_lightning_message(std::span<cons
...
+        u64 id;
+        u8 *reason;
+
+        if (!fromwire_update_fail_htlc(tmpctx, msg, &channel, &id, &reason)) {
+            return "";
+        }
+
+        // Skip messages that are bigger than 74 bytes (the maximum size of a
+        // update_fulfill_htlc). Since rust-lightning returns an error for
+        // messages that are too big.
```
morehouse: This comment is wrong -- 74 bytes is not accurate, and this is an `update_fail_htlc` message.

modules/clightning/module.cpp
```diff
@@ -643,6 +643,66 @@ std::optional<std::string> clightning_parse_p2p_lightning_message(std::span<cons
...
+        if (tal_bytelen(msg) - tal_bytelen(reason) > + 44) {
```
morehouse:
```suggestion
        if (tal_bytelen(msg) - tal_bytelen(reason) > 44) {
```

modules/ldk/.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[build]
+rustflags = ["--cfg", "fuzzing"]
```
morehouse, line 2: Why is this now required?

modules/clightning/module.cpp
```diff
@@ -643,6 +643,66 @@ std::optional<std::string> clightning_parse_p2p_lightning_message(std::span<cons
...
+        sha256 sha256_of_onion; 
+        u16 failure_code;
+
+        if (!fromwire_update_fail_malformed_htlc(msg, &channel, &id, &sha256_of_onion, &failure_code)) {
+            return "";
+        }
+
+        // Skip messages that are bigger than 74 bytes (the maximum size of a
+        // update_fulfill_htlc). Since rust-lightning returns an error for
+        // messages that are too big.
```
morehouse: Byte length and message type also don't match here.

modules/clightning/module.cpp
```diff
@@ -643,6 +643,66 @@ std::optional<std::string> clightning_parse_p2p_lightning_message(std::span<cons
...
+        if (tal_bytelen(msg) > + 76) {
```
morehouse:
```suggestion
        if (tal_bytelen(msg) > 76) {
```

modules/clightning/module.cpp
erickcestari: I updated the comments to be a `TODO` since `LND` and `CLN` hadn't implemented it yet.

modules/clightning/module.cpp
erickcestari: Updated

modules/clightning/module.cpp
erickcestari: Fixed

modules/clightning/module.cpp
erickcestari: Fixed

modules/clightning/module.cpp
erickcestari: Updated

modules/ldk/.cargo/config.toml
erickcestari: `OnionErrorPacket` is only available when the `fuzzing` flag is used. Adding the `.cargo/config.toml` file causes the rust-analyzer to compile the code with that flag, which should eliminate errors in the IDE. This is because the rust-analyzer compiles in the same way as Cargo with the `--cfg fuzzing` flag.

https://github.com/lightningdevkit/rust-lightning/blob/f6b4f0c41b9b8e4e5fff8c2dd2de08c8e1ece980/lightning/src/ln/msgs.rs#L2484-L2495

morehouse (approved):
LGTM

PR #354, opened 2025-12-08
refactor: simplify auto_build.sh with dynamic module discovery

erickcestari:
Replace 17 hardcoded module wrapper functions with a single directory lookup function. Modules are now derived from CXXFLAGS automatically using pattern matching.

### DIFF
--- a/auto_build.sh
+++ b/auto_build.sh
@@ -5,169 +5,86 @@ if [ -z "$CXXFLAGS" ]; then
     exit 1
 fi
 
-# Runs make clean to clean everything before starting
 echo "Cleaning previous builds..."
 make clean || { echo "Error: 'make clean' failed"; exit 1; }
 
-# Function to execute a command in a module's directory
-execute_in_module() {
-    local module_dir=$1
+# Execute a command in a directory
+execute_in_dir() {
+    local dir=$1
     local command=$2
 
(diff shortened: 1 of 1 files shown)


brunoerg:
Needs rebase.

brunoerg (approved):
ACK f9436158b06655216c81e3df63961ad6c137aaaf

I tested in practice with a bunch of different modules and worked fine, haven't taken a deep look at the code but lgtm in general.

erickcestari:
> ACK [f943615](https://github.com/bitcoinfuzz/bitcoinfuzz/commit/f9436158b06655216c81e3df63961ad6c137aaaf)
> 
> I tested in practice with a bunch of different modules and worked fine, haven't taken a deep look at the code but lgtm in general.

Rebased.  Should we start using `auto_build.sh` on `workflow.yml`? I think that would simplify our current CI process and test whether the `auto_build.sh` script is working properly. However, I believe this should be addressed in a follow-up PR.

brunoerg:
> Rebased. Should we start using auto_build.sh on workflow.yml? I think that would simplify our current CI process and test whether the auto_build.sh script is working properly. However, I believe this should be addressed in a follow-up PR.

Yes, we should start using it on workflow.yml.

PR #358, opened 2025-12-12
Remove MacOS support

### LINKED ISSUE #356: Should we remove support for macOS?
This is an issue to discuss further the possibility of removing bitcoinfuzz macOS support.

Some arguments in favor of removing macOS support:

1. Developers doing serious fuzzing typically use Linux VMs or containers regardless of their host OS
2. Fuzzing tools and sanitizers (ASan, MSan, UBSan) have better support and more consistent behavior on Linux
3. Coverage-guided fuzzers like AFL++ and libFuzzer are primarily developed and optimized for Linux
4. Limited maintainer time is better spent on improving fuzzing effectiveness rather than cross-platform compatibility
5. Continuous fuzzing infrastructure (OSS-Fuzz, clusterfuzz) runs on Linux
6. Docker provides a straightforward path for macOS users who want to run the fuzzer
7. CI/CD pipeline complexity increases with multiple platform targets

erickcestari:
Closes #356

### DIFF
--- a/.github/actions/setup-test-env/action.yml
+++ b/.github/actions/setup-test-env/action.yml
@@ -1,5 +1,5 @@
 name: Setup Test Environment
-description: Sets up the environment for a test job, including repo checkout, cache restore, and macOS-specific setup.
+description: Sets up the environment for a test job, including repo checkout, cache restore.
 inputs:
   os:
     description: Operating system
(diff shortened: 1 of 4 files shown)


brunoerg (changes requested):
Concept ACK. 

I think we don't need to be too agressive for now. Perhaps we could keep the support for macOS on the modules' Makefile, to not break who is already using it (at least to reproduce a bug, for example) and then we could simply remove macOS from our CI (which is a sign that we don't care about any specific macOS issue).

brunoerg (approved):
ACK 1fd925c1a87b620f9b42de0cc655f857ecc246cc

PR #360, opened 2025-12-15
remove the script_asm target

brunoerg:
This target has not been useful so far because implementations differ on how they represent the ASM of scripts (e.g. `OP_CHECKSEQUENCEVERIFY ` and `OP_CSV ` - they are the same but makes the target to crash). Also, we could return the ASM of the script in the other script targets, if it makes sense.


brunoerg:
Rebased.

PR #373, opened 2025-12-30
Add Schnorr signing fuzz target to secp256k1 module

### LINKED ISSUE #365: Add target for `schnorr` signatures (secp256k1)

ZenMachina16:
Fixes https://github.com/bitcoinfuzz/bitcoinfuzz/issues/365.

It adds a new fuzzing target for Schnorr signatures (BIP340). Although the
secp256k1 library already supports Schnorr signing and is built with the required
module enabled, the functionality was not previously exposed through the fuzzing
harness, so those code paths were not being exercised. This change introduces a
new sign_schnorr target, wires it through the driver and base module, and
implements the signing logic in the secp256k1 module using
secp256k1_schnorrsig_sign. As a result, the fuzzer can now feed arbitrary inputs
into the Schnorr signing implementation, improving coverage for
Taproot-related cryptographic code.

The target was tested locally using just run sign_schnorr. The fuzzer starts
successfully, reports new coverage (NEW_FUNC entries), and runs through tens of
thousands of iterations without crashes.
Screenshot from 2025-12-30 12-46-16

### DIFF
--- a/modules/btcd/btcd_wrapper/wrapper.go
+++ b/modules/btcd/btcd_wrapper/wrapper.go
@@ -19,8 +19,12 @@ import (
 	"strings"
 	"unsafe"
 
+	"encoding/hex"
+
 	"github.com/btcsuite/btcd/addrmgr"
 	"github.com/btcsuite/btcd/blockchain"
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/btcutil/hdkeychain"
 	"github.com/btcsuite/btcd/btcutil/psbt"
@@ -296,4 +300,26 @@ func BTCDBip32MasterKeygen(data C.ByteArray) *C.char {
 	}
 	return C.CString(masterKey.String())
 }
+
+//export BTCDSignSchnorr
+func BTCDSignSchnorr(privKey C.ByteArray, hash C.ByteArray, aux C.ByteArray) *C.char {
+	privKeyBytes := C.GoBytes(unsafe.Pointer(privKey.data), privKey.length)
+	hashBytes := C.GoBytes(unsafe.Pointer(hash.data), hash.length)
+	auxBytes := C.GoBytes(unsafe.Pointer(aux.data), aux.length)
+
+	// Ensure we have exactly 32 bytes for the aux data
+	var auxArray [32]byte
+	copy(auxArray[:], auxBytes)
+
+	priv, _ := btcec.PrivKeyFromBytes(privKeyBytes)
+
+	// Use CustomNonce to force determinism with the provided aux data
+	sig, err := schnorr.Sign(priv, hashBytes, schnorr.CustomNonce(auxArray))
+	if err != nil {
+		return C.CString("")
+	}
+
+	return C.CString(hex.EncodeToString(sig.Serialize()))
+}
+
 func main() {}
--- a/modules/secp256k1/module.cpp
+++ b/modules/secp256k1/module.cpp
@@ -3,6 +3,8 @@
 extern "C" {
 #include <secp256k1.h>
 #include <secp256k1_ecdh.h>
+#include <secp256k1_extrakeys.h>
+#include <secp256k1_schnorrsig.h>
 }
 
 #undef template
@@ -149,6 +151,33 @@ secp256k1_ecdh_generate(std::span<const uint8_t> buffer,
   return hex_encode(shared_secret.data(), SECP256K1_SHARED_SECRET_LEN);
 }
 
+std::optional<std::string>
+secp256k1_sign_schnorr(std::span<const uint8_t> buffer,
+                       std::span<const uint8_t> hash,
+                       std::span<const uint8_t> aux) {
+  int ret = 0;
+  secp256k1_keypair keypair;
+  std::vector<uint8_t> signature(64);
+  const uint8_t *privkey = buffer.data();
+
+  if (!secp256k1_ec_seckey_verify(secp256k1_ctx, privkey)) {
+    return std::nullopt;
+  }
+
+  ret = secp256k1_keypair_create(secp256k1_ctx, &keypair, privkey);
+  if (!ret)
+    return "";
+
+  ret = secp256k1_schnorrsig_sign32(secp256k1_ctx, signature.data(),
+                                    hash.data(), &keypair, aux.data());
+
+  if (!ret) {
+    return "";
+  }
+
+  return hex_encode(signature.data(), 64);
+}
+
 namespace bitcoinfuzz {
 namespace module {
 Secp256k1::Secp256k1(void) : BaseModule("Secp256k1") { init(nullptr, nullptr); }
@@ -181,6 +210,13 @@ std::optional<std::string>
 Secp256k1::ecdh(std::span<const uint8_t> buffer,
                 std::span<const uint8_t> pubkey) const {
   return secp256k1_ecdh_generate(buffer, pubkey);
+}
+
+std::optional<std::string>
+Secp256k1::sign_schnorr(std::span<const uint8_t> buffer,
+                        std::span<const uint8_t> hash,
+                        std::span<const uint8_t> aux) const {
+  return secp256k1_sign_schnorr(buffer, hash, aux);
 }
 
 } // namespace module
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -190,6 +190,9 @@ jobs:
             target: "ecdh"
             cxxflags: "-DDECRED_SECP256K1 -DSECP256K1"
             asan_options: "detect_leaks=0:detect_container_overflow=0"
+          - corpus: "sign_schnorr"
+            target: "sign_schnorr"
+            cxxflags: "-DBTCD -DSECP256K1"
 
     steps:
       - name: Checkout repo
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -260,3 +260,15 @@ services:
     env_file: .env
     volumes:
       - ./docker:/app/data
+
+  sign_schnorr:
+    build:
+      context: .
+      tags:
+        - bitcoinfuzz:sign_schnorr
+      args:
+        CXXFLAGS: "-DSECP256K1 -DBTCD"
+        FUZZ: sign_schnorr
+    env_file: .env
+    volumes:
+      - ./docker:/app/data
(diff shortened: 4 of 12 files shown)


brunoerg:
Concept ACK and thanks for working on it. However, I think you should check if decred's secp256k1 also supports it and add it here, otherwise, we will only have one module for this target which is not super productive for us.

ZenMachina16:
Thank you for the feedback. I will check whether Decred’s secp256k1 supports Schnorr signing and if it can be integrated into the same target. If feasible, I will update this PR to include it, otherwise, I’ll report back with what I find.

ZenMachina16:
I implemented the `sign_schnorr` target for the Decred module as well (using dcrec/secp256k1/v4/schnorr).

While both implementations produce valid signatures, the outputs differ byte-wise for the same inputs. This appears to be due to differences in nonce generation between Decred’s implementation and libsecp256k1.

Since differential fuzzing relies on deterministic, identical outputs, I’ve temporarily disabled the Decred implementation for this target (returning std::nullopt) to keep the fuzzer stable with secp256k1 alone.

Do you know if there is a way to force deterministic nonce derivation in Decred that matches libsecp256k1, or would it be preferable to keep this target single-module for now?

erickcestari:
Thanks for the PR! It's already looking good. I'm currently in a trip, but I'll try to review as soon as possible!

> Do you know if there is a way to force deterministic nonce derivation in Decred that matches libsecp256k1, or would it be preferable to keep this target single-module for now?

That's interesting. I think the best approach is to see how they implement each nonce generation. Is it deterministic? Are they both following an RFC? I'm just providing some guiding questions that I'll try to answer when I review to see what the best way to proceed is.

erickcestari:
The Decred Schnorr signatures are not related to Bitcoin, so it's fine not to implement them as a target, since the objective here is to test targets related to Bitcoin. Schnorr signatures should follow BIP-340.

Decred: https://github.com/decred/dcrd/blob/eae45642441e86de60ddfc0ca57b425656b15ed7/dcrec/secp256k1/schnorr/signature.go#L245

libsecp256k1: https://github.com/bitcoin-core/secp256k1/blob/694342fdb71722b31dd4b6bc7241b992fdf6f492/include/secp256k1_schnorrsig.h#L95

So how can LND have taproot channels? 

I searched and found that the implementation of BIP-340 (Schnorr signature scheme) is in their btcd library. We could add the sign_schnorr target to the btcd module. That would be great.

https://github.com/btcsuite/btcd/blob/7520523a9a54aff213f5455b048a55f818118a84/btcec/schnorr/signature.go#L248

erickcestari (changes requested):
```
make[1]: Entering directory '/home/runner/work/bitcoinfuzz/bitcoinfuzz/modules/secp256k1'
clang-format -Werror --fail-on-incomplete-format -n ./module.cpp ./module.h
./module.cpp:167:12: error: code should be clang-formatted [-Wclang-format-violations]
  if (!ret) return "";

```

The code isn't formatted correctly. Run `make format-all` at the root of the project to ensure that all modules are formatted correctly. That should fix the CI error.

ZenMachina16:
> Run `make format-all` at the root of the project to ensure that all modules are formatted correctly. That should fix the CI error.

I’ve removed the Decred-related changes to keep the PR focused on BIP-340 Schnorr and formatted the modified files with clang-format. Happy to make any further changes if needed.

brunoerg:
>  
> 
> > Run `make format-all` at the root of the project to ensure that all modules are formatted correctly. That should fix the CI error.
> 
> 
> 
> I’ve removed the Decred-related changes to keep the PR focused on BIP-340 Schnorr and formatted the modified files with clang-format. Happy to make any further changes if needed.

Cool, thanks, but please take a look at btcd and consider adding it as well. Otherwise, with just one module we will not have "differential" fuzzing.

ZenMachina16:
I’ve added the sign_schnorr target to the btcd module and wired it up for differential fuzzing against libsecp256k1.

During testing, I initially saw immediate mismatches due to differences in auxiliary nonce randomness used during signing (both implementations were producing valid but different signatures). To make differential fuzzing meaningful, I updated both sides to use an explicit deterministic 32-byte zeroed auxiliary buffer (secp256k1_schnorrsig_sign on the C++ side and schnorr.CustomNonce in btcd).

With this change, both implementations now produce identical signatures and the fuzzer runs cleanly locally. Please let me know if you’d like any adjustments.

justfile
```diff
@@ -1,10 +1,8 @@
 #!/usr/bin/env -S just --justfile
 
-# required_bins := require("docker") + require("docker-compose") + require("jq")
-required_bins := "" # Placeholder for older just versions (v1.21.0 does not support require)
-
-
+required_bins := require("docker") + require("docker-compose") + require("jq")
```
brunoerg, line 3: Can you explain this change?

brunoerg (changes requested):
The commit 24650697b3173a6f496937cd3b2b6934c95b69ee (which only format the code) must be squashed.

brunoerg (commented):
Code lgtm, I will leave it running overnight and take a final look tomorrow. Also, you should change "This PR addresses https://github.com/bitcoinfuzz/bitcoinfuzz/issues/365." to "Fixes #365" (or related).

ZenMachina16:
>>Can you explain this change?

This change isn’t introducing new behavior... it restores the original required_bins line that was already present upstream.

I had temporarily replaced it with an empty placeholder locally while debugging build issues on my machine, and that workaround was accidentally committed earlier. I’ve now reverted it back to the original require(...) version to stay in sync with the upstream justfile...

Sorry for the confusion ... no functional change intended here.

justfile
ZenMachina16: This change isn’t introducing new behavior... it restores the original required_bins line that was already present upstream.

I had temporarily replaced it with an empty placeholder locally while debugging build issues on my machine, and that workaround was accidentally committed earlier. I’ve now reverted it back to the original require(...) version to stay in sync with the upstream justfile...

Sorry for the confusion ... no functional change intended here.

modules/secp256k1/module.cpp
```diff
@@ -138,17 +140,42 @@ secp256k1_ecdh_generate(std::span<const uint8_t> buffer,
...
-  ret = secp256k1_ecdh(secp256k1_ctx, shared_secret.data(), &pubkey, privkey,
-                       nullptr, nullptr);
+
+  ret = secp256k1_keypair_create(secp256k1_ctx, &keypair, privkey);
+  if (!ret)
+    return "";
+
+  std::vector<uint8_t> aux(32, 0x00);
+  ret = secp256k1_schnorrsig_sign(secp256k1_ctx, signature.data(), hash.data(),
+                                  &keypair, aux.data());
```
erickcestari: Instead of providing a fixed zeroed nonce we could let the fuzzer generate it and send as a function parameter. to `sign_schnorr` functions. That would allow us to cover more edge cases.

modules/btcd/btcd_wrapper/wrapper.go
```diff
@@ -296,4 +299,26 @@ func BTCDBip32MasterKeygen(data C.ByteArray) *C.char {
...
+	var auxArray [32]byte
+	copy(auxArray[:], auxBytes)
+
+	priv, _ := btcec.PrivKeyFromBytes(privKeyBytes)
+
+	// Use CustomNonce to force determinism with the provided aux data
+	sig, err := schnorr.Sign(priv, hashBytes, schnorr.CustomNonce(auxArray))
+	if err != nil {
+		return nil
+	}
```
erickcestari, line 320: We should return an empty string to identify cases where one implementation can sign and the other cannot. We only return a null pointer if we want to skip comparing with other modules for some reason.

```suggestion
	// Use CustomNonce to force determinism with the provided aux data
	sig, err := schnorr.Sign(priv, hashBytes, schnorr.CustomNonce(auxArray))
	if err != nil {
		return C.CString("")
	}
```

erickcestari (changes requested):
Thanks! It's looking good! Just a few changes and it should be ready.

modules/secp256k1/module.cpp
ZenMachina16: Sure..I implemented sign_schnorr which now accepts a 32-byte auxiliary buffer from the fuzzer input, and I've updated both the  secp256k1 and btcd modules to use it for deterministic nonce generation.

erickcestari (changes requested):
LGTM! Just missing a rebase.

Could you please rebase the commits?

modules/secp256k1/module.cpp
```diff
@@ -138,17 +140,42 @@ secp256k1_ecdh_generate(std::span<const uint8_t> buffer,
...
     return std::nullopt;
   }
-  ret = secp256k1_ecdh(secp256k1_ctx, shared_secret.data(), &pubkey, privkey,
-                       nullptr, nullptr);
+
+  ret = secp256k1_keypair_create(secp256k1_ctx, &keypair, privkey);
+  if (!ret)
+    return "";
+
+  ret = secp256k1_schnorrsig_sign(secp256k1_ctx, signature.data(), hash.data(),
```
erickcestari:
```suggestion
  ret = secp256k1_schnorrsig_sign32(secp256k1_ctx, signature.data(), hash.data(),
```

`secp256k1_schnorrsig_sign` is deprecated

erickcestari (changes requested):

erickcestari (changes requested):
The CI format check failed because the secp256k1 module is missing to be formatted correctly. Run `make format`. And it should be good to go!

```
bitcoinfuzz/modules/secp256k1 on  pr-373 [?] via C v15.2.1-gcc 
❯ make check-format
clang-format -Werror --fail-on-incomplete-format -n ./module.cpp ./module.h
./module.cpp:171:69: error: code should be clang-formatted [-Wclang-format-violations]
  ret = secp256k1_schnorrsig_sign32(secp256k1_ctx, signature.data(), hash.data(),
                                                                    ^
./module.cpp:171:82: error: code should be clang-formatted [-Wclang-format-violations]
  ret = secp256k1_schnorrsig_sign32(secp256k1_ctx, signature.data(), hash.data(),
                                                                                 ^
make: *** [Makefile:36: check-format] Error 1
```

erickcestari (changes requested):
I almost forgot, but the target needs to be added to the CI. Then it should be ready to merge! Sorry for reiterating it so much instead of just requesting these changes once.
```
- corpus: "sign_schnorr"
            target: "sign_schnorr"
            cxxflags: "-DBTCD -DSECP256K1"
```

brunoerg:
@erickcestari Do you want to take a final look here?

erickcestari (approved):
Thanks! Nice work!

ACK 7cab633eb2a4deff3f06c6177780479b19965810

PR #377, opened 2026-01-06
Add differential fuzzing for nbitcoin-secp256k1 module

NishantBansal2003:
ref: #331 

Add differential fuzzing between the nbitcoin-secp256k1 (C#) implementation and the existing secp256k1 modules. This implementation covers all existing functions including private to public key operations, signing (both compact and DER formats), signature verification and ECDH operations.

The only point to discuss is that there were some memory leak issues which were addressed by forcing garbage collection. However, this makes the fuzzing slightly slower. The choice is therefore between accepting slower fuzzing with the current approach or suppressing the leaks on the libFuzzer side using -detect_leaks=0. I would appreciate any suggestions on improving this further or guidance on which of these options would be preferred.

Also, the file modules/nbitcoinsecp256k1/Makefile is mostly taken from the existing modules/nbitcoin/Makefile. Although I understand that macOS support has been removed the nbitcoinsecp256k1/Makefile still contains some macOS specific targets. Please let me know whether this is acceptable or if those targets should be removed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ ifneq ($(findstring -DBTCD,$(BASE_CXXFLAGS) $(CXXFLAGS)),)
 	MODULES += modules/btcd/module.a
 endif
 
-ifneq ($(findstring -DNBITCOIN,$(BASE_CXXFLAGS) $(CXXFLAGS)),)
+ifneq ($(filter -DNBITCOIN,$(BASE_CXXFLAGS) $(CXXFLAGS)),)
 	MODULES += modules/nbitcoin/module.a
 endif
 
@@ -89,6 +89,10 @@ ifneq ($(findstring -DSECP256K1,$(BASE_CXXFLAGS) $(CXXFLAGS)),)
 	MODULES += modules/secp256k1/module.a
 endif
 
+ifneq ($(filter -DNBITCOIN_SECP256K1,$(BASE_CXXFLAGS) $(CXXFLAGS)),)
+	MODULES += modules/nbitcoinsecp256k1/module.a
+endif
+
 ifeq ($(UNAME_S), Darwin)
 	LDFLAGS = -framework CoreFoundation -Wl,-ld_classic
 endif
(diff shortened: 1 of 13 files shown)


brunoerg:
Concept ACK, thanks for working on it. I will review it soon.

Makefile
```diff
@@ -32,7 +32,7 @@ ifneq ($(findstring -DBTCD,$(BASE_CXXFLAGS) $(CXXFLAGS)),)
 	MODULES += modules/btcd/module.a
 endif
 
-ifneq ($(findstring -DNBITCOIN,$(BASE_CXXFLAGS) $(CXXFLAGS)),)
+ifneq ($(filter -DNBITCOIN,$(BASE_CXXFLAGS) $(CXXFLAGS)),)
```
brunoerg, line 35: Can you please explain this change?

NishantBansal2003: findstring is like grep -- it can match even with `-DNBITCOIN_SECP256K1`. The filter is for exact matching between spaces

modules/nbitcoinsecp256k1/NBitcoinSecp256k1/NBitcoinSecp256k1Bridge.cs
```diff
@@ -0,0 +1,179 @@
...
+    private const int BufferLength = 32;
+
+    // Disposable struct that forces GC cleanup when disposed.
+    private readonly struct GCScope : IDisposable
+    {
+        public void Dispose()
+        {
+            // We force a GC here to clean up managed wrappers holding native resources.
+            // Normally the GC's finalizer thread takes care of this, but during fuzzing
+            // the GC may not run often enough and the leaks start to accumulate.
```
brunoerg, line 16: I think these aren't true leaks. They are delayed cleanup of finalizable objects. Once the GC runs, they get cleaned up. But yes, during fuzzing this delay can exhaust native resources (like native memory).

brunoerg:
Code lgtm. Also, I tested it and worked fine. I will do just one more look before merging, thank you.

modules/secp256k1/module.cpp
```diff
@@ -138,10 +138,12 @@ secp256k1_ecdh_generate(std::span<const uint8_t> buffer,
     return std::nullopt;
   }
 
-  ret = ret = secp256k1_ec_pubkey_parse(secp256k1_ctx, &pubkey,
-                                        pubkey_buf.data(), pubkey_buf.size());
-  ret = ret && secp256k1_ecdh(secp256k1_ctx, shared_secret.data(), &pubkey,
-                              privkey, nullptr, nullptr);
+  if (!secp256k1_ec_pubkey_parse(secp256k1_ctx, &pubkey, pubkey_buf.data(),
```
brunoerg, line 141: Can you please explain this change in secp256k1 module?

NishantBansal2003: This is done to sync with decred’s ecdh target, where we return nil for invalid public key bytes, and now also to align with NBitcoin’s ecdh target as well.

https://github.com/bitcoinfuzz/bitcoinfuzz/blob/acba761857640b3ac106bcb3df67f1158f1af441/modules/decredsecp256k1/decredsecp256k1_wrapper/wrapper.go#L143-L146

brunoerg: Ok, looks good.

brunoerg:
After merging this PR, we cannot forget to update our corpora's CI.

brunoerg (approved):
ACK 5878f973f623b5905004fc7d40171aaa85c5fc66

PR #380, opened 2026-01-12
fix: add NBitcoinSecp256k1.CppBridge.so to the artifacts cache

erickcestari:


brunoerg (approved):
ACK c6394e1ae546f2df68efbb5c9009d92b8db6d3fd

needs to update the corpora's CI as well?

erickcestari:
> ACK [c6394e1](https://github.com/bitcoinfuzz/bitcoinfuzz/commit/c6394e1ae546f2df68efbb5c9009d92b8db6d3fd)
> 
> needs to update the corpora's CI as well?

No, it shouldn't be needed since there we cache the whole bitcoinfuzz folder.

PR #387, opened 2026-01-16
addrv2: update string return format

### LINKED ISSUE #385: addrv2: Update the return format of the string
It might be useful to change the return format of the string for `addrv2` target to an array containing the addresses, time, service bits and port in string format. For the addresses, we might want to keep them in hexadecimal format, since formatting addresses in the same way using different libraries might be difficult. This would ensure that they are actually receiving the same addresses.

```rust
/// Address received from BIP-0155 addrv2 message
#[derive(Clone, PartialEq, Eq, Hash, Debug)]
pub struct AddrV2Message {
    /// Time that this node was last seen as connected to the network
    pub time: u32,
    /// Service bits
    pub services: ServiceFlags,
    /// Network ID + Network Address
    pub addr: AddrV2,
    /// Network port, 0 if not applicable
    pub port: u16,
}
```

What do you think? @brunoerg

GaneshPatil7517:
Fixes #385

This updates the addrv2 string output to return a structured array-style
format containing address, time, service bits, and port.

Addresses remain in hexadecimal form to ensure consistency across
libraries and platforms.

This change improves parseability while keeping behavior deterministic.

### DIFF
--- a/modules/rustbitcoin/rust_bitcoin_lib/Cargo.lock
+++ b/modules/rustbitcoin/rust_bitcoin_lib/Cargo.lock
@@ -157,6 +157,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "hex"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7f24254aa9a54b5c858eaee2f5bccdb46aaf0e486a595ed5fd8f86ba55232a70"
+
+[[package]]
 name = "hex-conservative"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


erickcestari (changes requested):
Hi @GaneshPatil7517, thanks for contributing to bitcoinfuzz! This PR is missing updates to the other modules that implement the `addrv2` target.

A note on AI-assisted contributions: they work best when you understand the end goal well enough to iterate with the AI toward a complete solution. Without that context, incomplete pull requests end up creating more work for the maintainer to review than if they implemented it themselves.

I'd be happy to review a revised version that covers all the affected modules.

erickcestari:
Since there's been no response and the PR remains incomplete (missing updates to the other modules implementing the `addrv2` target), I'm going to close this for now.

If you'd like to revisit this contribution in the future with a complete implementation covering all affected modules, feel free to open a new PR. Happy to review it then!

PR #390, opened 2026-01-16
add MODULES environment variable for selective module loading

erickcestari:
Introduce a ModuleRegistry singleton that allows users to specify which
modules to load at runtime via the MODULES environment variable.

Changes:
- Add include/bitcoinfuzz/module_registry.h with ModuleRegistry class:
  - Singleton pattern with lazy initialization
  - initialize(): parses MODULES env var, validates against compiled modules
  - isEnabled(): checks if a module should be loaded at runtime
  - getCompiledModules(): returns set of all compiled module names
  - parseModuleList(): parses comma-separated module names with whitespace trimming

- Update main.cpp:
  - Call InitializeRegistryOnce() at start of each fuzzer invocation
  - Abort with error if requesting modules that weren't compiled
  - Gate each LoadModule() call with registry.isEnabled() check

- Update driver.cpp and helpers/jvmloader.cpp:
  - Add runtime isEnabled() checks for module-specific workarounds and paths
  - Allows module-specific code to be skipped when that module is disabled

Usage: MODULES="BITCOIN_CORE,RUST_BITCOIN" FUZZ=bip32_master_keygen ./bitcoinfuzz
If MODULES is unset or empty, all compiled modules are loaded (existing behavior).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,6 +7,7 @@
 **/obj/
 **/bin/
 **/target/
+**/build/
 **.dylib
 **.zip
 **.jar
--- a/Dockerfile
+++ b/Dockerfile
@@ -128,6 +128,9 @@ COPY --from=builder \
     /usr/lib/llvm-18/bin/llvm-symbolizer /usr/bin/llvm-symbolizer
 ENV ASAN_SYMBOLIZER_PATH /usr/bin/llvm-symbolizer
 
+# Comma-separated list of modules to load (empty = all modules)
+ENV MODULES=""
+
 # interpreted language sources
 COPY --from=builder --parents \
     --exclude=modules/bitcoin/secp256k1/tools \
--- a/README.md
+++ b/README.md
@@ -157,6 +157,7 @@ If you prefer, you can still build the modules manually. Below are the steps for
     cd modules/rustbitcoinkernel
     make cargo && make
     export CXXFLAGS="$CXXFLAGS -DRUSTBITCOINKERNEL"
+    ```
 
 - ### py-bitcoinkernel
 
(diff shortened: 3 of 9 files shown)


brunoerg:
Interesting, I will surely test it next Monday.

erickcestari:
It's worth mentioning that @johnnyasantoss came up with this idea. Thanks!

brunoerg:
I'm better thinking about the approach, looks good but I have a question. We might have some checks based on the modules we defined with CXXFLAGS, e.g.:

```sh
#ifdef BTCD
  if (std::ranges::find(input_data, 0xAC) != input_data.end() ||
      std::ranges::find(input_data, 0xAE) != input_data.end() ||
      std::ranges::find(input_data, 0xAF) != input_data.end())
    return;
#endif
```

How would it work on this approach?

erickcestari:
> I'm better thinking about the approach, looks good but I have a question. We might have some checks based on the modules we defined with CXXFLAGS, e.g.:
> 
> ```shell
> #ifdef BTCD
>   if (std::ranges::find(input_data, 0xAC) != input_data.end() ||
>       std::ranges::find(input_data, 0xAE) != input_data.end() ||
>       std::ranges::find(input_data, 0xAF) != input_data.end())
>     return;
> #endif
> ```
> 
> How would it work on this approach?

True, I had completely forgotten about those. I changed it to use the `singleton` pattern, which is probably best here. This allows us to initialize and check for it using a `set` at runtime throughout the code.

johnnyasantoss:
It's maybe worth it to add `ENV MODULES=""` to dockerfile to signal that it can be tweaked.

I've tried building locally to test it with docker but `deserialize_offer` and `deserialize_invoice` failed to build with a maven error. Not sure if it's related.

erickcestari:
> It's maybe worth it to add `ENV MODULES=""` to dockerfile to signal that it can be tweaked.

Added!

> I've tried building locally to test it with docker but `deserialize_offer` and `deserialize_invoice` failed to build with a maven error. Not sure if it's related.

I don't think it's related. It worked on my personal machine. There is probably something missing from the .dockerignore file.
 
I had an issue with bitcoin core univalue build, that's why I am adding it in this PR.

brunoerg (approved):
ACK 4c5d911173229e85dcd3c2653f17d19ccd72665d 

I just tested it for the script_eval target, compiling btcd and bitcoin core but defining which would run at runtime. It worked well. Also, I like that the current behavior still remains. So if anyone just wants to use CXXFLAGS and the build as is, it's ok.

PR #394, opened 2026-01-20
feat: print python errors

erickcestari:
Sometimes, when I forget to use a virtual environment (venv) when trying to run a Python module, it looks like the fuzzing is working, but it's not.. Adding `PyErr_Print();` makes it clear that it's not working if the environment is not configured correctly.

### DIFF
--- a/modules/embit/module.cpp
+++ b/modules/embit/module.cpp
@@ -72,6 +72,7 @@ cleanup:
   Py_XDECREF(input_obj);
 
   if (PyErr_Occurred()) {
+    PyErr_Print();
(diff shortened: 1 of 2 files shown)


brunoerg:
> Sometimes, when I forget to use a virtual environment (venv) when trying to run a Python module, it looks like the fuzzing is working, but it's not.. 

It's because when you forget to use venv, you don't have the required libraries installed, right? Then it can't effectively run but you're not aware.

erickcestari:
> It's because when you forget to use venv, you don't have the required libraries installed, right? Then it can't effectively run but you're not aware.

Exactly

PR #395, opened 2026-01-20
build: improve Makefile dependency tracking and simplify build commands

erickcestari:
- Use proper file targets instead of phony cargo targets for incremental builds
- Simplify build commands: `make` now handles cargo builds automatically
- Add missing .PHONY declarations across all module Makefiles
- Fix tinyminiscript to use TINY_LIB_EXT variable consistently

### DIFF
--- a/.github/workflows/commit-by-commit-build.yml
+++ b/.github/workflows/commit-by-commit-build.yml
@@ -135,14 +135,14 @@ jobs:
         timeout-minutes: 5
         run: |
           rustup default nightly
-          cd modules/rustbitcoin && make cargo && make
+          cd modules/rustbitcoin && make
 
       - name: Build - rust-bitcoinkernel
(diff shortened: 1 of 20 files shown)


.github/workflows/commit-by-commit-build.yml
```diff
@@ -186,7 +186,7 @@ jobs:
         timeout-minutes: 5
         run: |
           rustup default nightly
-          cd modules/ldk && make cargo && make
```
brunoerg, line 189: It means that the `make` command will always call `cargo build`?

erickcestari: Yes!

brunoerg (approved):
code review ACK d039aa0241640da29b2279b31b6040f36425045b

PR #396, opened 2026-01-21
feat: add BitcoinJ to extkey deserialization target

kuliq23:
Due to how [deserialization](https://github.com/bitcoinj/bitcoinj/blob/ac593b4e162a126794b7b07123840ffb3d61f257/core/src/main/java/org/bitcoinj/crypto/DeterministicKey.java#L642) works in BitcoinJ, (parent key is needed to get a valid childnumber) `child=00000000;` is always present in the final output. This makes fuzzing fail.

see the "zeroed" child e.g. here:

```
xprv9s2SHxhPMGLurjhydgXZ33GXgYeay6pqsmiAdzuYV8LPqXwxtMogrAxqoe5xyjkjmjss9DADrpx7Qd8PPzcxry6GPv4uks6qustFDtx3o1F

Module: NBitcoin
Result: depth=00;fp=01000000;child=00000004;chaincode=0000000000000000000000000000000000000000000000000000000000000000;key=0000000000000000000000004a00000000000000000000000000000000000000
Module: BitcoinJ
Result: depth=00;fp=01000000;child=00000000;chaincode=0000000000000000000000000000000000000000000000000000000000000000;key=0000000000000000000000004a00000000000000000000000000000000000000
```

or here

```
tprv8ZgxMBitZRetiHuCVmbHHrYKs1sXtrcZ2ZyL2MfemneaEck838icS6QjaSjG9Y35Nrg47FXHtAwPjHv6ogvG3d9cB3XWmwCxSYaZ3xhQq5Q

Module: NBitcoin
Result: depth=00;fp=00000000;child=f7ffffff;chaincode=ffffffff00008200000000000000000000000000000000000000000000000000;key=000000000000000000000000000000000000000a000000000000000000000000
Module: BitcoinJ
Result: depth=00;fp=00000000;child=00000000;chaincode=ffffffff00008200000000000000000000000000000000000000000000000000;key=000000000000000000000000000000000000000a000000000000000000000000
```

Due to the nature of fuzzing itself, it is impossible to provide a valid parent key for every generated extkey. 

---

**I see 3 options we could go about this:**

- Leave out BitcoinJ from this target all together

- Make an exception in output comparison when BitcoinJ is used

- Theoretically, a parent child duo could be generated using [deriveChildKey](https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoinj/crypto/HDKeyDerivation.java#L151) but I find it out of the scope of this target, or?

Let me know your thoughts.

### DIFF
--- a/modules/bitcoinj/app/src/main/kotlin/wrapper/Wrapper.kt
+++ b/modules/bitcoinj/app/src/main/kotlin/wrapper/Wrapper.kt
@@ -1,7 +1,9 @@
 package wrapper
 
 import org.bitcoinj.base.BitcoinNetwork
+import org.bitcoinj.crypto.DeterministicKey
 import org.bitcoinj.crypto.HDKeyDerivation
+import java.math.BigInteger
 
 object Wrapper {
     @JvmStatic external fun init(): Unit
@@ -25,6 +27,103 @@ object Wrapper {
             } else {
                 ""
             }
+        }
+    }
+
+    @JvmStatic
+    fun deserializeExtendedKey(bytes: ByteArray): String {
+        if (bytes.isEmpty()) {
+            return "INVALID"
+        }
+        val base58 = String(bytes, Charsets.UTF_8)
+
+        // skip keys that are bigint 0 or bigint 1 in private key bytes - BitcoinJ does not allow them
+        val decoded =
+            try {
+                org.bitcoinj.base.Base58.decodeChecked(base58)
+            } catch (e: Exception) {
+                return "INVALID"
+            }
+        if (decoded.size == 78) {
+            val privBytes = decoded.sliceArray(46..77)
+            val priv = BigInteger(1, privBytes) // convert to positive BigInteger
+            if (priv == BigInteger.ZERO || priv == BigInteger.ONE) {
+                return "skip error"
+            }
+        }
+        val network =
+            when (base58[0]) {
+                'x' -> BitcoinNetwork.MAINNET
+                't' -> BitcoinNetwork.TESTNET
+                else -> return "INVALID"
+            }
+        val key =
+            try {
+                DeterministicKey.deserializeB58(base58, network)
+            } catch (e: Exception) {
+                return "INVALID"
+            }
+        val depth = key.depth
+        val fingerprint = key.parentFingerprint
+        val child =
+            if (key.depth == 0 && key.childNumber.i() == 0) {
+                extractChildNumberFromBase58(base58) ?: 0
+            } else {
+                key.childNumber.i()
+            }
+        val chainCodeHex =
+            key.chainCode.joinToString(
+                separator = "",
+            ) {
+                "%02x".format(it)
+            }
+        val keyHex =
+            if (key.hasPrivKey()) {
+                key.privKeyBytes.joinToString(
+                    separator = "",
+                ) {
+                    "%02x".format(it)
+                }
+            } else {
+                // Force lazy public key validation (decode + curve check).
+                try {
+                    key.getPubKeyPoint()
+                } catch (e: Exception) {
+                    return "INVALID"
+                }
+                key.pubKey.joinToString(
+                    separator = "",
+                ) {
+                    "%02x".format(it)
+                }
+            }
+        val result =
+            String.format(
+                "depth=%02x;fp=%02x%02x%02x%02x;child=%08x;chaincode=%s;key=%s",
+                depth,
+                (fingerprint ushr 24) and 0xff,
+                (fingerprint ushr 16) and 0xff,
+                (fingerprint ushr 8) and 0xff,
+                fingerprint and 0xff,
+                child,
+                chainCodeHex,
+                keyHex,
+            )
+        return result
+    }
+
+    private fun extractChildNumberFromBase58(base58: String): Int? {
+        return try {
+            val decoded = org.bitcoinj.base.Base58.decodeChecked(base58)
+            if (decoded.size != 78) return null
+
+            // child number is bytes 9..12
+            ((decoded[9].toInt() and 0xff) shl 24) or
+                ((decoded[10].toInt() and 0xff) shl 16) or
+                ((decoded[11].toInt() and 0xff) shl 8) or
+                (decoded[12].toInt() and 0xff)
+        } catch (_: Exception) {
+            null
         }
     }
 }
--- a/modules/bitcoinj/module.cpp
+++ b/modules/bitcoinj/module.cpp
@@ -14,6 +14,7 @@ namespace fs = std::filesystem;
 static JavaVM *jvm = nullptr;
 static jclass bitcoinJWrapperClass = nullptr;
 static jmethodID createMasterKeyMethod = nullptr;
+static jmethodID deserializeExtendedKeyMethod = nullptr;
 
 static bool init_jvm() {
   if (jvm != nullptr) {
@@ -37,6 +38,14 @@ static bool init_jvm() {
   if (!createMasterKeyMethod) {
     return false;
   }
+
+  deserializeExtendedKeyMethod = env->GetStaticMethodID(
+      bitcoinJWrapperClass, "deserializeExtendedKey", "([B)Ljava/lang/String;");
+
+  if (!deserializeExtendedKeyMethod) {
+    return false;
+  }
+
   return true;
 }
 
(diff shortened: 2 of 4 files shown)


modules/bitcoinj/app/src/main/kotlin/wrapper/Wrapper.kt
```diff
@@ -27,4 +28,63 @@ object Wrapper {
...
+            }
+        val key =
+            try {
+                DeterministicKey.deserializeB58(base58, network)
+            } catch (e: Exception) {
+                return "INVALID"
+            }
+        val depth = key.depth
+        val fingerprint = key.parentFingerprint
+        val child = key.childNumber.i()
```
erickcestari: To avoid the issue of `Bitcoinj` zeroing the child number because it is a root key, we can decode the Base58 and find the child number ourselves. This should work fine.

```suggestion
val child =
            if (key.depth == 0 && key.childNumber.i() == 0) {
                extractChildNumberFromBase58(base58) ?: 0
            } else {
                key.childNumber.i()
            }
```

```kt
private fun extractChildNumberFromBase58(base58: String): Int? {
        return try {
            val decoded = org.bitcoinj.base.Base58.decodeChecked(base58)
            if (decoded.size != 78) return null

            // child number is bytes 9..12
            ((decoded[9].toInt() and 0xff) shl 24) or
                ((decoded[10].toInt() and 0xff) shl 16) or
                ((decoded[11].toInt() and 0xff) shl 8) or
                (decoded[12].toInt() and 0xff)
        } catch (_: Exception) {
            null
        }
    }

```

modules/bitcoinj/app/src/main/kotlin/wrapper/Wrapper.kt
```diff
@@ -27,4 +28,63 @@ object Wrapper {
...
+                    separator = "",
+                ) {
+                    "%02x".format(it)
+                }
+            } else {
+                key.pubKey.joinToString(
+                    separator = "",
+                ) {
+                    "%02x".format(it)
+                }
```
erickcestari, line 98: We should trigger public key validation here, since BitcoinJ uses lazy public key verification.

```suggestion
                // Force lazy public key validation (decode + curve check).
                try {
                    key.getPubKeyPoint()
                } catch (e: Exception) {
                    return "INVALID"
                }
                key.pubKey.joinToString(
                    separator = "",
                ) {
                    "%02x".format(it)
                }
```

modules/bitcoinj/module.cpp
```diff
@@ -79,12 +89,61 @@ bitcoinj_bip32_master_keygen(std::span<const uint8_t> buffer) {
...
+  }
+
+  JNIEnv *env = nullptr;
+  jint status = jvm->GetEnv((void **)&env, JNI_VERSION_1_8);
+  if (status != JNI_OK || !env) {
+    return "";
+  }
+
+  jstring jInput =
+      env->NewStringUTF(reinterpret_cast<const char *>(buffer.data()));
```
erickcestari: `NewStringUTF` requires a NUL-terminated Modified UTF-8 string. The incoming buffer is length-delimited, so passing `buffer.data()` directly may cause an overread on the native side. Creating a bounded `std::string `first guarantees proper termination and avoids undefined behavior.

```suggestion
std::string input(reinterpret_cast<const char *>(buffer.data()),
                    buffer.size());

  jstring jInput = env->NewStringUTF(input.c_str());
```

modules/bitcoinj/app/src/main/kotlin/wrapper/Wrapper.kt
```diff
@@ -27,4 +28,63 @@ object Wrapper {
...
+                (fingerprint ushr 16) and 0xff,
+                (fingerprint ushr 8) and 0xff,
+                fingerprint and 0xff,
+                child,
+                chainCodeHex,
+                keyHex,
+            )
+
+        // System.err.println("deserializeExtendedKey result = $result")
+        // System.err.flush()
```
erickcestari: These comments can be removed.

erickcestari (changes requested):
Thanks!

BitcoinJ will only zero the child number when it receives a root key (depth = 0). However, in those cases, we can extract the child number from the base58 string ourselves, so we can avoid the fuzzer crashing due to false positives.

erickcestari:
BitcoinJ appears to have the same issue as NBitcoin: it misses test vector 5, and fails on some of them.

https://github.com/bitcoinj/bitcoinj/blob/master/core/src/test/java/org/bitcoinj/crypto/BIP32Test.java#L37

```
test vector: xprv9s21ZrT9nLigZg99v6iAHBtnXvm4XZPw6FuFn9ccQsFC29kkYQaURG1TwP58GadyYjt6xDPeNTQzDgcEwGdvTtcJtotWEAcBEPByoix8aWR

Module: Rustbitcoin
Result: INVALID (Xpriv error: non-zero parent fingerprint in master key; Xpub error: non-zero parent fingerprint in master key)
Module: BitcoinJ
Result: depth=00;fp=0000000a;child=00000000;chaincode=0000000000000000000000000000000000000000000000000000000000000000;key=0000000000000001000000000000000000000000000000000000000000000000
```
Nice work!

Do you want to open a PR/issue reporting it?

kuliq23:
I can create an Issue now, but I won't get to the PC until Wednesday. You can create the PR and link the Issue if you want.

As for the comments, I will incorporate them next week.
Thank you for understanding

erickcestari:
@kuliq23 Let me know when this PR is ready for review. Thanks!

kuliq23:
> @kuliq23 Let me know when this PR is ready for review. Thanks!

Ready, thanks for all the comments.
As you may noticed, I tend to apply the suggested changes manually instead of using GitHub’s commit suggestion feature, as it can complicate history during rebases. Hope that’s okay with you.

erickcestari:
> > @kuliq23 Let me know when this PR is ready for review. Thanks!
> 
> Ready, thanks for all the comments. As you may noticed, I tend to apply the suggested changes manually instead of using GitHub’s commit suggestion feature, as it can complicate history during rebases. Hope that’s okay with you.

It's fine to apply those manually. I actually prefer this way.

modules/bitcoinj/module.cpp
```diff
@@ -79,12 +89,58 @@ bitcoinj_bip32_master_keygen(std::span<const uint8_t> buffer) {
...
+                    buffer.size());
+  jstring jInput = env->NewStringUTF(input.c_str());
+  if (!jInput) {
+    return "";
+  }
+
+  jstring jResult = static_cast<jstring>(env->CallStaticObjectMethod(
+      bitcoinJWrapperClass, deserializeExtendedKeyMethod, jInput));
+
+  env->DeleteLocalRef(jInput);
```
erickcestari:
```suggestion
jbyteArray jBytes = env->NewByteArray(static_cast<jsize>(buffer.size()));
  if (!jBytes) {
    return "";
  }

  env->SetByteArrayRegion(jBytes, 0, static_cast<jsize>(buffer.size()),
                          reinterpret_cast<const jbyte *>(buffer.data()));

  jstring jResult = static_cast<jstring>(env->CallStaticObjectMethod(
      bitcoinJWrapperClass, deserializeExtendedKeyMethod, jBytes));
  env->DeleteLocalRef(jBytes);
```

modules/bitcoinj/module.cpp
```diff
@@ -37,6 +38,15 @@ static bool init_jvm() {
   if (!createMasterKeyMethod) {
     return false;
   }
+
+  deserializeExtendedKeyMethod =
+      env->GetStaticMethodID(bitcoinJWrapperClass, "deserializeExtendedKey",
+                             "(Ljava/lang/String;)Ljava/lang/String;");
```
erickcestari:
```suggestion
deserializeExtendedKeyMethod = env->GetStaticMethodID(
      bitcoinJWrapperClass, "deserializeExtendedKey", "([B)Ljava/lang/String;");
```

modules/bitcoinj/app/src/main/kotlin/wrapper/Wrapper.kt
```diff
@@ -27,4 +28,85 @@ object Wrapper {
             }
         }
     }
+
+    @JvmStatic
+    fun deserializeExtendedKey(base58: String): String {
+        if (base58.isEmpty()) {
+            return "INVALID"
+        }
```
erickcestari:
```suggestion
fun deserializeExtendedKey(bytes: ByteArray): String {
        if (bytes.isEmpty()) {
            return "INVALID"
        }

        val base58 = String(bytes, Charsets.UTF_8)
```

erickcestari (changes requested):
It's working, but I found one edge case in the way it sends input from C++ code to Java. It removes everything from the NUL byte onwards, resulting in BitcoinJ receiving only part of the input, which will cause the fuzzer to crash later when BitcoinJ parses it and accept as valid. 

It's better to send a byte array to the JVM, which interprets the bytes as a string. This makes the decodeBase58 in BitcoinJ receive the same string/bytes as other implementations.

example of input that would make the fuzzer crash:

xxd -p crash-a4c9dea519d7fafdbcd5f6b0b067d8084c354cc4 
```
7470756245677762737772566637677234724457396e39624e5554377544
37536f6a366a385879745a3771537742416932613934444a796868655672
425558386550325a6d644c775875517350694c41656334344b76534d7232
516f676955477a6b355772717858466a687a356f51000000000000000000
000000000000000000000000000000000000000000000000000000000000
000000000000000000000000000000000000000000000000000000000000
00000000000000000000000000000000
```

kuliq23:
> It's working, but I found one edge case in the way it sends input from C++ code to Java. It removes everything from the NUL byte onwards, resulting in BitcoinJ receiving only part of the input, which ...

Very good catch, thank you!!

EDIT: It seems like I updated submodules by accident, will fix this...

---

After these changes, just out of curiosity, I tried comparing only NBitcoin and BitcoinJ - BitcoinJ does NOT like when keybytes are `0x0...001` ?

try e.g. these keys:
`xprv9s21ZrQH143K24Mfq5zL5MhWK9hUhhGbd45hLXo2Pq2oqzMMo63oStZzF93Y5wvzdUayhgkkFoicQZcP3y52uPPxFnfoLZB21TeqtDeZVxb`
`xprv9s21ZrQH143K24Mfq5zL5MhWK9hUhhGbd45hLXo37H4B5uiRXehnYNiCxYXwAB7aUy3GPVrfB7ovuWy8CqZrPjymmga6bALUaxJGJg66Xsn`
`tprv8ZgxMCzDvi63V6VZQqkXooMk2DcE1nVJtGeE6RNtaFrRRmS1q2EcgiYUsJZ9WJUK5KDn3sKK12uj1sjfLA2CvATLSWW2TD66RsXk7wzdhko`

erickcestari (changes requested):
You have accidentally downgraded the `eclair` and 'core `lightning` submodules.

erickcestari:
> After these changes, just out of curiosity, I tried comparing only NBitcoin and BitcoinJ - BitcoinJ does NOT like when keybytes are `0x0...001` ?
> 
> try e.g. these keys: `xprv9s21ZrQH143K24Mfq5zL5MhWK9hUhhGbd45hLXo2Pq2oqzMMo63oStZzF93Y5wvzdUayhgkkFoicQZcP3y52uPPxFnfoLZB21TeqtDeZVxb` `xprv9s21ZrQH143K24Mfq5zL5MhWK9hUhhGbd45hLXo37H4B5uiRXehnYNiCxYXwAB7aUy3GPVrfB7ovuWy8CqZrPjymmga6bALUaxJGJg66Xsn` `tprv8ZgxMCzDvi63V6VZQqkXooMk2DcE1nVJtGeE6RNtaFrRRmS1q2EcgiYUsJZ9WJUK5KDn3sKK12uj1sjfLA2CvATLSWW2TD66RsXk7wzdhko`

Interesting. They have a comment that explains the rationale. So, we should probably skip those cases.

https://github.com/bitcoinj/bitcoinj/blob/b2d8af7aad0baff7a4dc2fb9bf67648805327ce7/core/src/main/java/org/bitcoinj/crypto/ECKey.java#L227-L238

kuliq23:
> Interesting. They have a comment that explains the rationale. So, we should probably skip those cases.
> 
>https://github.com/bitcoinj/bitcoinj/blob/b2d8af7aad0baff7a4dc2fb9bf67648805327ce7/core/src/main/java/org/bitcoinj/crypto/ECKey.java#L227-L238

---

Fixed the submodules and added a skip for those cases. I suppose checks fail due to previously discussed issue with not-defined CXX in the CI runner.

erickcestari:
> Fixed the submodules and added a skip for those cases. I suppose checks fail due to previously discussed issue with not-defined CXX in the CI runner.

Yep, I will hardcode the CXX flags to prevent this error from happening again.

erickcestari (approved):
Thanks! Nice work!

ACK 6ce3b41ed4c93f138e7df5a6908ddbb237d43d1d

PR #400, opened 2026-01-23
refactor(addrv2): change return format to JSON array with detailed address info

### LINKED ISSUE #385: addrv2: Update the return format of the string

erickcestari:
Change the `addrv2` target return format from aggregate counts (clearnet=X tor=Y cjdns=Z i2p=W) to a JSON array containing detailed information for each address entry.

Each entry now includes:
- addr: raw address bytes in hexadecimal format
- type: address type (ipv4, ipv6, tor, i2p, cjdns)
- time: timestamp as string
- services: service bits as string
- port: network port as string

Using hex format for addresses ensures consistent comparison across different implementations (Bitcoin Core, rust-bitcoin, btcd) without relying on potentially different address formatting logic.

Changes by module:
- `bitcoin/module.cpp`: Return JSON array with full address details, handle IPv4-mapped IPv6 format from GetAddrBytes()
- `btcd/wrapper.go`: Use reflection to extract raw address bytes from unexported btcd types, skip torv2 and IPv4-mapped IPv6 addresses
- `rustbitcoin/lib.rs`: Add addrv2_to_hex() and addrv2_type() helpers, implement is_routable_ipv4/ipv6() checks matching Core behavior
- `driver.cpp`: Add BTCD ifdef to skip i2p/cjdns mismatches (btcd skips these address types)

This enables byte-level verification that all implementations parse and route the same addresses from addrv2 messages.

closes #385

### DIFF
--- a/driver.cpp
+++ b/driver.cpp
@@ -274,6 +274,22 @@ void Driver::AddrV2Target(std::span<const uint8_t> buffer) const {
       continue;
     if (last_response.has_value()) {
       if (*res != *last_response) {
+#ifdef BTCD
+        // Skip mismatches involving btcd when i2p or cjdns addresses are
+        // present (btcd always skips i2p and cjdns)
+        bool involves_btcd =
+            (module.first == "Btcd" || last_module_name == "Btcd");
+        bool has_i2p_or_cjdns =
+            (res->find("i2p") != std::string::npos ||
+             res->find("cjdns") != std::string::npos ||
+             last_response->find("i2p") != std::string::npos ||
+             last_response->find("cjdns") != std::string::npos);
+        if (involves_btcd && has_i2p_or_cjdns) {
+          last_response = *res;
+          last_module_name = module.first;
+          continue;
+        }
+#endif
         std::cout << "Addrv2 parse failed" << std::endl;
         std::cout << "Module: " << module.first << std::endl;
         std::cout << "Result: " << *res << std::endl;
--- a/modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
+++ b/modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
@@ -18,6 +18,8 @@ use p2p::message::{AddrV2Payload, RawNetworkMessage};
 use p2p::Magic;
 use std::ffi::CStr;
 use std::ffi::CString;
+use std::net::Ipv4Addr;
+use std::net::Ipv6Addr;
 use std::os::raw::c_char;
 use std::slice;
 use std::str::{FromStr, Utf8Error};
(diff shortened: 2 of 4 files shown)


driver.cpp
```diff
@@ -274,6 +274,19 @@ void Driver::AddrV2Target(std::span<const uint8_t> buffer) const {
       continue;
     if (last_response.has_value()) {
       if (*res != *last_response) {
+#ifdef BTCD
+        // Skip mismatches involving btcd when i2p or cjdns addresses are
+        // present (btcd always skip i2p and cjdns)
```
brunoerg:
```suggestion
        // present (btcd always skips i2p and cjdns)
```

erickcestari: Fixed. And I've also added more validation for the ipv6 to BTCD and rust-bitcoin

modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
```diff
@@ -199,41 +201,144 @@ pub unsafe extern "C" fn rust_bitcoin_psbt_parse(data: *const u8, len: usize) ->
...
-                } else if matches!(a.addr, AddrV2::TorV3(_)) {
-                    tor += 1;
-                } else if matches!(a.addr, AddrV2::Cjdns(_)) {
-                    cjdns += 1;
-                } else if matches!(a.addr, AddrV2::I2p(_)) {
-                    i2p += 1;
+                if matches!(a.addr, AddrV2::Unknown(..)) {
+                    continue;
                 }
+                // Match the same behavior as isRoutable() function from core.
```
brunoerg, line 319: I'm not sure about this approach. You're implementing two functions to check if the addresses are routable that are not used in practice by any implementation/library, just to match Core's behavior. I mean, it seems that other implementations should use it (e.g. Floresta) or even this function should be part of rust-bitcoin.

brunoerg (approved):
code review ACK 7619d400bbd8078b8dc51d077c900f0b54f9cfcf

modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
erickcestari: Yes, that was necessary to prevent it from crashing every time. I like the idea of having both floresta and rust-bitcoin modules for this target, because Floresta may provide more abstraction and have more checks than this minimal version of Rust-Bitcoin. Therefore, testing Floresta would not be the same as testing rust-bitcoin with minimal checks to be compatible with the core.

erickcestari: @apoelstra. Curious if you'd consider adding `is_routable()` helpers to rust-bitcoin's p2p module? We needed it here for Core compatibility and figured other implementations (Floresta) probably hit the same thing. Happy to open an issue or draft a PR if there's interest.

PR #402, opened 2026-01-24
Add differential fuzzing for k256 module

NishantBansal2003:
ref: #331 

Adds differential fuzzing between the [K256](https://crates.io/crates/k256) implementation and the existing secp256k1 modules.

I used `-DRUST_K256` instead of `-DK256` because `-DK256` defines the `K256` macro, which breaks the code in places where `K256` is used as a variable or class name.


modules/rustk256/k256_lib/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+    let pub_key = match VerifyingKey::from_sec1_bytes(pubkey_slice) {
+        Ok(p) => p,
+        Err(_) => {
+            return ptr::null_mut();
+        }
+    };
+
+    // Here, we are not using the k256 provided diffie_hellman function because, although it currently performs
+    // the same internal operations we do, it only returns the x-coordinate of the generated secret.
+    // However, we need both the x and y coordinates to generate the compressed shared public key bytes.
```
erickcestari: nit:

```suggestion
    // However, we need the x coordinate and y parity to generate the compressed shared public key bytes.
```

modules/rustk256/k256_lib/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+
+[package]
+name = "k256_lib"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+hex = "0.4.3"
+k256 = { version = "0.13.4" }
+sha2 = "0.10.9"
```
erickcestari: To avoid finding bugs that have already been fixed, we should use a recent version of the library.

```suggestion
k256 = { git = "https://github.com/RustCrypto/elliptic-curves.git", rev = "a4af4fa03caaa135bac7ca04fd8b0d00e1c87592", package = "k256", features=["sha2"]}
```

erickcestari (changes requested):
It's looking good! Switching to a commit-based version might break the current code, though. Once that's done, it should be ready to go!

erickcestari (approved):
Thanks!!

We could also have `sign_schnorr` target for the k256 module. But this can be done in a follow-up PR. 

ACK 3b315d0689383ac07fb08d925f703349b81d6109

brunoerg:
Nice! Please, add this module to our corpora's CI.

PR #406, opened 2026-01-26
Prototype for dynamic loading bitcoinkernel module

KY-U:
# Draft PR — dynamic bitcoinkernel loading

## Summary
- Functional prototype to load multiple `libbitcoinkernel` versions dynamically in `bitcoinfuzz`.
- Goal: confirm the flow before automating download/build and load of the libs (currently manual).
- Library and wrapper headers are checked in only for the prototype; once automated, I plan to link them at build/run time to keep them out of history.

## Done so far
- Added the test function `library_tag` to prove two different libraries are loaded dynamically.
- `bitcoinfuzz` uses `FUZZ=library_tag` to exercise the load path for each `libbitcoinkernel*.so`.
- Implemented the `kernel_block` target (C interface) to compare behavior with existing bindings.

## How to validate locally
1) In `bitcoinkernel.cpp`, add before compiling:
   ```cpp
   const char* btck_library_tag(void) {
       return "KERNEL A";
   }
   ```
2) Add the declaration to `bitcoinkernel.h` inside the existing `extern "C"` block:
   ```cpp
   BITCOINKERNEL_API const char* btck_library_tag(void);
   ```
   You can copy the updated header from this PR if easier.
3) Build Bitcoin Core with shared libs: `-DBUILD_SHARED_LIBS=ON`.
4) Produce two variants and rename to `libbitcoinkernelA.so` and `libbitcoinkernelB.so`.
5) Place both in `modules/dynamicbitcoinkernel/libs/`.
6) Run `FUZZ=library_tag ./bitcoinfuzz -runs=1` (or any `-runs=n`); the first test should fail and you’ll see distinct libraries results.

## Additional validation
- You can also build `bitcoinfuzz` with another kernel binding such as `pybitcoinkernel` module and run the `kernel_block` target (already implemented for the C interface) to compare behavior.

## Proposed next steps
- Automate build/download of the `libbitcoinkernel` variants and the library loading.
- Remove headers from the repo and fetch/link them only at build/run time.
- Add automated tests for the new module.

## Feedback:
Every feedback is welcome! And if possible, I would like to clarify these specific topics:
- Confirm if this flow makes sense before continuing current implementation.
- Validate thoughts on keeping headers only at build/run time (avoiding commits).

### DIFF
--- a/modules/dynamicbitcoinkernel/module.cpp
+++ b/modules/dynamicbitcoinkernel/module.cpp
@@ -1,0 +1,295 @@
+#include "module.h"
+#include "bitcoinkernel_wrapper/bitcoinkernel_wrapper.h"
+#include <dlfcn.h>
+#include <iomanip>
+#include <memory>
+#include <span>
+#include <sstream>
+#include <string.h>
+#include <unordered_map>
+
+struct btck_Block;
+struct btck_BlockHash;
+struct btck_Transaction;
+struct btck_Txid;
+
+using btck_WriteBytes = int (*)(const void *, size_t, void *);
+
+using btck_block_create_signature = btck_Block *(*)(const void *, size_t);
+using btck_block_copy_signature = btck_Block *(*)(const btck_Block *);
+using btck_block_count_transactions_signature = size_t (*)(const btck_Block *);
+using btck_block_get_transaction_at_signature =
+    const btck_Transaction *(*)(const btck_Block *, size_t);
+using btck_block_get_hash_signature = btck_BlockHash *(*)(const btck_Block *);
+using btck_block_to_bytes_signature = int (*)(const btck_Block *,
+                                              btck_WriteBytes, void *);
+using btck_block_hash_to_bytes_signature = void (*)(const btck_BlockHash *,
+                                                    unsigned char[32]);
+using btck_block_hash_destroy_signature = void (*)(btck_BlockHash *);
+using btck_block_hash_copy_signature =
+    btck_BlockHash *(*)(const btck_BlockHash *);
+using btck_block_hash_equals_signature = int (*)(const btck_BlockHash *,
+                                                 const btck_BlockHash *);
+using btck_block_destroy_signature = void (*)(btck_Block *);
+using btck_transaction_get_txid_signature =
+    const btck_Txid *(*)(const btck_Transaction *);
+using btck_txid_copy_signature = btck_Txid *(*)(const btck_Txid *);
+using btck_txid_equals_signature = int (*)(const btck_Txid *,
+                                           const btck_Txid *);
+using btck_txid_to_bytes_signature = void (*)(const btck_Txid *,
+                                              unsigned char[32]);
+using btck_txid_destroy_signature = void (*)(btck_Txid *);
+using btck_library_tag_signature = const char *(*)(void);
+
+struct KernelApiTable {
+  void *handle{};
+  btck_block_create_signature block_create{};
+  btck_block_copy_signature block_copy{};
+  btck_block_count_transactions_signature block_count_transactions{};
+  btck_block_get_transaction_at_signature block_get_transaction_at{};
+  btck_block_get_hash_signature block_get_hash{};
+  btck_block_to_bytes_signature block_to_bytes{};
+  btck_block_hash_copy_signature block_hash_copy{};
+  btck_block_hash_equals_signature block_hash_equals{};
+  btck_block_hash_to_bytes_signature block_hash_to_bytes{};
+  btck_block_hash_destroy_signature block_hash_destroy{};
+  btck_block_destroy_signature block_destroy{};
+  btck_transaction_get_txid_signature transaction_get_txid{};
+  btck_txid_copy_signature txid_copy{};
+  btck_txid_equals_signature txid_equals{};
+  btck_txid_to_bytes_signature txid_to_bytes{};
+  btck_txid_destroy_signature txid_destroy{};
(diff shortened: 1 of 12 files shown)


brunoerg:
Needs rebase.

KY-U:
> Needs rebase.

Done.

modules/dynamicbitcoinkernel/module.cpp
```diff
@@ -0,0 +1,296 @@
...
+      api_tables_cache;
+  std::shared_ptr<const KernelApiTable> api_table;
+
+  auto cached_api_table = api_tables_cache.find(library_path);
+  if (cached_api_table != api_tables_cache.end()) {
+    api_table = cached_api_table->second;
+  } else {
+    dlerror();
+    void *handle =
+        dlmopen(LM_ID_NEWLM, library_path.c_str(), RTLD_NOW | RTLD_LOCAL);
```
erickcestari: I had some issues when trying to run bitcoinfuzz with the dlmopen. It couldn't find asan symbols and either coverage symbols that are needed for fuzzing. It only works when compiling bitcoinkernel without any sanitizers. We can use the `dlopen` to fix that.

```suggestion
    void *handle =
        dlopen(library_path.c_str(), RTLD_NOW | RTLD_GLOBAL);
```

erickcestari (changes requested):

modules/dynamicbitcoinkernel/module.cpp
brunoerg: I think the issue is that `dlmopen` loads the library into a completely separate namespace and the symbols are basically isolated. With `dlopen`, it loads in the default namespace so the symbols are available globally.

KY-U: > I had some issues when trying to run bitcoinfuzz with the dlmopen. It couldn't find asan symbols and either coverage symbols that are needed for fuzzing. It only works when compiling bitcoinkernel without any sanitizers. We can use the `dlopen` to fix that.
After implementing the suggested change:

When I tried running the fuzzer with the shared libs compiled without sanitizer, as I've been doing until now, I got:
```
==27194==ERROR: AddressSanitizer: heap-use-after-free on address 0x510000000340 at pc 0x562c199c5c58 bp 0x7fff18bd1540 sp 0x7fff18bd1538
```
Trying with libs compiled with `-DSANITIZERS=fuzzer`, I got:
```
==9987==ERROR: AddressSanitizer: SEGV on unknown address 0x0000000000e6 (pc 0x555a4f9c7bd7 bp 0x000000000000 sp 0x7ffd5aee80a0 T0)
```
And when trying with the address sanitizer flag, `-DSANITIZERS=address` and `-DSANITIZERS=fuzzer,address`. I got an odr-violation error on both cases.
```
==4813==ERROR: AddressSanitizer: odr-violation (0x7f70e78441c0):
```
Was the suggested line change really the only modification applied to module.cpp? That change alone doesn't seem to work, at least in my local tests.

erickcestari: True, I've tried just one of the libraries with coverage and address instrumentation. When running with both libraries instrumented it will fail because of the odr-violation.

Loading two copies of the same library (libbitcoinkernel) with ASAN causes an ODR violation because both export identical symbols (like typeinfo for CCoinsViewBacked). ASAN detects the duplicate and aborts.

Changing `RTLD_GLOBAL` -> `RTLD_LOCAL` fixed the issue for me. 

```c++
 void *handle =
        dlopen(library_path.c_str(), RTLD_NOW | RTLD_LOCAL);
```

erickcestari (commented):
Thanks for working on this!

I liked the approach here. It's really working which is great. However, I would like to see (if it's possible) a POC using static library instead of dynamic ones. I can list some advantages that static libraries would give to us.

1. The final executable binary would not be dependent of those dynamic libraries (e.g we would be able to move the executable without need to move the .so together)
2. It would actually simplify a lot the code in the `module.cpp` since we would not need to load a dynamic library at runtime and load the function symbols. (It might get more complex when building since probably we will need to rename symbols but I think is worth).
3. It would also help us to fix other modules where static libraries have conflict symbols.

KY-U:
> Thanks for working on this!
> 
> I liked the approach here. It's really working which is great. However, I would like to see (if it's possible) a POC using static library instead of dynamic ones. I can list some advantages that static libraries would give to us.
> 
> 1. The final executable binary would not be dependent of those dynamic libraries (e.g we would be able to move the executable without need to move the .so together)
> 2. It would actually simplify a lot the code in the `module.cpp` since we would not need to load a dynamic library at runtime and load the function symbols. (It might get more complex when building since probably we will need to rename symbols but I think is worth).
> 3. It would also help us to fix other modules where static libraries have conflict symbols.

Thanks for the review and feedback! 
You are right. While interesting, the dynamic symbol resolution implementation ended up not trivial to understand and mantain. I'll start implementing the static library approach. It was a fun exploratory work, but It seems the symbol renaming really is the best solution.

erickcestari:
> Thanks for the review and feedback! You are right. While interesting, the dynamic symbol resolution implementation ended up not trivial to understand and mantain. I'll start implementing the static library approach. It was a fun exploratory work, but It seems the symbol renaming really is the best solution.

I think you can keep this as a draft in case it turns out that the static approach isn't better than this one.  Then we can decide which PR to merge and close the other.

brunoerg:
Can we close this in favor of #429?

KY-U:
> Can we close this in favor of #429?

Absolutely.

PR #412, opened 2026-01-28
build: only link libsodium when clightning module is enabled

erickcestari:


brunoerg (approved):
ACK 29f837627d912bc95df9332340e715b8276bd90d

Issue #398, opened 2026-01-21
Implement sign_schnorr target for NBitcoin module


erickcestari:
Implement `sign_schnorr` target for the `NBitcoin` module to enable differential fuzzing of Schnorr signature generation.

Other modules (`secp256k1`, `btcd`) already implement this target.

bala-bhargav:
@erickcestari 
can i work on it ??

erickcestari:
> [@erickcestari](https://github.com/erickcestari) can i work on it ??

Yes, please feel free to work on it. If anyone else is working on it too, which I don't think they are, that's great, one of you will end up reviewing the other's work when the first PR is opened. It won't be "lost".

bala-bhargav:
@erickcestari 
can you review the PR #414

bala-bhargav:
@erickcestari 
Hi, I noticed the issue title changed from "NBitcoin" to "NBitcoinSecp256k1".
I've already implemented [sign_schnorr] for the NBitcoin module. 
Do you want it implemented in NBitcoinSecp256k1 only?

erickcestari:
> [@erickcestari](https://github.com/erickcestari) Hi, I noticed the issue title changed from "NBitcoin" to "NBitcoinSecp256k1". I've already implemented [sign_schnorr] for the NBitcoin module. Do you want it implemented in NBitcoinSecp256k1 only?

I don't have a strong opinion either way. Would it make sense to include it in both? Since `NBitcoin` abstracts `NbitcoinSecp256k1`, which could misbehave.

I have actually been thinking of removing `NBitcoinSecp256k1`. We could then work only with the `NBitcoin` abstractions, which would cover more code. If some target isn't possible to do with the `NBitcoin` library then we should fallback to `NBitcoinSecp256k1`.

@brunoerg what do you think?

brunoerg:
> I have actually been thinking of removing NBitcoinSecp256k1. We could then work only with the NBitcoin abstractions, which would cover more code. If some target isn't possible to do with the NBitcoin library then we should fallback to NBitcoinSecp256k1.

It sounds to good to work only with the NBitcoin because I think this is the only way this library is really used in practice.

PR #397, opened 2026-01-21
Add verify_schnorr differential fuzzing target 

erickcestari:
Implements Schnorr signature verification differential fuzzing between `secp256k1`, `btcd`, and `NBitcoinSecp256k1`. Changes:

- Add `schnorr_verify` target to base module interface
- Implement `secp256k1` verification using `secp256k1_schnorrsig_verify`
- Implement `btcd` verification using `schnorr.ParseSignature` and Verify
- Implement `NBitcoinSecp256k1` verification using `ECXOnlyPubKey.SigVerifyBIP340`
- Add custom mutator for generating valid Schnorr signatures (50% probability)
- Add CI workflow and Docker Compose configuration

PR #414, opened 2026-01-30
Add sign_schnorr target for NBitcoin module

bala-bhargav:
Implements the [sign_schnorr] fuzzing target for the NBitcoin module, enabling differential fuzzing of Schnorr signature generation.


bala-bhargav:
@erickcestari
this is the PR for #398

brunoerg (changes requested):
You need to NBitcoin into this target on CI.

erickcestari (changes requested):
Sorry about that. This change should be in the `NbitcoinSecp256k1` module rather than being in the `Nbitcoin` module directly. I'll update the issue to reflect this, too.

It should be straightforward to convert this csharp code to use the NbitcoinSecp256k1 API instead. You can find some insights into how the `Nbitcoin` and `NbitcoinSecp256k1` APIs differ here. 

https://github.com/bitcoinfuzz/bitcoinfuzz/pull/397/changes#diff-b398b22ffe6d55c4a8c6d086324933a0e172fb420f66538f95f8af29bc44491d

bala-bhargav:
@erickcestari @brunoerg 
can you review the changes

brunoerg:
Needs rebase.

bala-bhargav:
Rebased onto latest v2 and resolved merge conflicts

erickcestari (changes requested):
Sorry for the back-and-forth. 😅

Indeed, it's better to implement it in the `NBitcoin` module, as you did before.

bala-bhargav:
@erickcestari 
Updated per feedback:
- Kept sign_schnorr in NBitcoin module (reverted NBitcoinSecp256k1 move)
- Added `-DNBITCOIN` to sign_schnorr CI target

modules/nbitcoin/NBitcoin/NBitcoinBridge.cs
```diff
@@ -224,6 +225,37 @@ private static bool TryParseXpub(string input, Network network, out string resul
...
+            Marshal.Copy(hashPtr, hashBytes, 0, 32);
+
+            byte[] auxBytes = new byte[32];
+            Marshal.Copy(auxPtr, auxBytes, 0, 32);
+
+            Key key = new Key(privkeyBytes);
+            var hash256 = new uint256(hashBytes);
+            var aux256 = new uint256(auxBytes);
+            
+            TaprootSignature sig = key.SignTaprootKeySpend(hash256, merkleRoot: null, aux: aux256, TaprootSigHash.Default);
```
erickcestari: Use the function `SignTaprootScriptSpend` should give us the same behavior as the other implementations. The `SignTaprootKeySpend` makes a tweak in the private key causing to create different signatures.

```suggestion
            TaprootSignature sig = key.SignTaprootScriptSpend(hash256, merkleRoot: null, aux: aux256, TaprootSigHash.Default);
```

modules/nbitcoin/NBitcoin/NBitcoinBridge.cs
```diff
@@ -224,6 +225,37 @@ private static bool TryParseXpub(string input, Network network, out string resul
...
+            
+            byte[] sigBytes = sig.SchnorrSignature.ToBytes();
+            string hexSignature = Hex(sigBytes);
+
+            return Marshal.StringToHGlobalAnsi(hexSignature);
+        }
+        catch
+        {
+            return IntPtr.Zero;
```
erickcestari: We want to catch the case where NBitcoin doesn't generate a signature and the other implementation does. Therefore, we should return an empty string here instead of skipping (return `IntPtr.Zero`).

```suggestion
            return Marshal.StringToCoTaskMemUTF8("");
```

erickcestari (changes requested):
It's looking good! Just a few more changes and it should be ready! Also, the `NBitcoin` at `sign_schnorr` target needs to be added to Docker Compose.

modules/nbitcoin/NBitcoin/NBitcoinBridge.cs
bala-bhargav: Changed to `Marshal.StringToCoTaskMemUTF8("")`

modules/nbitcoin/NBitcoin/NBitcoinBridge.cs
bala-bhargav: Taken care

erickcestari (changes requested):
It looks good to go. The only thing missing is for the code to be formatted correctly.

Run `make format` inside the nbitcoin module.

erickcestari (changes requested):
Also the CI that check if the code is correctly formatted isn't passing. Try to run again the make format inside the nbitcoin module it should fix it. Also you can see in the CI error where exactly it's not formatted.

modules/nbitcoin/NBitcoin/NBitcoinBridge.cs
```diff
@@ -224,6 +225,37 @@ private static bool TryParseXpub(string input, Network network, out string resul
...
+            byte[] sigBytes = sig.SchnorrSignature.ToBytes();
+            string hexSignature = Hex(sigBytes);
+
+            return Marshal.StringToHGlobalAnsi(hexSignature);
+        }
+        catch
+        {
+            return Marshal.StringToCoTaskMemUTF8("");
+        }
+    }
```
erickcestari, line 268: I just noticed that the CI crashed when comparing BTCD and NBitcoin because BTCD doesn't validate the private key during parsing. We should try parsing the key, and if that throws an error, we should return a null pointer.

```suggestion
    [UnmanagedCallersOnly(EntryPoint = "nbitcoin_sign_schnorr")]
    public static IntPtr SignSchnorr(IntPtr privkeyPtr, IntPtr hashPtr, IntPtr auxPtr)
[UnmanagedCallersOnly(EntryPoint = "nbitcoin_sign_schnorr")]
    public static IntPtr SignSchnorr(IntPtr privkeyPtr, IntPtr hashPtr, IntPtr auxPtr)
    {
        byte[] privkeyBytes = new byte[32];
        Marshal.Copy(privkeyPtr, privkeyBytes, 0, 32);
        byte[] hashBytes = new byte[32];
        Marshal.Copy(hashPtr, hashBytes, 0, 32);
        byte[] auxBytes = new byte[32];
        Marshal.Copy(auxPtr, auxBytes, 0, 32);
        var hash256 = new uint256(hashBytes);
        var aux256 = new uint256(auxBytes);

        Key key;
        try
        {
            key = new Key(privkeyBytes);
        }
        catch
        {
            return IntPtr.Zero;
        }
        try
        {
            TaprootSignature sig = key.SignTaprootScriptSpend(hash256, merkleRoot: null, aux: aux256, TaprootSigHash.Default);
            byte[] sigBytes = sig.SchnorrSignature.ToBytes();
            string hexSignature = Hex(sigBytes);

            return Marshal.StringToHGlobalAnsi(hexSignature);
        }
        catch
        {
            return Marshal.StringToCoTaskMemUTF8("");
        }
    }
```

bala-bhargav: @erickcestari 
taken care

bala-bhargav:
@erickcestari Hey! I've pushed the clang-format fix. Could you please re-review when you get a chance and approve the workflow run? Thanks!

erickcestari (changes requested):
It looks like the `NBitcoinBridge.cs` file isn't formatted properly. Always run `make format` inside the modules when you make changes.


from ci:
```
/home/runner/work/bitcoinfuzz/bitcoinfuzz/modules/nbitcoin/NBitcoin/NBitcoinBridge.cs(256,1): error WHITESPACE: Fix whitespace formatting. Replace 25 characters with '\n\s\s\s\s\s\s\s\s\s\s\s\s'. [/home/runner/work/bitcoinfuzz/bitcoinfuzz/modules/nbitcoin/NBitcoin.CppBridge.csproj]
/home/runner/work/bitcoinfuzz/bitcoinfuzz/modules/nbitcoin/NBitcoin/NBitcoinBridge.cs(258,1): error WHITESPACE: Fix whitespace formatting. Replace 25 characters with '\n\s\s\s\s\s\s\s\s\s\s\s\s'. [/home/runner/work/bitcoinfuzz/bitcoinfuzz/modules/nbitcoin/NBitcoin.CppBridge.csproj]
```

brunoerg (changes requested):
Please, there is no need to have commits that only format the code. You should squash this kind of commits.

erickcestari (approved):
Looking good. Thanks for the contribution!

ACK 692048c8e1e316bcaae52e5ce327453d90312f9d

PR #422, opened 2026-02-04
Add `verify_script` target

### LINKED ISSUE #300: Add new target for `VerifyScript`
Bitcoin Core:
```cpp
bool VerifyScript(const CScript& scriptSig, const CScript& scriptPubKey, const CScriptWitness* witness, unsigned int flags, const BaseSignatureChecker& checker, ScriptError* serror)
{
    static const CScriptWitness emptyWitness;
    if (witness == nullptr) {
        witness = &emptyWitness;
    }
    bool hadWitness = false;

    set_error(serror, SCRIPT_ERR_UNKNOWN_ERROR);
```

NBitcoin:
```csharp
public bool VerifyScript(Script scriptSig, WitScript witness, Script scriptPubKey, TransactionChecker checker)
		{
			scriptSig = scriptSig ?? Script.Empty;
			witness = witness ?? WitScript.Empty;
			ExecutionData = new ExecutionData();
			SetError(ScriptError.UnknownError);
```

brunoerg:
Fixes #300

This PR adds a `verify_script` target which is a target to test functions that verify the scriptSig and scriptPubKey in a appropriate context. Also, this PR removes the support for the `script_eval` target from btcd since it's buggy due to some incompatibilities with other implementations as well as most of that is covered by this new target.


brunoerg:
the btcd bump is pointing to the fork from the bitcoinfuzz organization.

modules/bitcoin/module.cpp
```diff
@@ -280,6 +280,22 @@ std::optional<bool> Bitcoin::script_eval(const std::vector<uint8_t> &input_data,
                     nullptr);
 }
 
+std::optional<bool>
+Bitcoin::verify_script(const std::vector<uint8_t> &script_sig,
+                       const std::vector<uint8_t> &script_pubkey,
+                       unsigned int flags) const {
```
erickcestari, line 286: Is `flags` needed? It isn't being used.

erickcestari (commented):
We may also remove the `libbtcd_wrapper.h` and put it in the .gitignore. So whenever we change the `wrapper.go` function definitions it would not be needed to commit it together.

modules/bitcoin/module.cpp
brunoerg: I forgot to mention it. It's currently unused but will be used in a follow-up I'm working on.

brunoerg: Just added a comment.

erickcestari (approved):
ACK 872073829b1c1348e1802d10edce4cdb61ab7294

PR #423, opened 2026-02-05
BIP-32 path parse

kuliq23:
**DRAFT PR**
This is a WIP on the custom mutator for path parsing. 
I am trying to find an equilibrium between tying the path generation to the input bytes (maximalize libFuzzer coverage and utilization) and targeting edgecases (results in repeated paths). 
Added the target for RUST_BITCOIN and BITCOIN_CORE so far.

### DIFF
--- a/driver.cpp
+++ b/driver.cpp
@@ -709,6 +709,18 @@ void Driver::Bip32DeriveFromPathTarget(std::span<const uint8_t> buffer) const {
   }
 }
 
+void Driver::Bip32PathParseTarget(std::span<const uint8_t> buffer) const {
+  std::optional<std::string> last_response{std::nullopt};
+  std::string last_module_name;
+  for (auto &module : modules) {
+    std::optional<std::string> res{module.second->bip32_path_parse(buffer)};
+    if (!res.has_value())
+      continue;
+    VerifyMatchingResponse(last_response, last_module_name, module.first, *res,
+                           "BIP32 path parse failed");
+  }
+}
+
 void Driver::Run(const uint8_t *data, const size_t size,
                  const std::string &target) const {
   std::span<const uint8_t> buffer{data, size};
@@ -772,6 +784,8 @@ void Driver::Run(const uint8_t *data, const size_t size,
     this->StumpModifyAddTarget(buffer);
   } else if (target == "bip32_derive_from_path") {
     this->Bip32DeriveFromPathTarget(buffer);
+  } else if (target == "bip32_path_parse") {
+    this->Bip32PathParseTarget(buffer);
   } else {
     std::cout << "Unknown target: " << target << std::endl;
     assert(false);
--- a/modules/bitcoin/module.cpp
+++ b/modules/bitcoin/module.cpp
@@ -712,5 +712,18 @@ Bitcoin::bip32_derive_from_path(std::span<const uint8_t> buffer) const {
   return EncodeExtKey(key);
 }
 
+std::optional<std::string>
+Bitcoin::bip32_path_parse(std::span<const uint8_t> buffer) const {
+  const std::string path_str(reinterpret_cast<const char *>(buffer.data()),
+                             buffer.size());
+  std::vector<uint32_t> keypath;
+  if (!ParseHDKeypath(path_str, keypath)) {
+    // Failed to parse
+    return "INVALID";
+  }
+
+  return "CORRECT";
+}
+
 } // namespace module
 } // namespace bitcoinfuzz
--- a/modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
+++ b/modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
@@ -518,3 +518,17 @@ pub unsafe extern "C" fn rust_bitcoin_bip32_derive_from_path(
         Err(_) => str_to_c_string("INVALID"),
     }
 }
+
+#[no_mangle]
+pub unsafe extern "C" fn rust_bitcoin_bip32_path_parse(data: *const u8, len: usize) -> *mut c_char {
+    // Safety: Accepts any bytes, does not assume null termination.
+    let bytes = std::slice::from_raw_parts(data, len);
+    let s = match std::str::from_utf8(bytes) {
+        Ok(s) => s,
+        Err(_) => return str_to_c_string("INVALID"), // Not UTF8
+    };
+    match bitcoin::bip32::DerivationPath::from_str(s) {
+        Ok(_) => str_to_c_string("CORRECT"),
+        Err(_) => str_to_c_string("INVALID"),
+    }
+}
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -247,6 +247,11 @@ jobs:
             target: "bip32_derive_from_path"
             cxxflags: "-DRUST_BITCOIN -DBITCOIN_CORE -DNBITCOIN"
             asan_options: "detect_leaks=0:detect_container_overflow=0"
+          - corpus: "bip32_path_parse"
+            target: "bip32_path_parse"
+            cxxflags: "-DRUST_BITCOIN -DBITCOIN_CORE -DNBITCOIN -DBITCOINJ -DLIBWALLY_CORE -DEMBIT -DBITCOINS"
+            asan_options: "detect_leaks=0:detect_container_overflow=0"
+            setup_java: "true"
     steps:
       - name: Checkout repo
         uses: actions/checkout@v4
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -433,3 +433,19 @@ services:
     env_file: .env
     volumes:
       - ./docker:/app/data
+
+  bip32_path_parse:
+    build:
+      context: .
+      tags:
+        - bitcoinfuzz:bip32_path_parse
+      args:
+        CXXFLAGS: "-DRUST_BITCOIN -DBITCOIN_CORE -DNBITCOIN -DBITCOINJ -DLIBWALLY_CORE -DEMBIT -DBITCOINS"
+        FUZZ: bip32_path_parse
+    environment:
+      MODULES: ${MODULES:-}
+      LIBFUZZ_DETECT_LEAKS: 0
+      ASAN_OPTIONS: detect_leaks=0
+    env_file: .env
+    volumes:
+      - ./docker:/app/data
--- a/driver.h
+++ b/driver.h
@@ -64,5 +64,6 @@ public:
   void DecodeOnionTarget(std::span<const uint8_t> buffer) const;
   void StumpModifyAddTarget(std::span<const uint8_t> buffer) const;
   void Bip32DeriveFromPathTarget(std::span<const uint8_t> buffer) const;
+  void Bip32PathParseTarget(std::span<const uint8_t> buffer) const;
 };
 } // namespace bitcoinfuzz
--- a/include/bitcoinfuzz/basemodule.cpp
+++ b/include/bitcoinfuzz/basemodule.cpp
@@ -167,4 +167,9 @@ BaseModule::bip32_derive_from_path(std::span<const uint8_t> /*buffer*/) const {
   return std::nullopt;
 }
 
+std::optional<std::string>
+BaseModule::bip32_path_parse(std::span<const uint8_t> /*buffer*/) const {
+  return std::nullopt;
+}
+
 } // namespace bitcoinfuzz
--- a/include/bitcoinfuzz/basemodule.h
+++ b/include/bitcoinfuzz/basemodule.h
@@ -83,6 +83,9 @@ public:
   virtual std::optional<std::string>
   bip32_derive_from_path(std::span<const uint8_t> buffer) const;
 
+  virtual std::optional<std::string>
+  bip32_path_parse(std::span<const uint8_t> buffer) const;
+
   virtual ~BaseModule() noexcept;
 };
 } // namespace bitcoinfuzz
--- a/modules/bitcoin/module.h
+++ b/modules/bitcoin/module.h
@@ -37,6 +37,8 @@ public:
       std::span<const uint8_t> buffer) const override;
   std::optional<std::string>
   bip32_derive_from_path(std::span<const uint8_t> buffer) const override;
+  std::optional<std::string>
+  bip32_path_parse(std::span<const uint8_t> buffer) const override;
   ~Bitcoin() noexcept override = default;
 };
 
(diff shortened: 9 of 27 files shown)


erickcestari (changes requested):
Thanks for working on it! I don't think we need a custom mutator for BIP-32 paths.

BIP-32 paths are just plain text. LibFuzzer handles them well natively. However, the current approach breaks the coverage feedback loop because GeneratePathFromBytes replaces LibFuzzer's mutations entirely. This prevents LibFuzzer from learning the actual input structure.

By the way, we generally use the custom mutator for cases where we first need to make checksums that would make it almost impossible for the fuzzer to generate a valid one by itself or decode something (like base58), so that we can then mutate the raw bytes. Mutating a base58 string will not help the fuzzer understand the input structure. 

https://github.com/google/fuzzing/blob/master/docs/structure-aware-fuzzing.md

kuliq23:
> Thanks for working on it! I don't think we need a custom mutator for BIP-32 paths.
> 
> BIP-32 paths are just plain text. LibFuzzer handles them well natively. However, the current approach breaks the coverage feedback loop because GeneratePathFromBytes replaces LibFuzzer's mutations entirely. This prevents LibFuzzer from learning the actual input structure.
> 
> By the way, we generally use the custom mutator for cases where we first need to make checksums that would make it almost impossible for the fuzzer to generate a valid one by itself or decode something (like base58), so that we can then mutate the raw bytes. Mutating a base58 string will not help the fuzzer understand the input structure. 
> 
> https://github.com/google/fuzzing/blob/master/docs/structure-aware-fuzzing.md

Thank you so much, makes sense, ill make sure to check this out:)

kuliq23:
LibFuzzer alone explores the BIP-32 path dimension effectively according to code-coverage metrics, consistently reaching derivation depths above 255. This directly exercises the depth-byte overflow case I want to test. I will extend the same approach to other modules before finalizing the PR.

quapka:
> LibFuzzer alone explores the BIP-32 path dimension effectively according to code-coverage metrics, consistently reaching derivation depths above 255. This directly exercises the depth-byte overflow case I want to test. I will extend the same approach to other modules before finalizing the PR.

Great to hear that. Sorry for misguiding you towards the custom mutators approach.

custommutator/mutators/bip32_path.h
```diff
@@ -0,0 +1,164 @@
...
+
+// Normalize a single child segment:
+//  - keep digits
+//  - keep one trailing H (hardened marker)
+//  - remove leading zeros
+// Normalize a single child segment:
+//  - keep digits
+//  - keep one trailing H (hardened marker)
+//  - remove leading zeros
+//  - >>> ADDED: implicitně převést >= 2^31 na hardened
```
quapka: nit: Keep comments in English.

kuliq23:
Due to the inconsistencies in internal representations of BIP-32 paths in different modules, I plan to expand the target to "derive_from_path". This is a less granular but more practical approach, as deriving new keys from a path is the typical use case.

Another obstacle is the sanitization of Path string input. Some modules expect an exact format while others accept more general forms. (e.g. "h","H"," ' ") .... Currently, I am planning a custommutator/sanitizer. That would convert the path strings to a form acceptable for most modules. Different layers of this mutator could be "turned on and off" using e.g. env vars to help fuzz different aspects of the paths. 

Motivation: If we want to test depth-byte overflows, failing on "H" instead of "h" would be inconvenient.

(just a WIP for rust and core pushed)

custommutator/mutators/bip32_path.h
kuliq23: oops, :) thx

modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
```diff
@@ -438,3 +438,48 @@ pub unsafe extern "C" fn rust_bitcoin_bip32_deserialize_extended_key(
         }
     }
 }
+
+#[no_mangle]
+pub unsafe extern "C" fn rust_bitcoin_bip32_path_parse(data: *const c_char) -> *mut c_char {
+    if data.is_null() {
+        //println!("Received null pointer for key path");
```
quapka: Remove irrelevant comments, also below.

driver.cpp
```diff
@@ -942,8 +968,8 @@ void Driver::Run(const uint8_t *data, const size_t size,
     this->SchnorrVerifyTarget(buffer);
   } else if (target == "decode_onion") {
     this->DecodeOnionTarget(buffer);
-  } else if (target == "stump_modify_add") {
-    this->StumpModifyAddTarget(buffer);
+  } else if (target == "bip32_path_parse") {
+    this->Bip32PathParseTarget(buffer);
```
quapka, line 788: This looks like it should be only an addition, but you are removing `stump_modify_add`?

modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
kuliq23: this was still a WIP, it's gone now, thx for the comments

driver.cpp
kuliq23: fixed, thx

kuliq23:
WIP - Still draft PR - working on it

kuliq23:
Left only `RUST_BITCOIN` and `BITCOIN_CORE` to make the contribution bite-sized as I am not sure about it's mergeability.

Modules deal with path strings differently, as I mentioned before:
> Some modules expect an exact format while others accept more general forms. (e.g. "h","H"," ' ") .... 

Due to this, this target fails after a few inputs.
This behavior is not unexpected and not something that I would consider as a real issue...


But we can also observe some interesting results:
```
BIP32 path parse failed for +22
Module: Rustbitcoin
Result: CORRECT
Module: Bitcoin
Result: INVALID
```
Do you think we should add this target?
Thanks

erickcestari:
> > Some modules expect an exact format while others accept more general forms. (e.g. "h","H"," ' ") ....

Hmmm, interesting! Sorry for the late reply. I think an easy way to avoid the fuzzer crashing all the time is to make those more restrictive implementations skip those specific errors in `module.cpp` instead of returning "INVALID", so that the `driver.cpp` doesn't compare the outputs. Since we know it's going to fail anyway.

I'll take a closer look later in the PR. Thanks for working on this!

modules/bitcoin/module.cpp
```diff
@@ -640,6 +641,18 @@ Bitcoin::bip32_deserialize_extended_key(std::span<const uint8_t> buffer) const {
...
+Bitcoin::bip32_path_parse(std::span<const uint8_t> buffer) const {
+  const std::string path_str(reinterpret_cast<const char *>(buffer.data()),
+                             buffer.size());
+  std::vector<uint32_t> keypath;
+  if (!ParseHDKeypath(path_str, keypath)) {
+    // Failed to parse
+    return "INVALID";
+  }
+
+  return "CORRECT";
```
erickcestari, line 725: We could return a formatted string to see if the implementations produced the same uint32 numbers.

erickcestari (changes requested):
```
IndexOutOfRange(IndexOutOfRangeError { index: 3709551615 })
BIP32 path parse failed for 00000003709551615
Module: Rustbitcoin
Result: INVALID
Module: Bitcoin
Result: CORRECT
```

 Bitcoin Core's `ParseHDKeypath` accepts unhardened path components with values ≥ 2^31 (e.g. 3709551615) without error, silently treating them as hardened keys since bit 31 is already set in the raw integer. This means `m/3709551615` and `m/1562068223'` parse identically. The function should reject unhardened components where the parsed value exceeds `2^31 - 1`, as rust-bitcoin does. Might be worth reporting upstream.
 
 Found this by adding some skips to be able to found more meaningful discrepancies.
 
 1. skip these inputs in `driver.cpp` since we know that core will not be able to parse it:
 
 ```c++
  if (path.find('+') != std::string::npos) {
    return;
  }

  if (path.find('h') != std::string::npos) {
    return;
  }
 ```
 
 2.  skip  skip cases where bitcoin core accepts`33/` and rust-bitcoin rejects because it expects an integer after the "/" in rustbitcoin `lib.rs`.
 
 ```rust
 match bitcoin::bip32::DerivationPath::from_str(s) {
        Ok(_) => str_to_c_string("CORRECT"),
        Err(bitcoin::bip32::ParseChildNumberError::ParseInt(e)) => {
            if e.kind() == &std::num::IntErrorKind::Empty {
                return std::ptr::null_mut();
            }
            str_to_c_string("INVALID")
        }
        Err(_) => str_to_c_string("INVALID"),
    }
 ```

quapka:
> skip these inputs in driver.cpp since we know that core will not be able to parse it

You previously suggested skipping the troublesome inputs in `module.cpp`, which seems to provide more information. Say bitcoin does not parse `+`, but two other modules do; then fuzzing all three together could identify a discrepancy in how the other modules interpret `+`.

quapka:
> This means `m/3709551615` and `m/1562068223'` parse identically. The function should reject unhardened components where the parsed value exceeds `2^31 - 1`, as rust-bitcoin does. Might be worth reporting upstream.

I would be interested in hearing their take on that - I've mentioned some of the discrepancies in https://github.com/bitcoin/bitcoin/issues/32201#issuecomment-4097105293 already. However, after re-reading parts of the spec (emphasis mine):

> Each extended key has 2<sup>31</sup> normal child keys, and 2<sup>31</sup> hardened child keys. Each of these child keys has an index. The normal child keys use indices 0 through 2<sup>31</sup>-1. **The hardened child keys use indices 2<sup>31</sup> through 2<sup>32</sup>-1**. To ease notation for hardened key indices, a number i<sub>H</sub> represents i+2<sup>31</sup>.
> ...
> The algorithm to do so depends on whether the child is a hardened key or not (**or, equivalently, whether i ≥ 2<sup>31</sup>**), and whether we're talking about private or public keys.

I think it's just under- or poorly specified in the spec.

modules/bitcoin/module.cpp
kuliq23: I agree, the return value is simplified here to showcase just the constraints for accepting 'pathstrings' as valid and because again, each implementation has it's own` path_to_str` equivalent, if any. 

Populating a formatted string from the internal path representation makes sense and I will incorporate it.

kuliq23:
> Bitcoin Core's `ParseHDKeypath` accepts unhardened path components with values ≥ 2^31 (e.g. 3709551615) without error, silently treating them as hardened keys since bit 31 is already set in the raw integer. This means `m/3709551615` and `m/1562068223'` parse identically. The function should reject unhardened components where the parsed value exceeds `2^31 - 1`, as rust-bitcoin does. Might be worth reporting upstream.

I noticed this behavior earlier and assumed it might be an intentional safeguard. While this basically introduces three ways to denote the same path (` m/3709551615` , ` m/1562068223'` ,` m/3709551615'` ), I assumed it is more beneficial to a user than outright rejecting such inputs, as it tolerates mistakes and still results in a valid derivation.

However, this still underlines the under-specificiation of derivation paths as a whole in BIP32.
I am currently working on a behavior table comparing handling of this and other 'path cases' between implemented modules:

- can start with `m`
- can start with `M`
- must start with `m`
- must start with `M`
- allows `H` for hardened index
- allows `h` for hardened index
- allows `'` for hardened index
- allows index \> 0x7FFFFFFF
- allows index \> 0x7FFFFFFF with `'`|`H`|`h`
- allows `+` in index
- allows alphabetic chars (excl. `m`,`M`,`h`and`H`)
- allows depth > 255

I'd be happy to to document more edgecases if you can think of any!

Thank you both for exploring this!

quapka:
> * allows `´` for hardened index
> * allows index > 0x7FFFFFFF with `´`|`H`|`h`

Where have you seen `´`? I thought it was the single quote character `'`.

> * allows H for hardened index

Also, I wonder what the source of truth should be for the hardening indicators. After skimming the BIP32, I can't find it. There are mentions in [BIP380](https://github.com/bitcoin/bips/blob/master/bip-0380.mediawiki?plain=1#L89).

Interestingly, the situation with `H` changed a few times last year. Originally, it was supported, but in conflict with that, it was also part of an invalid test vector. This changed with https://github.com/bitcoin/bips/pull/1803. But then it got reverted in https://github.com/bitcoin/bips/pull/1888 while https://github.com/bitcoin/bitcoin/pull/32788 was being implemented...

> * allows `+`,`-` in index

From what I have seen, this is due to the index being parsed via a function for parsing (unsigned) integers, e.g., [NBitcoin](https://github.com/MetacoSA/NBitcoin/blob/6943d663c0a2f0ea0a59e78cfd2dd7ead882d97b/NBitcoin/BIP32/KeyPath.cs#L114) (see some [docs](https://learn.microsoft.com/en-us/dotnet/api/system.uint32.tryparse?view=net-10.0#system-uint32-tryparse(system-string-system-uint32@))). Point being whatever is parseable as an unsigned int could be potentially seen as a valid item in a path for certain languages.

kuliq23:
> > * allows `´` for hardened index
> 
> > * allows index > 0x7FFFFFFF with `´`|`H`|`h`
> 
> 
> 
> Where have you seen `´`? I thought it was the single quote character `'`.

Mistake(s) fixed, thanks.

(Also removed `-` from "`+`,`-`")

Thank you for all the comments!

kuliq23:
@erickcestari I think the approach with skipping in individual `module.cpp`s is the most logical. I plan using this approach on a `derive_from_path` target and leave the `path_parse` target as is. Reasons:

- `derive_from_path` is more use-case realistic - makes more sense to test it
- some modules (e.g. rustbitcoin) parse paths of depth > 255 successfully, but later refuse to derive using them - testing parsing only does not show the actual behavior
- the discrepancies in `path_parse` are expected as there is no single strict source of truth for BIP32 path formatting 

So overall, I’d treat `path_parse` more as a diagnostic/behavior comparison target, and focus consistency checks on `derive_from_path`, where differences are more meaningful in practice.

Does this approach make sense to you? Do you think `path_parse` should be merged in its current form, or do we prefer to drop it? Thanks

erickcestari:
> @erickcestari I think the approach with skipping in individual `module.cpp`s is the most logical. I plan using this approach on a `derive_from_path` target and leave the `path_parse` target as is. Reasons:
> 
> * `derive_from_path` is more use-case realistic - makes more sense to test it
> * some modules (e.g. rustbitcoin) parse paths of depth > 255 successfully, but later refuse to derive using them - testing parsing only does not show the actual behavior
> * the discrepancies in `path_parse` are expected as there is no single strict source of truth for BIP32 path formatting
> 
> So overall, I’d treat `path_parse` more as a diagnostic/behavior comparison target, and focus consistency checks on `derive_from_path`, where differences are more meaningful in practice.
> 
> Does this approach make sense to you? Do you think `path_parse` should be merged in its current form, or do we prefer to drop it? Thanks

Good question! I think we can keep this is as draft for now. `derive_from_path` target will provide us more meaningful results and and also cover the code that `path_parse` is covering. I think when we have `derive_from_path` working and we have sure that `path_parse` doesn't cover any new code we can close this PR.

> You previously suggested skipping the troublesome inputs in `module.cpp`, which seems to provide more information. Say bitcoin does not parse `+`, but two other modules do; then fuzzing all three together could identify a discrepancy in how the other modules interpret `+`.

Indeed! It does make more sense to skip it directly in the `module.cpp`

PR #424, opened 2026-02-09
refactor: extract module loading, move init out of fuzzer hot path

erickcestari:
Move all initialization into a one-time static initializer, extracting module loading logic into module_loader.{h,cpp}. Previously every fuzzer iteration re-created the Driver, reloaded modules, and called getenv.
Running these only once yields ~2x throughput improvement (~51k to ~104k exec/s) on lightweight targets.


brunoerg:
Concept ACK - going to review it tomorrow.

main.cpp
```diff
@@ -1,266 +1,21 @@
...
-
-static void InitializeRegistryOnce() {
-  std::string error = bitcoinfuzz::ModuleRegistry::instance().initialize();
-  if (!error.empty()) {
-    std::cerr << error << std::endl;
-    std::abort();
-  }
+static const char *g_target = nullptr;
+
+static void Initialize() {
```
brunoerg: It looks good but maybe we could use `LLVMFuzzerInitialize`?

erickcestari: Yes, it should have the same behavior as before. Updated!

brunoerg (approved):
code review ACK 0c6881454d024fd828dd506eba5fbd6f69f6f13f

PR #426, opened 2026-02-10
Fix some nits at decode_onion

erickcestari:


custommutator/mutators/onion.h
```diff
@@ -522,16 +523,21 @@ static size_t serialize_tlv_hop_payload(const TLVHopPayload &payload,
...
 
+static std::vector<uint8_t> randomize_and_mutate_buffer(std::vector<uint8_t> vec,
+                                        size_t max_value) {
+  vec.resize(max_value);
+  for (size_t i = 0; i < vec.size(); i++) {
+    vec[i] = static_cast<uint8_t>(rand() & 0xFF);
+  }
+  LLVMFuzzerMutate(vec.data(), vec.size(), vec.size());
+  return vec;
+}
```
morehouse: - the `vec` parameter is unnecessary
- we should follow the `generate_X` naming convention we've established
- `size` is a more accurate name than `max_value`


```suggestion
static std::vector<uint8_t> generate_buffer(size_t size) {
  std::vector<uint8_t> vec(size);
  for (size_t i = 0; i < vec.size(); i++) {
    vec[i] = static_cast<uint8_t>(rand() & 0xFF);
  }
  LLVMFuzzerMutate(vec.data(), vec.size(), vec.size());
  return vec;
}
```

erickcestari: Updated!

custommutator/mutators/onion.h
```diff
@@ -522,16 +523,20 @@ static size_t serialize_tlv_hop_payload(const TLVHopPayload &payload,
...
 static std::vector<uint8_t> generate_tu_int(size_t max_bytes) {
   std::vector<uint8_t> result;
   size_t actual_bytes = 1 + (rand() % max_bytes);
-  result.resize(actual_bytes);
-
-  for (size_t i = 0; i < actual_bytes; i++) {
-    result[i] = static_cast<uint8_t>(rand() & 0xFF);
-  }
-  LLVMFuzzerMutate(result.data(), actual_bytes, actual_bytes);
+  result = generate_buffer(actual_bytes);
```
morehouse: Nit: declare and define `result` in the same line

```c++
  size_t actual_bytes = 1 + (rand() % max_bytes);
  std::vector<uint8_t> result = generate_buffer(actual_bytes);
```

custommutator/mutators/onion.h
morehouse: Nit: move definition of `value` to within the cases where it is used

morehouse (approved):

custommutator/mutators/onion.h
erickcestari: Done!

custommutator/mutators/onion.h
erickcestari: Done!

morehouse (approved):
LGTM

PR #429, opened 2026-02-11
Feat/differential static bitcoinkernel

KY-U:
# Differential static libbitcoinkernel modules

Alternative implementation for #406.

## Status
- Ready for review.
- Commit history cleaned up.

## Summary
This PR adds support for loading **two static `libbitcoinkernel` builds** into the same `bitcoinfuzz` binary.

The implementation supports comparing **any two revisions or builds**, including identical commits built with different configurations.

Each module links its own static `libbitcoinkernel`. To make this possible within a single final binary, one of the modules applies symbol renaming to avoid symbol collisions.

## Build selection
The two compared builds are selected through:

- `BITCOINKERNEL_REF`
- `BITCOINKERNEL_VARIANT_REF`

If either variable is not set, its default value is `master`.

## Current layout
This PR currently provides two static modules:

- `BitcoinKernel`
- `BitcoinKernelVariant`

Both modules can expose the same fuzz target interface, allowing the same target to be executed against two statically linked `libbitcoinkernel` builds in the same process.

## Local validation
To compare two refs explicitly:

```bash
export BITCOINKERNEL_REF=<branch-or-commit>
export BITCOINKERNEL_VARIANT_REF=<branch-or-commit>
```

If either variable is not provided, it defaults to `master`.

Then build the modules and `bitcoinfuzz` as usual, and run a differential target such as:

```bash
FUZZ=kernel_block ./bitcoinfuzz -runs=1
```

## Notes
- This approach enables behavior comparison between two `libbitcoinkernel` builds without relying on runtime dynamic loading.
- It can also be used to compare builds produced from the same revision with different build configurations.
- When linking two static `libbitcoinkernel` builds into the same binary, **AddressSanitizer may introduce ODR/IDR-related conflicts**, depending on how the builds are produced.
- In practice, this may require **at least one side** of the comparison to be built without AddressSanitizer.


brunoerg:
Needs rebase.

KY-U:
> Needs rebase.

Done. The poc is now working with the recent module loading refactor.

brunoerg:
From ci:
```sh
Run make -k check-format-all
make[1]: Entering directory '/home/runner/work/bitcoinfuzz/bitcoinfuzz'
clang-format -Werror --fail-on-incomplete-format -n include/bitcoinfuzz/*.h include/bitcoinfuzz/*.cpp driver.cpp driver.h main.cpp helpers/*.cpp helpers/*.h
driver.cpp:499:9: error: code should be clang-formatted [-Wclang-format-violations]
    // }
```

You can fix it by running `make format-all`.

brunoerg:
cc @sedited @stickies-v 

it might be interesting for you, the idea is having version-based differential fuzzing of libbitcoinkernel bindings.

erickcestari (changes requested):
Needs rebase and format the code.

erickcestari (commented):
> Is ASan strictly required?

No, we probably don't need to the bitcoinkernel to be instrumented with ASan since we want max performance and ASan would be very unlikely to find something in this scenario.

> Build automation for .a: should we use git submodule, or another approach? Is it necessary? Given there is only one implementation at the master branch of btccore, would the usecase only be comparing different commits?

This probably can be added in a follow-up PR, but we could automatize it. I have thought we could use some shell script to basically receive 2 commits, where it would clone the bitcoin core, checkout to the commit, build the libbitcoinkernel with the first commit and then checkout to the other commit and compile the other libbitcoinkernel.

> Prefix before compile: does it make sense to apply the symbol prefix at build time to run two ASan versions?

In this case, we would not use ASan.

> Proposed flow: user provides the branch/commit to compare; one module builds master while the other fetches the specific commit. Is this model acceptable?

I've mentioned a better flow above, the commit in the master branch could be a default value in case the first commit wasn't provided, per example.

KY-U:
> > Is ASan strictly required?
> 
> No, we probably don't need to the bitcoinkernel to be instrumented with ASan since we want max performance and ASan would be very unlikely to find something in this scenario.
> 
> > Build automation for .a: should we use git submodule, or another approach? Is it necessary? Given there is only one implementation at the master branch of btccore, would the usecase only be comparing different commits?
> 
> This probably can be added in a follow-up PR, but we could automatize it. I have thought we could use some shell script to basically receive 2 commits, where it would clone the bitcoin core, checkout to the commit, build the libbitcoinkernel with the first commit and then checkout to the other commit and compile the other libbitcoinkernel.
> 
> > Prefix before compile: does it make sense to apply the symbol prefix at build time to run two ASan versions?
> 
> In this case, we would not use ASan.
> 
> > Proposed flow: user provides the branch/commit to compare; one module builds master while the other fetches the specific commit. Is this model acceptable?
> 
> I've mentioned a better flow above, the commit in the master branch could be a default value in case the first commit wasn't provided, per example.

Thanks for the feedback.

I've removed the address sanitizer flag from the "second" kernel module, If the "fuzzer" sanitizer also isn't necessary on both modules, I suppose I could remove that flag too.

I've also followed your suggestion to checkout from an already downloaded btccore repo. The only nuance is that, in order to avoid having both modules sharing the same branch/working tree, nn the second module's [Makefile](https://github.com/KY-U/bitcoinfuzz/blob/feat/differential-static-bitcoinkernel/modules/bitcoinkernelbranch/Makefile#L97), the core repo is cloned utilizing the first repo as a reference (**--reference** flag), which shares the same git objects as the reference repo, but creates a separate working tree.

Just one final point I would like to confirm, once the approach on this draft is approved, should I open another "normal" PR or simply rebase this one?

KY-U:
In the last force-push rebase:
- Removed `library_tag` test target.
- Simplified header logic. 
    - Modules no longer need to manually include Bitcoinkernel headers.
- Simplified prefix header logic.
    - Removed the unnecessary prefix header wrapper.
- Renamed all `branch` references  to `variant` to better reflect intent.
- Removed sanitizers from Bitcoinkernel build process.
- Squashed/rebased changes into 2 commits, one for each module.

I believe the PR is now ready for review.

erickcestari:
Should we also put it in CI/docker? It'll probably increase the time for CI finish, but I think is worth having it.
@brunoerg

include/bitcoinfuzz/module_defs.h
```diff
@@ -65,4 +65,11 @@ MODULE_ENTRY(RUST_K256, "RUST_K256", K256)
 #endif
 #ifdef LIBWALLY_CORE
 MODULE_ENTRY(LIBWALLY_CORE, "LIBWALLY_CORE", LibwallyCore)
+#endif
+#ifdef BITCOINKERNEL
+MODULE_ENTRY(BITCOINKERNEL, "BITCOIN_KERNEL", BitcoinKernel)
+#endif
+#ifdef BITCOINKERNEL_VARIANT
+MODULE_ENTRY(BITCOINKERNEL_VARIANT, "BITCOIN_KERNEL_VARIANT",
+             BitcoinKernelVariant)
```
erickcestari:
```suggestion
#endif
#ifdef BITCOINKERNEL
MODULE_ENTRY(BITCOINKERNEL, "BITCOINKERNEL", BitcoinKernel)
#endif
#ifdef BITCOINKERNEL_VARIANT
MODULE_ENTRY(BITCOINKERNEL_VARIANT, "BITCOINKERNEL_VARIANT",
             BitcoinKernelVariant)
```

erickcestari (changes requested):

modules/bitcoinkernel/Makefile
```diff
@@ -0,0 +1,77 @@
...
+		-DENABLE_IPC=OFF \
+		-DBUILD_BITCOIN_BIN=OFF \
+		-DBUILD_DAEMON=OFF \
+		-DBUILD_CLI=OFF \
+		-DBUILD_TESTS=OFF \
+		-DBUILD_TX=OFF \
+		-DBUILD_UTIL=OFF \
+		-DBUILD_KERNEL_TEST=OFF \
+		-DENABLE_WALLET=OFF \
+		-DBUILD_WALLET_TOOL=OFF \
```
erickcestari, line 37: The bitcoinkernel isn't being compiled with coverage instrumentation

```suggestion
		-DBUILD_WALLET_TOOL=OFF \
		-DSANITIZERS=fuzzer \
```

erickcestari (changes requested):

include/bitcoinfuzz/module_defs.h
KY-U: Done.

modules/bitcoinkernel/Makefile
KY-U: Done.

erickcestari:
It's missing the docs of how to use it.

modules/bitcoinkernel/Makefile
```diff
@@ -0,0 +1,78 @@
...
+	$(CXX) $(CXXFLAGS) -I $(BITCOIN_DIR)/src -fPIC -c $< -o $@
+
+format:
+	clang-format -i module.h module.cpp
+
+check-format:
+	clang-format -Werror --fail-on-incomplete-format -n module.h module.cpp
+
+clean:
+	rm -rf *.o module.a
```
erickcestari: let's also clean the bitcoin core build

```suggestion
clean:
	rm -rf *.o module.a bitcoin/build-kernel
```

modules/bitcoinkernelvariant/Makefile
```diff
@@ -0,0 +1,127 @@
+all: module.a
+
+CXXFLAGS += -Wall -Wextra -std=c++20 -I ../../include
+SYMBOL_PREFIX ?= variant
+BITCOIN_DIR := bitcoin
```
erickcestari, line 6: Bitcoin core is being cloned twice, one time in bitcoinkernel and other in bitcoinkernelvariant. We should also first clean the bitcoin core code before compiling the variant

```suggestion
BITCOIN_DIR := ../bitcoinkernel/bitcoin
```

erickcestari (changes requested):

modules/bitcoinkernelvariant/Makefile
KY-U: bitcoinkernelvariant is [cloned](https://git-scm.com/docs/git-clone) with the bitcoinkernel dir being passed to the `--reference` flag, which means that if the reference repository i.e. bitcoinkernel, is present, `.git/objects/info/alternates` is automatically setup to obtain objects from the reference repository, reducing network and local storage costs.

KY-U:
> It's missing the docs of how to use it.

Added.

modules/bitcoinkernel/Makefile
KY-U: Done.

modules/bitcoinkernelvariant/Makefile
erickcestari: > bitcoinkernelvariant is [cloned](https://git-scm.com/docs/git-clone) with the bitcoinkernel dir being passed to the `--reference` flag, which means that if the reference repository i.e. bitcoinkernel, is present, `.git/objects/info/alternates` is automatically setup to obtain objects from the reference repository, reducing network and local storage costs.

Makes sense. But does the bitcoinkernel variant need to wait for the bitcoinkernel to finish compilint? Wouldn't they have a different copy of source files which means they both can compile at the same time without any problem?

KY-U: You are correct, good catch.
- Modified the build rule to only require the reference git repo.
- Renamed the "wait" rule to better communicate the specific condition.

KY-U:
Fixed some build inconsistencies catched by the docker image check.

modules/bitcoinkernelvariant/Makefile
```diff
@@ -0,0 +1,127 @@
...
+		-DBUILD_BITCOIN_BIN=OFF \
+		-DBUILD_DAEMON=OFF \
+		-DBUILD_CLI=OFF \
+		-DBUILD_TESTS=OFF \
+		-DBUILD_TX=OFF \
+		-DBUILD_UTIL=OFF \
+		-DBUILD_KERNEL_TEST=OFF \
+		-DENABLE_WALLET=OFF \
+		-DBUILD_WALLET_TOOL=OFF \
+		-DENABLE_EXTERNAL_SIGNER=OFF \
```
erickcestari, line 73: In the bitcoinkernelvariant we are also missing the coverage instrumentation
```suggestion
		-DENABLE_EXTERNAL_SIGNER=OFF \
		-DSANITIZERS=fuzzer \
```

modules/bitcoinkernelvariant/Makefile
```diff
@@ -0,0 +1,127 @@
...
+OBJCOPY := $(shell command -v objcopy 2> /dev/null)
+ifeq ($(OBJCOPY),)
+$(error objcopy not found. Install with: apt-get install binutils)
+endif
+
+# Map globally defined symbols in libbitcoinkernel.a, excluding weak functions (Ww)
+variant_symbols.map: $(KERNEL_LIB)
+	@echo "Generating deterministic symbol map..."
+	@nm -g --defined-only $(KERNEL_LIB) | awk '$$2 !~ /^[Ww]$$/ {print $$3}' | \
+	grep -Ev '^$$' | sort -u | while read sym; do \
```
erickcestari: The variant module wasn't actually running its own copy of the kernel code due to two symbol resolution issues:

  1. Weak symbols excluded from rename map — the `nm filter (!~ /^[Ww]$/)` skipped weak symbols, which include most of the deserialization code (`CTransaction`, `UnserializeTransaction`,  `ReadCompactSize`). These are C++ template instantiations emitted as weak symbols. The variant's entry points ended up calling the original kernel's internal functions.
  
   2. `COMDAT` group signatures not renamed — `objcopy --redefine-syms` renamed function symbols but not `COMDAT` group keys (type n), so the linker deduplicated the variant's template function bodies as duplicates of the original's.

  After fixing all two, the symbol map grows from 1,727 → 3,123 entries and coverage roughly doubles when both modules are loaded.


```suggestion
	@{ nm -g --defined-only $(KERNEL_LIB) | awk '{print $$3}'; \
	   nm --defined-only $(KERNEL_LIB) | awk '$$2 == "n" {print $$3}'; \
	} | grep -Ev '^$$' | sort -u | while read sym; do \
```

erickcestari (changes requested):

modules/bitcoinkernelvariant/Makefile
KY-U: Thanks for the detailed feedback!

You were right that I had not accounted for C++ template instantiations here. While renaming only the strong symbols was enough to avoid direct symbol conflicts, it did not actually ensure that the variant module used its own isolated copy of the kernel code.

I have now updated the symbol map generation logic to include those weak symbols and added the COMDAT group signature renaming as suggested so the linker does not deduplicate the variant's template bodies.

KY-U:
Rebased on the latest upstream.

KY-U:
> Should we also put it in CI/docker? It'll probably increase the time for CI finish, but I think is worth having it. @brunoerg

Added to CI.

erickcestari (approved):
Nice work @KY-U!! It's looking good. Thanks for working on this!

ACK a55555196991bccb3117947f275ded8a9924fdba

KY-U:
> Nice work @KY-U!! It's looking good. Thanks for working on this!
> 
> ACK [a555551](https://github.com/bitcoinfuzz/bitcoinfuzz/commit/a55555196991bccb3117947f275ded8a9924fdba)

Great! Thanks for all the constructive feedback!

brunoerg:
Merged, congrats on this, @KY-U!

PR #435, opened 2026-02-14
feat: mock the replaylog to remove memoryReplayLog overhead

erickcestari:


morehouse (approved):
LGTM.

AFAICT, the replay log isn't meaningfully exercised by this fuzz test anyway.

PR #447, opened 2026-02-22
Kickstart differential fuzzing between `utreexo` and `rustreexo`

### LINKED ISSUE #443: Make differential fuzzing for utreexo accumulators
Create a target for Bitcoinfuzz that checks whether different implementations of the utreexo accumulator result in the same internal state by adding and deleting utxos, as they should. 

See: https://github.com/bitcoin/bips/pull/1923

We need to implement this target and add the modules to Bitcoinfuzz.

- [X] [utreexod](https://github.com/utreexo/utreexod)
- [X] [rustreexo](https://github.com/mit-dci/rustreexo)

Edit: It will probably need at least 3 targets, one for Stump, Pollard and Forest accumulator.

Edit2:

- [x] Make differential fuzzing target for Stump accumulator.
- [ ] Make differential fuzzing target for Pollard accumulator.
- [ ] Make differential fuzzing target for Forest accumulator.

cc @Davidson-Souza @luisschwab

luisschwab:
Closes #443

## Changelog

This PR adds two new modules: 
1. [`utreexo`](https://github.com/utreexo/utreexo), the Go implementation,
2. [`rustreexo`](https://github.com/mit-dci/rustreexo), the Rust implementation,

and implements a single `stump_verify_add` target, which checks whether the state of both `Stump`s are a match after adding the same leaf hashes. Note that this target does not check deletions.

## Testing

```shell
just run stump_modify

# OR

cd modules/utreexo && make && export CXXFLAGS="$CXXFLAGS -DUTREEXO"
cd ../..
cd modules/rustreexo && make && export CXXFLAGS="$CXXFLAGS -DRUSTREEXO"
cd ../..
make && FUZZ=stump_modify_add ./bitcoinfuzz
```

PS: 1 second of fuzzing was enough to find a crash!

I've intentionally left the "verify" targets as unimplemented, which will be tackled on a follow up.

The first 4 commits fix some lint warnings and add a `compile_flags.txt` file so `clangd` works properly.


brunoerg (changes requested):
Concept ACK on adding this target but strong NACK on these macOS commits. No serious fuzzing development is done on macOS and we recently decided to not support it anymore (especially because of the cost to maintain it was high).

modules/rustreexo/rustreexo_lib/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+///  - The [hashes](BitcoinNodeHash) of new transaction outputs
+///  - The [hashes](BitcoinNodeHash) of spent transaction outputs
+///  - The [`Proof`]s for the spent transaction outputs
+pub unsafe extern "C" fn rustreexo_stump_modify(
+    buffer: *const u8,
+    buffer_len: usize,
+) -> *mut c_char {
+    // We need at least 2 bytes for `new_txout_count`.
+    if buffer_len < 2 {
+        return ptr::null_mut();
```
erickcestari: You should only return a null pointer if you want to skip the output comparison which is used in very specific scenarios and should usually be documented the way it should skip. Probably in both wrapper.go and in the lib.rs you can replace to return an empty string and it should continuous working fine.

```suggestion
        return str_to_c_string("");
```

erickcestari (changes requested):

erickcestari:
It should also have structured input for the targets. When reading the bytes from the input, we could check the byte before the 32 bytes. If it's 0, we could try to remove the hash from the tree. If it's 1, we could add the hash.

I think we could put this logic in the `driver.cpp` file in the target function. You can easily do that with the `FuzzerDataProvider` and send two vectors: one with the hashes to be added and the other with the hashes to be removed. We probably don't need the first u16 number to indicate how many hashes to add.

luisschwab:
@erickcestari we can't do this since we would also need to create a `MemForest` to generate the proofs for TXOuts being removed. This should be done on another target.

luisschwab:
Also, this commit-by-commit workflow is not behaving correctly. The 1st commit fails due to a missing `rustreexo` module even though it was introduced in the 4th commit. cc @brunoerg @erickcestari

erickcestari:
> @erickcestari we can't do this since we would also need to create a `Forest` to generate the proofs for TXOuts being removed. This should be done on another target.

How would the target that remove hashes from the accumulator work? You also probably would need to add the hashes into the stump first to then remove, at the end it does make sense to be in the same target.

Maybe we could have the `Forest` in this target only to generate the `Proofs` for us?

luisschwab:
> How would the target that remove hashes from the accumulator work? You also probably would need to add the hashes into the stump first to then remove, at the end it does make sense to be in the same target.

What you describe could be a different target and I can rename this one to `stump_modify_add_only`. To remove them we would need to keep both `Stump` and `MemForest` so we can generate the relevant proofs. It would be a lot more complicated. I can implement this scenario on a `MemForest` specific target.

modules/rustreexo/rustreexo_lib/src/lib.rs
luisschwab: I returned `ptr::null_mut` here because `buffer_len` can be <2 bytes, so we would not have the necessary amount of bytes for the hashes. We want it to be a no-op in this case, right? Let me know if I'm missing any context here.

brunoerg:
> Also, this commit-by-commit workflow is not behaving correctly. The 1st commit fails due to a missing `rustreexo` module even though it was introduced in the 4th commit. cc @brunoerg @erickcestari

Just opened an issue - https://github.com/bitcoinfuzz/bitcoinfuzz/issues/450.

compile_flags.txt
```diff
@@ -0,0 +1,7 @@
+-std=c++20
```
brunoerg, line 1: I was inclined to remove this file but I think we can keep it. Since we don't generate a `compile_commands.json`, having this file is useful.

modules/rustreexo/rustreexo_lib/src/lib.rs
erickcestari: You could also return an empty string in both module, that would have the same behavior as skipping.

This way is also fine.

modules/rustreexo/rustreexo_lib/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+        slice::from_raw_parts(buffer.add(2), new_txout_count * 32)
+            .chunks_exact(32)
+            .map(|chunk| BitcoinNodeHash::new(chunk.try_into().unwrap()))
+            .collect();
+
+    // Create a new stump and add the new txout hashes to it.
+    let stump = Stump::new();
+    let (stump, _update_data) = match stump.modify(&new_txout_hashes, &[], &Proof::default()) {
+        Ok(s) => s,
+        Err(_) => return ptr::null_mut(),
```
erickcestari: We also should update the rust code to return an empty string instead of null pointer.

```suggestion
        Err(_) => return str_to_c_string(""),
```

erickcestari (changes requested):

modules/rustreexo/rustreexo_lib/src/lib.rs
luisschwab: Fixed

brunoerg:
Can you please rebase? The commit-by-commit issue has been fixed (I guess).

luisschwab:
@brunoerg all green now

modules/utreexo/utreexo_wrapper/wrapper.go
```diff
@@ -0,0 +1,74 @@
...
+//export UtreexoStumpUpdate
+func UtreexoStumpUpdate(buffer *C.char, buffer_len C.int) *C.char {
+	buffer_bytes := C.GoBytes(unsafe.Pointer(buffer), buffer_len)
+	// We need at least 2 bytes for `new_txout_count`.
+	if buffer_len < 2 {
+		return nil
+	}
+
+	// Define the amount of new inputs from the first 2 bytes of the buffer.
+	newTxoutCount := int(binary.BigEndian.Uint16(buffer_bytes[:2]))
```
brunoerg: The approach is not correct (or elegant) doing this way. You're basically sending the raw buffer to the modules (utreexo and rustreexo) and letting them to define the amount of new inputs, build the txout hashes, etc. You could simplify the code in the utreexo and rustreexo modules by consuming what's required on `driver` (using `FuzzedDataProvider`) and sending them exactly what they need.

luisschwab: Makes sense, I'll fix it up and push in a few hours.

brunoerg: Were you able to do it?

driver.cpp
```diff
@@ -849,6 +850,42 @@ void Driver::DecodeOnionTarget(std::span<const uint8_t> buffer) const {
...
+  size_t add_hashes_count = provider.ConsumeIntegralInRange<size_t>(0, 64);
+
+  std::vector<std::vector<uint8_t>> add_hashes;
+  add_hashes.reserve(add_hashes_count);
+  for (size_t i = 0; i < add_hashes_count; i++) {
+    auto hash = provider.ConsumeBytes<uint8_t>(32);
+    if (hash.size() < 32)
+      return;
+    add_hashes.push_back(std::move(hash));
+  }
```
erickcestari, line 865: Why do we need the hashes count? Can't we only do this?


```suggestion
std::vector<std::vector<uint8_t>> add_hashes;
  while (true) {
    auto hash = provider.ConsumeBytes<uint8_t>(32);
    if (hash.size() < 32) {
      break;
    }
    add_hashes.push_back(std::move(hash));
  }
```

modules/rustreexo/rustreexo_lib/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+        Ok(s) => s,
+        Err(_) => return str_to_c_string(""),
+    };
+
+    // Serialize the Stump into a hex string.
+    let mut stump_ser: Vec<u8> = Vec::new();
+    match stump.serialize(&mut stump_ser) {
+        Ok(_) => str_to_c_string(&hex::encode(stump_ser)),
+        Err(_) => ptr::null_mut(),
+    }
```
erickcestari: We should use `expect` here since it should always be able to serialize.

```suggestion
let mut stump_ser: Vec<u8> = Vec::new();
stump.serialize(&mut stump_ser).expect("Vec<u8> serialization should not fail");
str_to_c_string(&hex::encode(stump_ser))
```

erickcestari (changes requested):

erickcestari (approved):
Thanks!
ACK 8e2a2852f9f4ce0d15e09f368c387271ad43fe09

brunoerg (approved):
ACK 8e2a2852f9f4ce0d15e09f368c387271ad43fe09

PR #452, opened 2026-02-26
fix: delegate commit-by-commit builds to a script in the checked-out …

### LINKED ISSUE #450: Bug on commit-by-commit workflow
Commit-by-commit has a problem. If you have a new module added on commit B, it tries to build this module on top of commit A, which will obviously fail.  The issue is that the matrix checkout checks out each commit independently, but all build steps are statically defined in the workflow — they don't reflect what actually existed at that commit .The proper fix would be to run the build logic from within the checked-out commit itself rather than from the workflow file, e.g:

```sh
- name: Build all modules
  timeout-minutes: 60
  run: |
    make all  # or ./ci/build.sh or whatever makes sense 
```

brunoerg:
…commit

Fixes #450

The workflow had hardcoded build steps for every known module, so checking out an earlier commit (e.g. before a new module was added) would fail trying to build directories that didn't exist yet.

Add scripts/build-all.sh that dynamically discovers and builds every modules/*/ directory present at the checked-out commit. The workflow now calls this script instead of maintaining a static list of steps, so each commit defines its own set of modules. A graceful fallback is included for commits that predate the script.

Also moves the embit pip install into the script (guarded by a file existence check) to avoid a similar failure when embit is absent.

### DIFF
--- a/.github/workflows/commit-by-commit-build.yml
+++ b/.github/workflows/commit-by-commit-build.yml
@@ -99,14 +99,11 @@ jobs:
         with:
           dotnet-version: "9.0.x"
 
-      - name: Setup Python and install embit
+      - name: Setup Python
         uses: actions/setup-python@v5
         with:
           python-version: "3.11"
-      - run: |
-          python -m pip install --upgrade pip
-          pip install -r modules/embit/requirements.txt
-          pip install mako
+      - run: python -m pip install --upgrade pip
 
       - name: Install build dependencies
         run: |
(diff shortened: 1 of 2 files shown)


KY-U:
Doesn’t the [.github/scripts](https://github.com/bitcoinfuzz/bitcoinfuzz/tree/v2/.github/scripts) folder already serve the purpose of storing GitHub Actions scripts? Is there a reason to create an additional [scripts](https://github.com/bitcoinfuzz/bitcoinfuzz/tree/v2/scripts) folder at the repository root?

PR #455, opened 2026-03-02
ci(workflows): add docker image check workflow

### LINKED ISSUE #319: Add docker to our CI
We need to test if the Docker image is being built correctly, without errors.

KY-U:
Closes #319

Currently, the proposed workflow simply builds the image with predefined modules. In order to optimize the effectiveness of the image check, I'll make the workflow identify modified modules in the PR and build the image using those specific modules.

## Todo:
- [x] Automate module selection.
## Status
- This PR is now ready for review.

### DIFF
--- a/.github/workflows/docker-image-check.yml
+++ b/.github/workflows/docker-image-check.yml
@@ -1,0 +1,42 @@
+name: Docker Image Check
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths-ignore:
+      - '**/*.md'
+      - LICENSE
+
+jobs:
+  select-modules:
+    runs-on: ubuntu-24.04
+    outputs:
+      cxxflags: ${{ steps.detect.outputs.cxxflags }}
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Detect changed modules and resolve CXXFLAGS
+        id: detect
+        run: |
+          BASE_SHA="${{ github.event.pull_request.base.sha }}"
+          HEAD_SHA="${{ github.event.pull_request.head.sha }}"
+
+          ./.github/scripts/select-cxxflags.sh "$BASE_SHA" "$HEAD_SHA" >> "$GITHUB_OUTPUT"
+
+  docker-build:
+    needs: select-modules
+    runs-on: ubuntu-24.04
+    timeout-minutes: 30
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+
+      - name: Build Docker Image
+        run: |
+          docker build -t bitcoinfuzz:image-check \
+            --build-arg FUZZ=image-check \
+            --build-arg CXXFLAGS="${{ needs.select-modules.outputs.cxxflags }}" \
+            .
--- a/.github/scripts/select-cxxflags.sh
+++ b/.github/scripts/select-cxxflags.sh
@@ -1,0 +1,63 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+BASE_SHA="${1:-}"
+HEAD_SHA="${2:-}"
+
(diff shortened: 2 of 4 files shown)


brunoerg:
Concept ACK

.github/workflows/docker-image-check.yml
```diff
@@ -0,0 +1,42 @@
...
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Detect changed modules and resolve CXXFLAGS
+        id: detect
+        run: |
+          BASE_SHA="${{ github.event.pull_request.base.sha }}"
```
brunoerg, line 24: I'm not sure but when triggered via workflow_dispatch, `github.event.pull_request` is null, so both vars become empty strings. In select-cxxflags.sh with set -euo pipefail, git diff --name-only "" "" will fail and exit.

.github/workflows/docker-image-check.yml
```diff
@@ -0,0 +1,42 @@
...
+    paths-ignore:
+      - '**/*.md'
+      - LICENSE
+  workflow_dispatch:
+jobs:
+  select-modules:
+    runs-on: ubuntu-24.04
+    outputs:
+      cxxflags: ${{ steps.detect.outputs.cxxflags }}
+      selected_modules: ${{ steps.detect.outputs.selected_modules }}
```
brunoerg: Where is this exported `selected_modules` used?

.github/workflows/docker-image-check.yml
```diff
@@ -0,0 +1,42 @@
...
+
+      - name: Detect changed modules and resolve CXXFLAGS
+        id: detect
+        run: |
+          BASE_SHA="${{ github.event.pull_request.base.sha }}"
+          HEAD_SHA="${{ github.event.pull_request.head.sha }}"
+
+          ./.github/scripts/select-cxxflags.sh "$BASE_SHA" "$HEAD_SHA" >> "$GITHUB_OUTPUT"
+
+  docker-build:
```
brunoerg, line 29: I would consider to add a timeout for this, otherwise it might consume CI minutes until the global 6-hour limit.

brunoerg (commented):
Left some comments.

.github/workflows/docker-image-check.yml
KY-U: I forgot to remove `workflow_dispatch` execution in the final commit, it was only for testing.

Removed.

.github/workflows/docker-image-check.yml
KY-U: Good catch, it's unnecessary and redundant!

The only output the job actually needs to export to be consumed by the build job is `cxxflags`.

.github/workflows/docker-image-check.yml
KY-U: Added a 30 minute timeout for docker image build, that should be enough, I think.

KY-U:
The good news is that the workflow has already shown its value by identifying error:

From ci:
```
#19 27.07 module.cpp:(.text._Z14python_cleanupv[_Z14python_cleanupv]+0x0): multiple definition of `python_cleanup()'; modules/embit/module.a(module.o):module.cpp:(.text._Z14python_cleanupv[_Z14python_cleanupv]+0x0): first defined here
```
In [pybitcoinkernel's leak PR](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/457), I forgot to add the `static` key word to embit and pybitcoinkerl modules cleanup routines, not giving internal linkage to the routine function symbols. This is a problem because the routines aren't class methods, but global definitions, so If both modules are built at the same time, we get `multiple definition` errors.

The bad news is that the PR has already been merged. My mistake for not catching that. 

I've added the fix commit to this PR.

brunoerg (approved):
ACK f4839110f82367e6484a0d06b9293865550f54c8

PR #458, opened 2026-03-04
fix: correct loop condition in CompactBlocksTarget function

### LINKED ISSUE #456: [BUG] : Infinite loop and out-of-bounds read in `Driver::CompactBlocksTarget` (missing loop condition)
In `Driver::CompactBlocksTarget`(`driver.cpp`) the byte-dumping loop used when a module mismatch is detected is missing a comparison operator in its loop condition. The loop currently uses:

Image

## Steps to reproduce:

- Build the project and run a scenario that triggers a mismatch in `cmpctblocks_parse` between modules (e.g., fuzz until a mismatch is logged).
- When the mismatch logging runs, the process will hang or crash while printing the buffer bytes.

## Actual behavior:
The program enters an infinite loop and performs out-of-bounds reads, causing undefined behavior (hang, crash, or memory corruption) in the mismatch-reporting path.

## Expected behavior:
The loop should print up to the first 32 bytes (or fewer if `[buffer.size()]< 32`) and then stop.

## Suggested fix:
- Replace the loop with:
Image

nikunjkumar05:
Fixes : #456 

## About PR 
- Fixes an infinite loop and potential out-of-bounds read in `Driver::CompactBlocksTarget` by correcting the for-loop condition in `driver.cpp`.
- The loop header was missing the `i <` comparison, causing the loop to evaluate the wrong expression and iterate incorrectly. This could lead to a non-terminating loop and read past buffer bounds.

## Changes 
corrected loop from `for (size_t i = 0; std::min(size_t(32), buffer.size()); ++i)` to `for (size_t i = 0; i < std::min(size_t(32), buffer.size()); ++i)`.

## Before 
Screenshot 2026-03-04 210155

## After
Screenshot 2026-03-04 210557

### DIFF
--- a/driver.cpp
+++ b/driver.cpp
@@ -392,7 +392,7 @@ void Driver::CompactBlocksTarget(std::span<const uint8_t> buffer) const {
     if (last_response.has_value()) {
       if (*res != *last_response) {
         if (!buffer.empty()) {
-          for (size_t i = 0; std::min(size_t(32), buffer.size()); ++i)
+          for (size_t i = 0; i < std::min(size_t(32), buffer.size()); ++i)
             printf("%02x", buffer[i]);
           if (buffer.size() > 32)
             std::cout << "...";


brunoerg:
Nice catch, thank you!

brunoerg (approved):
ACK 05580f8b640a41e309a42c7414cc0f1cf1f545f4

PR #463, opened 2026-03-05
fix(driver): improve error handling for undefined targets instead of assert(false)

### LINKED ISSUE #462: [Improvement] : Replace unconditional `assert(false)` calls that crash process
In `driver.cpp` call [`assert(false)`] for unexpected runtime conditions (e.g. unknown Driver targets). If any of these code paths are reached the process aborts immediately with no useful diagnostics, making fuzzing results and debugging unrecoverable.

## Expected 
Log a clear error message (stderr or logger) with context (function, input summary, target/module) and either:
- return safely from the function; or
- propagate a structured error (exception / error code) so callers can handle, log, and continue

Suggested Fix 
From : 
 `std::cout << "Target not defined!" << std::endl;`
`assert(false);`

To : 

Image

nikunjkumar05:
Fixes : #462 
## About
In `driver.cpp` call [`assert(false)`] for unexpected runtime conditions (e.g. unknown Driver targets). If any of these code paths are reached the process aborts immediately with no useful diagnostics, making fuzzing results and debugging unrecoverable.

## Fix 
image

### DIFF
--- a/driver.cpp
+++ b/driver.cpp
@@ -909,8 +909,8 @@ void Driver::Run(const uint8_t *data, const size_t size,
   } else if (target == "decode_onion") {
     this->DecodeOnionTarget(buffer);
   } else {
-    std::cout << "Target not defined!" << std::endl;
-    assert(false);
+    std::cerr << "Target not defined: " << target << std::endl;
+    return;
   }
 };
 


brunoerg (commented):
NACK. This is wrong and will end up getting an "infinite loop"

PR #467, opened 2026-03-06
fix: add bounds validation in ExtractDestination before copying witness solutions

nikunjkumar05:
Adds size checks before `std::copy` in ExtractDestination for witness destination cases. Without these, a malformed vSolutions buffer could cause an out-of-bounds write (V0 key/script hash, Taproot) or an `OOB` read `(WitnessUnknown)`. Now returns CNoDestination if sizes don't match expectations + using instead of typedef

### DIFF
--- a/modules/bitcoin/addresstype.cpp
+++ b/modules/bitcoin/addresstype.cpp
@@ -15,7 +15,7 @@
 #include <cassert>
 #include <vector>
 
-typedef std::vector<unsigned char> valtype;
+using valtype = std::vector<unsigned char>;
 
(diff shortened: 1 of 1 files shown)


erickcestari:
NACK

This code comes from Bitcoin Core. We typically do not modify it manually; changes are only made when upgrading to a newer version. In any case, these changes are unlikely to be a real issue.

brunoerg:
Also, please stop AI-slop.

nikunjkumar05:
I apologize next time it will not happen

PR #468, opened 2026-03-06
build(eclair): clean external/eclair Maven artifacts during make clean

KY-U:
Eclair module depends on Maven artifacts created in `external/eclair/*/targets`. Without cleaning those directories, `make` routine reutilized those artifacts.

This patch enhances the `make clean` routine to ensure deterministic clean builds.


brunoerg (approved):
ACK ce4e8dc5e837a48715da30d46db147640cec676b

PR #476, opened 2026-03-10
embit: add bip32 master keygen target

kuliq23:


docker-compose.yml
```diff
@@ -242,7 +242,7 @@ services:
       tags:
         - bitcoinfuzz:bip32_master_keygen
       args:
-        CXXFLAGS: "-DRUST_BITCOIN -DNBITCOIN -DBITCOIN_CORE -DBTCD -DBITCOINJ"
+        CXXFLAGS: "-DRUST_BITCOIN -DNBITCOIN -DBITCOIN_CORE -DBTCD -DBITCOINJ -DLIBWALLY_CORE -DEMBIT"
```
quapka, line 245: The `-DLIBWALLY_CORE` was missing, I guess? It's a small thing, but maybe worth adding as a separate PR?

brunoerg: Could be a separate PR or just a separate commit here, no problem.

kuliq23: Yes, will put as separate commit next time...

quapka: I think @brunoerg meant to either
- remove that change from this PR and create another, single-line change, PR, or
- split the commits within this PR and have a single commit adding the `-LIBWALLY_CORE`,

but do not leave it as it is now. The separation of commits:
- helps the reviewers (this would not be a problem here), and
- if ever in the future someone were to use `git bisect` to identify a commit with a breaking change, and it would point to this commit, it could be misleading that an edition of embit had something to do with libwally.

I know the code change here is subtle, so it could seem like a nit, but it's a reasonable practice.

kuliq23: I see, of course, fixed, thanks:)

erickcestari (approved):
Thanks!! Nice work!

ACK 4360dc49654c182d3fadf75b246f481d55c32d3c

docker-compose.yml
erickcestari, line 245: nit: Actually, the Libwally core is being added to Docker in the same commit as the Embit one. Can you add this change to the Libwally commit?

erickcestari (changes requested):

docker-compose.yml
kuliq23: Ofc, thx, my overlook... fixed

erickcestari (changes requested):
nit: can you remove the description "docker: add libwally master keygen to dockerfile" from the embit commit? Also, change the "ci:" prefix to "docker:" in the libwally commit. Thanks!

```
commit e1876632a945976f08141a93f6caec9286505708 (HEAD -> pr-476)
Author: Ondřej Češka <xceska2@fi.muni.cz>
Date:   Mon Mar 16 16:42:54 2026 +0100

    ci: add libwally master keygen to docker

commit 79f8176169f15ea9d28ad37abfb8d780fad14713
Author: Ondřej Češka <xceska2@fi.muni.cz>
Date:   Tue Mar 10 19:05:07 2026 +0100

    embit: add bip32 master keygen target

    docker: add libwally master keygen to dockerfile
```

erickcestari (approved):
Thanks!! Nice work!

ACK b2e9831b15d3b7de083f370e14729f19381bd889

Issue #14, opened 2024-04-29
Add target for coin selection


brunoerg:
A target to perform differential fuzzing between BDK and Bitcoin Core, especially for BnB, would be nice!

evanlinjin:
This is a really cool project, cannot wait to see it take off!

Just a note about the branch-and-bound provided by the `bdk_coin_select` crate... it is unique in the sense that it is actual branch and bound! However, the waste metric is not provided out-of-the-box (although anyone can try implement any metric for bnb). This is because the waste metric is deemed too wasteful in certain scenarios. We only provide a single metric out-of-the-box; we call it the lowest-fee metric which attempts to reduce the fee of the selection while taking into account the cost to spend the change (using the long-term feerate).

Helpful links:

* Some background: https://github.com/bitcoindevkit/coin-select/pull/21#issuecomment-1915811752
* Docs for the `LowestFee` metric: https://docs.rs/bdk_coin_select/latest/bdk_coin_select/metrics/struct.LowestFee.html

evanlinjin:
cc @nymius

nymius:
Hello @brunoerg

I'm working on [coin-selection simulator](https://github.com/nymius/bdk-coin-select-simulation/tree/main). The idea is to have a common framework to run test consistently over different bitcoin coin selection implementations. I mainly followed this [simulator](https://github.com/achow101/coin-selection-simulation), but dropped the node requirements of bitcoin-core test framework and tried to generalize it so more coin-selection projects can implement the needed methods to take advantage of the simulator without much work.

Currently the simulator has interfaces with [bitcoindevkit/coin-select](https://github.com/bitcoindevkit/coin-select), [bitshala/rust-coin-select](https://github.com/Bitshala-Incubator/rust-coinselect) and an independent [python implementation of bitcoin core algorithms](https://github.com/Kevingislason/bitcoin_coin_selection). I implemented them as a proof of concept, the code can be flaky.

Now I'm trying to isolate the bitcoin-core coin-selection core code, looking for testing it without having to work with all the other aspects of a wallet.

I mention all this because I've never applied `fuzzing` myself, besides thinking that I have an idea of how it works. I guess "differential testing" refers to a comparative approach, but cannot get the meaning of "target" in this context.
My poor understanding makes me feel our projects are correlated, but would like to be corrected if not.

Do you already have some considerations about how this should be done?

[...]

PR #480, opened 2026-03-11
feat(bitcoinerlab_miniscript): add bitcoinerlab miniscript module

### LINKED ISSUE #421: Add module bitcoinerlab/miniscript
https://github.com/bitcoinerlab/miniscript

KY-U:
Closes #421

## Build/runtime design

The module does not use Node.js as its embedded runtime.

Instead, the integration is:

1. TypeScript source is bundled together with it's dependencies into a `.js` file with `esbuild`
2. The generated `bundle.js` is converted into C with `qjsc -c`
3. The resulting C artifact is compiled and merged into `module.a`
4. The embedded bundle is executed at runtime through QuickJS

This was chosen deliberately over embedding Node.js.

### Why `esbuild` + QuickJS

- `esbuild` produces a single bundled JS artifact from the TypeScript entrypoint
  and npm dependencies
- QuickJS is much smaller and simpler to embed in-process
- The final module can carry the JS bundle and the JS runtime inside the static
  archive
- Runtime startup and execution are lighter-weight, which is preferable for
  fuzzing

## Current status

`miniscript_parse` target is currently functional.

Seeking Concept ACK before adding more targets

## Note
While this setup is the optimal integration setup that I've found for integrating typescript with the C++ module **until now**, it isn't the only one. Any suggestions or feedback are welcome!


KY-U:
Removed address sanitizer and extra newline

KY-U:
Added format.

brunoerg:
From CI:
```sh
Using corpora for miniscript_parse
/usr/bin/clang++ -fsanitize=address,fuzzer -Wall -Wextra -std=c++20 -I include -I .  -DBITCOIN_CORE -DRUST_MINISCRIPT -DBITCOINERLAB_MINISCRIPT -DBITCOINFUZZ_DIR=\"/home/runner/work/bitcoinfuzz/bitcoinfuzz\"  -c driver.cpp -o driver.o
driver.cpp:853:10: warning: unused variable 'add_hashes_count' [-Wunused-variable]
  853 |   size_t add_hashes_count = provider.ConsumeIntegralInRange<size_t>(0, 64);
      |          ^~~~~~~~~~~~~~~~
1 warning generated.
/usr/bin/clang++ -fsanitize=address,fuzzer -Wall -Wextra -std=c++20 -I include -I .  -DBITCOIN_CORE -DRUST_MINISCRIPT -DBITCOINERLAB_MINISCRIPT -DBITCOINFUZZ_DIR=\"/home/runner/work/bitcoinfuzz/bitcoinfuzz\"  -c include/bitcoinfuzz/basemodule.cpp -o include/bitcoinfuzz/basemodule.o
/usr/bin/clang++ -fsanitize=address,fuzzer -Wall -Wextra -std=c++20 -I include -I .  -DBITCOIN_CORE -DRUST_MINISCRIPT -DBITCOINERLAB_MINISCRIPT -DBITCOINFUZZ_DIR=\"/home/runner/work/bitcoinfuzz/bitcoinfuzz\"  -c include/bitcoinfuzz/modulelogger.cpp -o include/bitcoinfuzz/modulelogger.o
/usr/bin/clang++ -fsanitize=address,fuzzer -Wall -Wextra -std=c++20 -I include -I .  -DBITCOIN_CORE -DRUST_MINISCRIPT -DBITCOINERLAB_MINISCRIPT -DBITCOINFUZZ_DIR=\"/home/runner/work/bitcoinfuzz/bitcoinfuzz\"   main.cpp driver.o include/bitcoinfuzz/basemodule.o include/bitcoinfuzz/modulelogger.o  modules/bitcoin/module.a modules/rustminiscript/module.a modules/bitcoinerlabminiscript/module.a     -o bitcoinfuzz  
clang++: error: no such file or directory: 'modules/bitcoinerlabminiscript/module.a'
```

.github/workflows/workflow.yml
```diff
@@ -143,7 +143,7 @@ jobs:
             asan_options: "detect_leaks=0:detect_container_overflow=0"
           - corpus: "miniscript_parse_clean"
             target: "miniscript_parse"
-            cxxflags: "-DBITCOIN_CORE -DRUST_MINISCRIPT"
+            cxxflags: "-DBITCOIN_CORE -DRUST_MINISCRIPT -DBITCOINERLAB_MINISCRIPT"
```
brunoerg, line 156: Perhaps it's missing to build this module?

KY-U: Indeed, I had forgotten to add the new module to the `Build Modules` step.

README.md
```diff
@@ -119,6 +119,14 @@ If you prefer, you can still build the modules manually. Below are the steps for
     export CXXFLAGS="$CXXFLAGS -DRUST_MINISCRIPT"
     ```
 
+- ### [bitcoinerlab miniscript](https://github.com/bitcoinerlab/miniscript)
+
+    ```bash
+    cd modules/bitcoinerlabminiscript
+    make
```
brunoerg, line 129: `make` should be enough to build it? After installing the requried libraries, I tried to run `make` and got:

```sh
clang++-20 -Wall -Wextra -fsanitize=fuzzer -std=c++20 -I ../../include -fPIC -c module.cpp -o module.o
mkdir -p ts/dist/
./ts/node_modules/.bin/esbuild ./ts/src/index.ts \
	--bundle \
	--platform=node \
	--format=iife \
	--outfile=ts/dist/bundle.js
make: ./ts/node_modules/.bin/esbuild: No such file or directory
make: *** [Makefile:25: ts/dist/bundle.js] Error 127
make: *** Waiting for unfinished jobs....
```

KY-U: Yes, it should be enough.

Since the `BUNDLE_JS` rule depends on the `NODE_MODULES_DIR` rule, `npm ci` should have been ran before and installed all dependencies.

```
$(NODE_MODULES_DIR): $(TS_DIR)/package.json $(TS_DIR)/package-lock.json
	cd $(TS_DIR) && npm ci

$(BUNDLE_JS): $(NODE_MODULES_DIR) $(TS_DIR)/src/index.ts
	mkdir -p $(dir $@)
	$(ESBUILD) $(TS_DIR)/src/index.ts \
		--bundle \
		--platform=node \
		--format=iife \
		--outfile=$@
```
The `npm ci` command should install `esbuild`, since it's declared as a dependency in [package.json](https://github.com/KY-U/bitcoinfuzz/blob/feat/add-bitcoinerlab-miniscript-module/modules/bitcoinerlabminiscript/ts/package.json#L17):

```
  "devDependencies": {
    "esbuild": "^0.27.3",
    "typescript": "^5.9.3"
  }
```
which adds it to the [package.lock](https://github.com/KY-U/bitcoinfuzz/blob/feat/add-bitcoinerlab-miniscript-module/modules/bitcoinerlabminiscript/ts/package-lock.json) that is read by the `npm ci` command.

I have a few questions which may help us figure the root of the problem:

- Is your local machine architecture included in this list?
```
      "bin": {
        "esbuild": "bin/esbuild"
      },
      "engines": {
        "node": ">=18"
      },
      "optionalDependencies": {
        "@esbuild/aix-ppc64": "0.27.3",
        "@esbuild/android-arm": "0.27.3",
        "@esbuild/android-arm64": "0.27.3",
        "@esbuild/android-x64": "0.27.3",
        "@esbuild/darwin-arm64": "0.27.3",
        "@esbuild/darwin-x64": "0.27.3",
        "@esbuild/freebsd-arm64": "0.27.3",
        "@esbuild/freebsd-x64": "0.27.3",
        "@esbuild/linux-arm": "0.27.3",
        "@esbuild/linux-arm64": "0.27.3",
        "@esbuild/linux-ia32": "0.27.3",
        "@esbuild/linux-loong64": "0.27.3",
        "@esbuild/linux-mips64el": "0.27.3",
        "@esbuild/linux-ppc64": "0.27.3",
        "@esbuild/linux-riscv64": "0.27.3",
        "@esbuild/linux-s390x": "0.27.3",
        "@esbuild/linux-x64": "0.27.3",
        "@esbuild/netbsd-arm64": "0.27.3",
        "@esbuild/netbsd-x64": "0.27.3",
        "@esbuild/openbsd-arm64": "0.27.3",
        "@esbuild/openbsd-x64": "0.27.3",
        "@esbuild/openharmony-arm64": "0.27.3",
        "@esbuild/sunos-x64": "0.27.3",
        "@esbuild/win32-arm64": "0.27.3",
        "@esbuild/win32-ia32": "0.27.3",
        "@esbuild/win32-x64": "0.27.3"
      }
```
If not, that may be the problem.

If yes:
- Did `make` create a `bitcoinerlabminiscript/ts/node_modules` folder?
- Did it also create a `.bin` or `esbuild` folders inside `node_modules`?

brunoerg: I had to run `npm ci` into the ts folder, now it worked but I was expecting Makefile to run it.

brunoerg:
Running with our corpora, I found:

```sh
=== BitcoinFuzz Module Activation Log ===
Target: miniscript_parse
Loaded Modules (2):
  [1] Bitcoin
  [2] BitcoinerlabMiniscript
=========================================
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 2436076440
INFO: Loaded 1 modules   (40720 inline 8-bit counters): 40720 [0x569b2d3228c0, 0x569b2d32c7d0), 
INFO: Loaded 1 PC tables (40720 PCs): 40720 [0x569b2d32c7d0,0x569b2d3cb8d0), 
INFO:      634 files found in ../corpora/miniscript_parse
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 164227 bytes
INFO: seed corpus: files: 634 min: 1b max: 164227b total: 1922447b rss: 51Mb
Miniscript parse failed for jjjjjj:pk(8)
Module: BitcoinerlabMiniscript
Result: 1
Module: Bitcoin
Result: 0

```

README.md
KY-U: I still can't reproduce the error locally, but I suspect it has to do with local environment variables at time of build. I've added some temporary logs so I can better diagnose the problem.

Could you please fetch the latest changes and provide me with the return of the `make` command with the verbosity option toggled?

`make --debug=v`

KY-U:
> Running with our corpora, I found:
> 
> ```shell
> === BitcoinFuzz Module Activation Log ===
> Target: miniscript_parse
> Loaded Modules (2):
>   [1] Bitcoin
>   [2] BitcoinerlabMiniscript
> =========================================
> INFO: Running with entropic power schedule (0xFF, 100).
> INFO: Seed: 2436076440
> INFO: Loaded 1 modules   (40720 inline 8-bit counters): 40720 [0x569b2d3228c0, 0x569b2d32c7d0), 
> INFO: Loaded 1 PC tables (40720 PCs): 40720 [0x569b2d32c7d0,0x569b2d3cb8d0), 
> INFO:      634 files found in ../corpora/miniscript_parse
> INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 164227 bytes
> INFO: seed corpus: files: 634 min: 1b max: 164227b total: 1922447b rss: 51Mb
> Miniscript parse failed for jjjjjj:pk(8)
> Module: BitcoinerlabMiniscript
> Result: 1
> Module: Bitcoin
> Result: 0
> ```

Since core's module [uses an integer as an index for the dummy keys arrays](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/modules/bitcoin/module.cpp#L172-L177):

```
/**
 * Context to parse a Miniscript node to and from Script or text representation.
 * Uses an integer (an index in the dummy keys array from the test data) as keys
 * in order to focus on fuzzing the Miniscript nodes' test representation, not
 * the key representation.
 */
```
The [FromString](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/modules/bitcoin/module.cpp#L212-L219) function, which is called from the `miniscript_parse` target
```

  const ParserContext parser_ctx{miniscript::MiniscriptContext::P2WSH};
  auto ret{miniscript::FromString(str, parser_ctx)};
  if (ret && ret->IsSane()) {
    return true;
  }

  const ParserContext parser_ctx_tap{miniscript::MiniscriptContext::TAPSCRIPT};
  ret = miniscript::FromString(str, parser_ctx_tap);
  if (ret && ret->IsSane()) {
    return true;
  }
 ``` 
only accepts a Key with exactly 2 hex chars:
```
  std::optional<Key> FromString(std::span<const char> &in) const {
    if (in.size() != 2)
      return {};
    auto idx = ParseHex(std::string(in.begin(), in.end()));
    if (idx.size() != 1)
      return {};
    return TEST_DATA.dummy_keys[idx[0]];
  }
  ```
 We can see that behaviour illustraded by the fact that, while `jjjjjj:pk(8)` is deemed as invalid by core's module, and valid by the other modules, `jjjjjj:pk(08)` is accepted.
 
 Basically, since core module's keys expect  a dummy 1 byte index, while other modules aren't using the same symbolic representations of keys, in order to allow differential fuzzing between core and the other miniscript modules, we have to either:
 - Align other modules with core module's implicit contract, which does not require changing the btccoremodule. But we'll be rejecting inputs that are valid for the other modules.

 - Create an explicit contract for symbolic representation of keys, and make all modules follow that contract. Which is the more "elegant" solution, but breaks compatibility with existing corpus

README.md
KY-U: Moving PR to draft while this issue isn't solved.

brunoerg: I'll try it later again.

KY-U:
Added a key normalization step when converting a string to a key in bitcoin module's `FromString` so that a key with one character `n` is normalized into the representation `0n`, which can be successfully parsed.

That way, core's module interprets `jjjjjj:pk(8)` as if it was `jjjjjj:pk(08)` and does not reject it, which is the expected result, given that it's a valid miniscript.

The consequence is that bitcoin's module won't be validating the exact same input as the other modules in some cases. But I think it's a reasonable tradeoff to ensure compatibility without enforcing an arbitrary restriction on other modules or implementing a new symbolic representation standard for all modules.

KY-U:
The latest CI errors are related to utreexo build failure, which this PR shouldn't interfere with. Is it possible to run the CI workflow again?

KY-U:
> The latest CI errors are related to utreexo build failure, which this PR shouldn't interfere with. Is it possible to run the CI workflow again?

The same errors appear even after re-running CI. I'll investigate further.

brunoerg:
> > The latest CI errors are related to utreexo build failure, which this PR shouldn't interfere with. Is it possible to run the CI workflow again?
> 
> 
> 
> The same errors appear even after re-running CI. I'll investigate further.

I think I fixed it in #498, take a look at the ci commit please.

brunoerg:
> > > The latest CI errors are related to utreexo build failure, which this PR shouldn't interfere with. Is it possible to run the CI workflow again?
> > 
> > 
> > The same errors appear even after re-running CI. I'll investigate further.
> 
> I think I fixed it in #498, take a look at the ci commit please.

edit: I've opened a PR with the fix, https://github.com/bitcoinfuzz/bitcoinfuzz/pull/502, going to merge soon to not block here.

KY-U:
> > > > The latest CI errors are related to utreexo build failure, which this PR shouldn't interfere with. Is it possible to run the CI workflow again?
> > > 
> > > 
> > > The same errors appear even after re-running CI. I'll investigate further.
> > 
> > 
> > I think I fixed it in #498, take a look at the ci commit please.
> 
> edit: I've opened a PR with the fix, #502, going to merge soon to not block here.

Weird, after rebasing to the latest head, CI is still failing to build `utreexo` even though this PR does not touch that module at all, and other PR's aren't having this problem during the `commit-by-build` phase. I'm still trying to discover the cause.

modules/bitcoinerlabminiscript/Makefile
```diff
@@ -0,0 +1,53 @@
...
+	$(CC) -Wall -Wextra -fsanitize=fuzzer -c $(BUNDLE_C) -o $@
+
+module.o: module.cpp module.h
+	$(CXX) $(CXXFLAGS) -fPIC -c module.cpp -o module.o
+
+format:
+	clang-format -i ./module.cpp ./module.h
+
+check-format:
+	clang-format -Werror --fail-on-incomplete-format -n ./module.cpp ./module.h
```
erickcestari: We could have some formatter for the .ts code.

README.md
```diff
@@ -119,6 +119,14 @@ If you prefer, you can still build the modules manually. Below are the steps for
...
+    export CXXFLAGS="$CXXFLAGS -DBITCOINERLAB_MINISCRIPT"
+    ```
```
erickcestari, line 131: Shouldn't we document that it needs to have quickjs and libquickjs installed?

modules/bitcoinerlabminiscript/Makefile
```diff
@@ -0,0 +1,53 @@
+all: module.a
+
+QUICKJS_LIB_DIR := /usr/lib/quickjs
+QJSC := /usr/bin/qjsc
```
erickcestari: I think we can let the system resolve the path.

```suggestion
QJSC := qjsc
```

modules/bitcoinerlabminiscript/Makefile
```diff
@@ -0,0 +1,53 @@
+all: module.a
+
+QUICKJS_LIB_DIR := /usr/lib/quickjs
```
erickcestari: I received an error when compiling the bitcoinerlabminiscript module. Because, I've installed quickjs from the source code and the files were saved to `/usr/local/lib/quickjs` instead of `/usr/lib/quickjs`.  Maybe we could dynamically find the correct folder?

erickcestari (changes requested):
Nice! I've added some nits to be addressed:

README.md
KY-U: Would it perhaps be better to include quickjs as a submodule to keep the dependency strictly repo-native and make it so the user doesn't have the responsibility to install it?

Update: added quickjs external submodule to facilitate build process. Instructions have been added to README

modules/bitcoinerlabminiscript/Makefile
KY-U: Added prettier formater.

modules/bitcoinerlabminiscript/Makefile
KY-U: With quickj submodule, it is now resolved to a local path.

modules/bitcoinerlabminiscript/Makefile
KY-U: With the quickjs submodule, it is no longer necessary for the user to manualy install quickjs. All envs will always be resolved to the same local paths.

KY-U:
> Nice! I've added some nits to be addressed:

Thanks for the feedback! I've dealt with all the addressed points and the PR finally passed all CI checks!

modules/bitcoin/module.cpp
```diff
@@ -213,18 +213,26 @@ struct ParserContext {
   }
 
   std::optional<Key> FromString(std::span<const char> &in) const {
```
brunoerg, line 217: e80bfa3693cee3496205cb4e6f1899345d407f3f: I would put a brief comment right before this function to explain this change (https://github.com/bitcoinfuzz/bitcoinfuzz/pull/480#issuecomment-4069546765).

README.md
KY-U: Update: added documentation instructing the user to install QuickJs

modules/bitcoinerlabminiscript/Makefile
KY-U: Update: the Makefile dynamically finds the correct path

modules/bitcoinerlabminiscript/Makefile
KY-U: Update: now the Makefile dynamically finds the correct path as suggested.

modules/bitcoin/module.cpp
KY-U: Done.

KY-U:
- Removed quickjs external module.
- Added comment in `modules/bitcoin/module.cpp` explaining the normalization step.

erickcestari (commented):
I'm getting some weird results when running locally. Is this expected?

```
Miniscript parse failed for 0

Module: BitcoinerlabMiniscript
Result: 1
Module: Bitcoin
Result: 0
```

KY-U:
> I'm getting some weird results when running locally. Is this expected?
> 
> ```
> Miniscript parse failed for 0
> 
> Module: BitcoinerlabMiniscript
> Result: 1
> Module: Bitcoin
> Result: 0
> ```

That's not expected. Shouldn't that case be skipped by [this logic](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/driver.cpp#L155-L158)?

Actually, it may be related to the fact that bitcoinerlab's module converts `str.c_str()` to `JS_NewString()`. Currently investigating it.

erickcestari:
> > I'm getting some weird results when running locally. Is this expected?
> > ```
> > Miniscript parse failed for 0
> > 
> > Module: BitcoinerlabMiniscript
> > Result: 1
> > Module: Bitcoin
> > Result: 0
> > ```
> 
> That's not expected. Shouldn't that case be skipped by [this logic](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/driver.cpp#L155-L158)?
> 
> Actually, it may be related to the fact that bitcoinerlab's module converts `str.c_str()` to `JS_NewString()`. Currently investigating it.

The input has spaces and \n characteres.

KY-U:
> > > I'm getting some weird results when running locally. Is this expected?
> > > ```
> > > Miniscript parse failed for 0
> > > 
> > > Module: BitcoinerlabMiniscript
> > > Result: 1
> > > Module: Bitcoin
> > > Result: 0
> > > ```
> > 
> > 
> > That's not expected. Shouldn't that case be skipped by [this logic](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/driver.cpp#L155-L158)?
> > Actually, it may be related to the fact that bitcoinerlab's module converts `str.c_str()` to `JS_NewString()`. Currently investigating it.
> 
> The input has spaces and \n characteres.

Right, thats's exactly it. Bitcoin Core and Rust Miniscript handle the miniscript inputs "as is", while both [Nbitcoin](https://github.com/MetacoSA/NBitcoin/blob/master/NBitcoin/WalletPolicies/ParsingContext.cs#L47) and [Bitcoinerlab](https://github.com/bitcoinerlab/miniscript/blob/main/src/compiler/parse.ts#L214) trim the miniscript expressions while parsing.

Btccore [rejects if any character is left after parsing the miniscript](https://github.com/bitcoin/bitcoin/blob/master/src/script/miniscript.h#L2207). And rust-miniscript probably rejects at [checksum](https://github.com/rust-bitcoin/rust-miniscript/blob/master/src/descriptor/checksum.rs#L94-L102)

KY-U:
> > > > I'm getting some weird results when running locally. Is this expected?
> > > > ```
> > > > Miniscript parse failed for 0
> > > > 
> > > > Module: BitcoinerlabMiniscript
> > > > Result: 1
> > > > Module: Bitcoin
> > > > Result: 0
> > > > ```
> > > 
> > > 
> > > That's not expected. Shouldn't that case be skipped by [this logic](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/driver.cpp#L155-L158)?
> > > Actually, it may be related to the fact that bitcoinerlab's module converts `str.c_str()` to `JS_NewString()`. Currently investigating it.
> > 
> > 
> > The input has spaces and \n characteres.
> 
> Right, thats's exactly it. Bitcoin Core and Rust Miniscript handle the miniscript inputs "as is", while both [Nbitcoin](https://github.com/MetacoSA/NBitcoin/blob/master/NBitcoin/WalletPolicies/ParsingContext.cs#L47) and [Bitcoinerlab](https://github.com/bitcoinerlab/miniscript/blob/main/src/compiler/parse.ts#L214) trim the miniscript expressions while parsing.
> 
> Btccore [rejects if any character is left after parsing the miniscript](https://github.com/bitcoin/bitcoin/blob/master/src/script/miniscript.h#L2207). And rust-miniscript probably rejects at [checksum](https://github.com/rust-bitcoin/rust-miniscript/blob/master/src/descriptor/checksum.rs#L94-L102)

Considered trimming inputs for invalid characters such as `\n` and empty spaces, but it would put a lot of load into the fuzzing process. Instead I've applied a filter to simply skip inputs with invalid characters since they are outside the shared domain of the modules.

KY-U:
It is worth noting I did find other issues regarding compatibility with other miniscript modules, however, since they are actuall diversions in how the different modules implement their miniscript parser, I believe addressing them is outside the scope of this PR.

erickcestari (changes requested):
Ideally, we would pass the entire byte array to the `bitcoinerlabminiscript` function and then cast it to a javascript string. This would enable us to test more inputs and avoid missing some interesting cases by skipping a lot of them. I think these changes should work well:

[changes.txt](https://github.com/user-attachments/files/26875334/changes.txt)

Btw, bitcoinerlabminiscript miss a lot of sane checks, per example:
```
Miniscript parse failed for pkh(~)
Module: BitcoinerlabMiniscript
Result: 1
Module: Bitcoin
Result: 0
```

erickcestari:
Looks like the code format check is failing. I've forgotten to check it before creating the diff. Also looks like the build is failing now.

KY-U:
> Looks like the code format check is failing. I've forgotten to check it before creating the diff. Also looks like the build is failing now.

Yes, already looking into it. There is also a case `jjjjjj:pk(8)` from corpora that is failing. I'm trying to get that case to work and also looking into your example `pk(~)` and why it fails.

Thanks for the detailed suggestions, btw!

KY-U:
With regards to `pk(~)`, even after the adding of padding so it doesn't fail due to the key being composed of one char, since `0~` isn't a valid hex, core's modules rejects it at [`ParseHex()`](https://github.com/KY-U/bitcoinfuzz/blob/feat/add-bitcoinerlab-miniscript-module/modules/bitcoin/module.cpp#L226). It's a problem related to the specifics of core's harness, and I'm currently trying to think of the best way to address it.

KY-U:
Increased quickjs's stack size to avoid exception errors when parsing deeply nested inputs such as [this one](https://github.com/bitcoinfuzz/corpora/blob/main/miniscript_parse/aa38c5c58d1b1863eea4fe9a442c1c8ea8a748ad).

KY-U:
Documented bitcoinerlabs and nbitcoin's permissiveness behaviour in`RUNNING.md`, recommending the user to pair btccore with rust miniscript and bitcoinerlabs miniscript with nbitcoin when fuzzing for miniscript parsing due to their similar permissiveness level.

KY-U:
Fixed check-docker-image workflow error.

modules/bitcoinerlabminiscript/module.cpp
```diff
@@ -0,0 +1,89 @@
...
+  if (js_runtime == nullptr) {
+    js_runtime = JS_NewRuntime();
+    if (js_runtime == nullptr) {
+      return false;
+    }
+  }
+
+  // Deeply nested miniscript expressions can result in stack-overflow
+  // exceptions, so it's best to increase QuickJs's default stack size of 1MB
+  JS_SetMaxStackSize(js_runtime, 4 * 1024 * 1024);
```
erickcestari: I'm not sure we should do it. If a Miniscript input can trigger a stack overflow, that likely means the implementation uses a recursive parser, so we should probably skip those cases.

This is a common issue across different Miniscript implementations. Many parsers are written recursively, which can exhaust the call stack because each function call allocates additional stack space, eventually causing stack-overflow errors. How is the exception thrown in this case? Can it be caught and handled from the JavaScript side?

cc @brunoerg

erickcestari: More context here: https://github.com/rust-bitcoin/rust-miniscript/issues/696

KY-U: > I'm not sure we should do it. If a Miniscript input can trigger a stack overflow, that likely means the implementation uses a recursive parser, so we should probably skip those cases.
> 
> This is a common issue across different Miniscript implementations. Many parsers are written recursively, which can exhaust the call stack because each function call allocates additional stack space, eventually causing stack-overflow errors. How is the exception thrown in this case? Can it be caught and handled from the JavaScript side?
> 
> cc @brunoerg

The problem isn't with bitcoinerlab's library, but with the Javascript engine, i.e., Quicjk Js. I've done the following tests to verify that:
1. Added an error log to the parser function:
```typescript
function bitcoinerlabMiniscriptMiniscriptParse(input: ArrayBuffer): boolean {
   "..."
    } catch (error) {
        console.log("Erro:", error);
        return false;
    }
}
```
2. Created a custom `.ts` file where I manually ran that input through the parser function. Then transpiled the file to javascript and ran it using *node*:
```
./node_modules/.bin/esbuild src/test.ts --bundle --platform=node --format=cjs --outfile=dist/test.js

node dist/test.js
```
Resulting in no error, the miniscript expression is validated as true, likewise btccore module.
```
Result: true
```
3. Then, ran the testfile with *quickjs*:
```
qjs dist/test.js

Erro: InternalError: stack overflow
    at <anonymous> (dist/test.js:536:43)
    at map (native)
    at splitArgs (dist/test.js:536:25)
    at parseBase (dist/test.js:574:29)
    at parseExpression (dist/test.js:687:27)
    at parseBase (dist/test.js:645:49)
    at parseExpression (dist/test.js:687:27)
    at parseBase (dist/test.js:645:86)
    at parseExpression (dist/test.js:687:27)
    [...]
    at analyzeMiniscript2 (dist/test.js:2108:48)
    at bitcoinerlabMiniscriptMiniscriptParse (dist/test.js:2409:62)
    at <eval> (dist/test.js:2424:51)
Result: false
```
And we get a stack overflow.

4. Running with quickjs, but increasing the stack size, we get no errors:
```
qjs --stack-size 4M dist/test.js
Result: true
```

KY-U: However, bitcoinerlab's miniscript parser is in fact written recursively. You can see that by how `parseExpression` calls [`parseBase`](https://github.com/bitcoinerlab/miniscript/blob/main/src/compiler/parse.ts#L234), [which itself calls `parseExpression` as it "runs down" the AST](https://github.com/bitcoinerlab/miniscript/blob/main/src/compiler/parse.ts#L133) everytime it finds a subexpression.

So should we skip those big corpora cases to avoid out-of-memory errors in the CI?

KY-U:
The current failing CI check is related to a mismatch in the `psbt_parse` target. It is unrelated to the changes made in this PR.

modules/bitcoinerlabminiscript/module.cpp
erickcestari: Yes, I think we should skip the input when this exception is thrown.
```
Erro: InternalError: stack overflow
    at <anonymous> (dist/test.js:536:43)
    at map (native)
    at splitArgs (dist/test.js:536:25)
    at parseBase (dist/test.js:574:29)
    at parseExpression (dist/test.js:687:27)
    at parseBase (dist/test.js:645:49)
    at parseExpression (dist/test.js:687:27)
    at parseBase (dist/test.js:645:86)
    at parseExpression (dist/test.js:687:27)
    [...]
    at analyzeMiniscript2 (dist/test.js:2108:48)
    at bitcoinerlabMiniscriptMiniscriptParse (dist/test.js:2409:62)
    at <eval> (dist/test.js:2424:51)
Result: false
```

KY-U:
- Removed the previous increase in quickjs's stack size.
- If a stack exception occurs, the error is now captured and the module returns `std::nullopt`, which makes `driver.cpp` skip that input.
    - For robustness, while not strictly necessary, I've added stack overflow error identification both in the js/ts layer as well as in the `module.cpp` layer.
- Added a strict 32kb input size limit to the CI, since higher values where utilizing more memory than the CI limit (2048 Mb).

KY-U:
The Docker Image Check CI failure seems to be related to a connection error to the ubuntu mirror:
```
#14 97.24 W: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/noble-updates/InRelease  Could not connect to archive.ubuntu.com:80 (185.125.190.83), connection timed out
``

modules/bitcoinerlabminiscript/module.cpp
KY-U: Done!

KY-U: > I'm not sure we should do it. If a Miniscript input can trigger a stack overflow, that likely means the implementation uses a recursive parser, so we should probably skip those cases.
> 
> This is a common issue across different Miniscript implementations. Many parsers are written recursively, which can exhaust the call stack because each function call allocates additional stack space, eventually causing stack-overflow errors. How is the exception thrown in this case? Can it be caught and handled from the JavaScript side?
> 
> cc @brunoerg

Stack size limit removed!

erickcestari (approved):
Nice work!!

ACK d7a03b2bd41fd32da0862394819921ffef5e9d57

PR #484, opened 2026-03-17
Ignore workflows when changing documentation

KY-U:
- Standardized an ignore rule in all workflows when all the chages are documentation-only.
- Removed `LICENSE` from `docker-image-check.yml` ignore rule since it realistically makes no practical difference.


brunoerg:
Concept ACK

brunoerg (approved):
ACK 48c5dec869a67cf1c7182c2c9a22454390bb7d01

CI failure is unrelated.

PR #485, opened 2026-03-19
bitcoinj: Refactor module.cpp to use shared call_wrapper_method

kuliq23:
- Using a shared `call_wrapper_method` will ease adding future targets and reduce boiler plate code.
- Added `bool jvm_initialized` to guard against partial JVM/class/method initialization

### DIFF
--- a/modules/bitcoinj/module.cpp
+++ b/modules/bitcoinj/module.cpp
@@ -12,24 +12,31 @@
 namespace fs = std::filesystem;
 
 static JavaVM *jvm = nullptr;
+static bool jvm_initialized = false;
 static jclass bitcoinJWrapperClass = nullptr;
 static jmethodID createMasterKeyMethod = nullptr;
 static jmethodID deserializeExtendedKeyMethod = nullptr;
 
 static bool init_jvm() {
-  if (jvm != nullptr) {
+  if (jvm_initialized) {
     return true;
   }
 
   jvm = JvmLoader::get_jvm();
+  if (!jvm) {
+    return false;
+  }
+
   JNIEnv *env = nullptr;
   jint ge = jvm->GetEnv((void **)&env, JNI_VERSION_1_8);
+  if (ge != JNI_OK || !env) {
+    return false;
+  }
 
   bitcoinJWrapperClass = env->FindClass("wrapper/Wrapper");
   if (!bitcoinJWrapperClass) {
     return false;
   }
-
   bitcoinJWrapperClass =
       static_cast<jclass>(env->NewGlobalRef(bitcoinJWrapperClass));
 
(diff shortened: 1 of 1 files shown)


modules/bitcoinj/module.cpp
```diff
@@ -136,13 +108,15 @@ bitcoinj_bip32_deserialize_extended_key(std::span<const uint8_t> buffer) {
 namespace bitcoinfuzz {
 namespace module {
 BitcoinJ::BitcoinJ(void) : BaseModule("BitcoinJ") {}
+
 std::optional<std::string>
 BitcoinJ::bip32_master_keygen(std::span<const uint8_t> buffer) const {
-  return bitcoinj_bip32_master_keygen(buffer);
+  return call_wrapper_method(createMasterKeyMethod, buffer);
```
brunoerg, line 114: I'm not sure about this, but also not super familitar to JVM so I'd like to hear your opinion. The motivation for this change is basically to facilitate adding new targets, right? However, as far as I can see, `call_wrapper_method` is called with the _method_ and the _buffer_. But there are some targets that might send more stuff than the buffer directly, how would it work?

kuliq23: Hi and sorry for late reply.

Yes,  the intention was so we can plug in additional targets without duplicating glue code.

I didn't really consider targets that need additional parameters. That's a good point.

We could keep it and rename it to e.g. `call_wrapper_method_buffer_only` to make the limitation explicit and introduce additional wrappers later if needed in more complex targets. 

That said, I realize this isn't solution ideal and if you think this abstraction isn’t worth introducing at this stage, I’m also fine with dropping it for now.

Thanks

quapka: I like the idea of simplifying/DRYing out the module. To strike a middle ground, you could have setup/teardown functions.

The setup [here](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/modules/bitcoinj/module.cpp#L54-L58) and [here](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/modules/bitcoinj/module.cpp#L94C1-L101C4) and teardown [here](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/modules/bitcoinj/module.cpp#L70-L88) and [here](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/modules/bitcoinj/module.cpp#L115-L133)? Possibly also, a helper function for the byte arrays [here](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/modules/bitcoinj/module.cpp#L60-L65) and [here](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/modules/bitcoinj/module.cpp#L103-L109).

The DRYing out could help with the possible copy-paste mistakes - some might already be sneaking up there, since the return value of the `GetEnv` call is checked only [here](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/modules/bitcoinj/module.cpp#L99-L101), but not in [this](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/modules/bitcoinj/module.cpp#L26) or [this](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/modules/bitcoinj/module.cpp#L58) case.

Btw., and here I am just thinking aloud, couldn't the [`JNIEnv`](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/modules/bitcoinj/module.cpp#L25-L26) be got only once and then passed around?

With these changes, I'd imagine the targets to look something like this (in pseudo code):
```
if not setup() fail;
jBuffer = getBuffer(..);
jResult = _call target method_;
tearDown();
```

kuliq23: This looks good. Less boilerplate, but still flexible enough. (It will also help avoid inconsistencies - like the missing checks you mentioned:)

About `JNIEnv` - from what I gathered, it's thread-local, so each thread needs to get its own instance...

brunoerg (approved):
code review ACK ea843424e4f8151c925226e414cc036ae0c5c1be

PR #487, opened 2026-03-24
Remove byte order inversion when serializing scriptpubkey

KY-U:
Bitcoinkernel's `kernel_transaction` target was inverting the byte order of the `scriptpubkey` field. Since the filed is just a serialized bytecode, it's order should not be inverted. The convention of inverting internal byte order in visual representation is mainly for `txid` and `block hashes`.

### DIFF
--- a/modules/bitcoinkernel/module.cpp
+++ b/modules/bitcoinkernel/module.cpp
@@ -12,8 +12,20 @@ namespace {
 std::string bytes_to_hex(const uint8_t *bytes, int length) {
   std::stringstream string_stream;
   string_stream << std::hex;
+  for (int i = 0; i < length; ++i) {
+    string_stream << std::setw(2) << std::setfill('0')
+                  << static_cast<int>(bytes[i]);
+  }
+
+  return string_stream.str();
+}
+
+std::string hash_bytes_to_hex(const uint8_t *bytes, int length) {
+  std::stringstream string_stream;
+  string_stream << std::hex;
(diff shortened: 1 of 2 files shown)


modules/bitcoinkernelvariant/module.cpp
```diff
@@ -10,7 +10,17 @@
 #include <string>
 
 namespace {
-std::string bytes_to_hex(const uint8_t *bytes, int length) {
+std::string bytes_to_hex(const uint8_t *bytes, int lenght) {
+  std::stringstream string_stream;
+  string_stream << std::hex;
+  for (int i = 0; i < lenght; ++i) {
```
brunoerg, line 16:
```suggestion
  for (int i = 0; i < length; ++i) {
```

modules/bitcoinkernel/module.cpp
```diff
@@ -10,6 +10,16 @@
 
 namespace {
 std::string bytes_to_hex(const uint8_t *bytes, int length) {
+  std::stringstream string_stream;
+  string_stream << std::hex;
+  for (int i = 0; i < length; ++i) {
+    string_stream << std::setw(2) << std::setfill('0') << (int)bytes[i];
```
brunoerg: nit:
```suggestion
    string_stream << std::setw(2) << std::setfill('0') << static_cast<int>(bytes[i]);
```

brunoerg (approved):
ACK 31092a9994b9437e897adb0edade56404211599c

PR #495, opened 2026-03-27
feat: add bip32_derive_from_path target for rust-bitcoin and Bitcoin Core

kuliq23:
Based on the long and fruitful discussion in [PR423](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/423), here is the `bip32_derive_from_path` target (for _rust-bitcoin_ & _Bitcoin Core_ so far).

- Syntax checks are skipped in `module.cpp` as discussed.
- The same master private is used every time (to make sure we are really testing the derivation only), but this can be adjusted if we feel it's better to e.g. generate a new one for each run.
- Target successfully explores deep (>255) paths as intended.

### DIFF
--- a/modules/bitcoin/module.cpp
+++ b/modules/bitcoin/module.cpp
@@ -22,8 +22,10 @@ const TranslateFn G_TRANSLATION_FUN{nullptr};
 #include "script/interpreter.h"
 #include "script/miniscript.h"
 #include "script/script.h"
+#include "secp256k1.h"
 #include "span.h"
 #include "streams.h"
+#include "util/bip32.h"
 #include "util/chaintype.h"
 #include "validation.h"
 
@@ -660,6 +662,44 @@ Bitcoin::bip32_deserialize_extended_key(std::span<const uint8_t> buffer) const {
   } catch (...) {
     return "INVALID";
   }
+}
+
+std::optional<std::string>
+Bitcoin::bip32_derive_from_path(std::span<const uint8_t> buffer) const {
+  std::string path_str(reinterpret_cast<const char *>(buffer.data()),
+                       buffer.size());
+
+  // Reject hardened notation ('h')
+  if (path_str.find('h') != std::string::npos) {
+    return std::nullopt;
+  }
+
+  // Parse derivation path
+  std::vector<uint32_t> path;
+  if (!ParseHDKeypath(path_str, path) || path.empty()) {
+    return "INVALID";
+  }
+
+  static ECC_Context ecc_context;
+  SelectParams(ChainType::MAIN);
+
+  constexpr std::array<uint8_t, 32> seed_raw{
+      0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0a, 0x0b,
+      0x0c, 0x0d, 0x0e, 0x0f, 0x10, 0x11, 0x12, 0x13, 0x14, 0x15, 0x16,
+      0x17, 0x18, 0x19, 0x1a, 0x1b, 0x1c, 0x1d, 0x1e, 0x1f, 0x20};
+
+  CExtKey key;
+  key.SetSeed(std::as_bytes(std::span(seed_raw)));
+
+  for (uint32_t child : path) {
+    CExtKey next;
+    if (!key.Derive(next, child)) {
+      return "INVALID";
+    }
+    key = next;
+  }
+
+  return EncodeExtKey(key);
 }
 
 } // namespace module
--- a/modules/rustbitcoin/module.cpp
+++ b/modules/rustbitcoin/module.cpp
@@ -1,3 +1,4 @@
+#include <algorithm>
 #include <span>
 
 #include "module.h"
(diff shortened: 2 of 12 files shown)


kuliq23:
@erickcestari What do you think? 

Do you advise any design changes (like putting the filtering into a separate function) or shall I continue adding modules in this manner?
Thanks (Sorry for @mention, I cannot assign you as reviewer myself:)

modules/bitcoin/module.cpp
```diff
@@ -644,5 +646,45 @@ Bitcoin::bip32_deserialize_extended_key(std::span<const uint8_t> buffer) const {
...
+  for (uint32_t child : path) {
+    CExtKey next;
+    if (!current.Derive(next, child)) {
+      return "INVALID";
+    }
+    current = next;
+  }
+
+  return EncodeExtKey(current);
+}
```
erickcestari, line 703: nit: Just a bit of refactor using more idiomatic c++

```suggestion
std::optional<std::string>
Bitcoin::bip32_derive_from_path(std::span<const uint8_t> buffer) const {
  const std::string path_str(reinterpret_cast<const char *>(buffer.data()),
                             buffer.size());

  // Reject hardened notation ('h')
  if (path_str.find('h') != std::string::npos) {
    return std::nullopt;
  }

  // Parse derivation path
  std::vector<uint32_t> path;
  if (!ParseHDKeypath(path_str, path) || path.empty()) {
    return "INVALID";
  }

  static ECC_Context ecc_context;
  SelectParams(ChainType::MAIN);

  constexpr std::array<uint8_t, 32> seed_raw{
      0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0a, 0x0b,
      0x0c, 0x0d, 0x0e, 0x0f, 0x10, 0x11, 0x12, 0x13, 0x14, 0x15, 0x16,
      0x17, 0x18, 0x19, 0x1a, 0x1b, 0x1c, 0x1d, 0x1e, 0x1f, 0x20};

  CExtKey key;
  key.SetSeed(std::as_bytes(std::span(seed_raw)));

  for (uint32_t child : path) {
    CExtKey next;
    if (!key.Derive(next, child)) {
      return "INVALID";
    }
    key = next;
  }

  return EncodeExtKey(key);
}
```

modules/rustbitcoin/module.cpp
```diff
@@ -98,5 +99,48 @@ std::optional<std::string> Rustbitcoin::bip32_deserialize_extended_key(
   free_c_string(result_ptr);
   return result;
 }
+std::optional<std::string>
+Rustbitcoin::bip32_derive_from_path(std::span<const uint8_t> buffer) const {
+  // PATH FILTERING BEGINS
```
erickcestari: I think we should add these skips inside the `lib.rs`. We usually we don't skip in the `module.cpp`, only for C/C++ code.

erickcestari (changes requested):
I think this is the way to go! I would only change the skips place for modules that aren't written in C/C++ since they have their own file, and we could skip there.

kuliq23:
Ok, we can do that. Is there a particular reason for that though?

The idea was that skips in `module.cpp` in a single language would make it more organized and unified.

erickcestari:
> Ok, we can do that. Is there a particular reason for that though?
> 
> The idea was that skips in `module.cpp` in a single language would make it more organized and unified.

Yes, it’s a matter of preference, and both approaches can work. My reasoning is based on consistency with other modules like LDK, LND, and BTCD. In those cases, the skips are handled in the module implementation code rather than in module.cpp.

Although those skips are typically triggered by error conditions rather than input format, it still seems more coherent to keep all logic within the module implementation itself in the native language. That way, module.cpp, which serves just as a wrapper layer to be able to call functions using FFI, doesn’t end up splitting logic between itself and the real module implementation. There will be only one place to look and understand the code behavior.

@brunoerg do you have any preference?

brunoerg:
> @brunoerg do you have any preference?

My preference is keeping consistency with other modules and move the skips to the `lib.rs`. It looks better.

kuliq23:
OK. Agreed, I will continue this way. Thank you for the quick answers.

modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
```diff
@@ -438,3 +440,87 @@ pub unsafe extern "C" fn rust_bitcoin_bip32_deserialize_extended_key(
...
+        0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0a, 0x0b, 0x0c, 0x0d, 0x0e, 0x0f,
+        0x10, 0x11, 0x12, 0x13, 0x14, 0x15, 0x16, 0x17, 0x18, 0x19, 0x1a, 0x1b, 0x1c, 0x1d, 0x1e,
+        0x1f, 0x20,
+    ];
+
+    let master_key = Xpriv::new_master(NetworkKind::Main, &seed);
+    if path.as_ref().len() > 255 {
+        println!("Path length {} exceeds maximum of 255", path.as_ref().len());
+        println!("Path: {}", path);
+    }
```
erickcestari: nit: these prints can be removed
```suggestion
```

erickcestari (changes requested):
After removing the print debug it should be good to go!

modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
kuliq23: Oops... These slipped through, removed. TY

erickcestari (approved):
Thanks! Nice work!

ACK b373bc9fca5d130d6b20bcace98e3d12500cbf69

quapka:
I suppose this will require a rebase and update in the driver.cpp, due to https://github.com/bitcoinfuzz/bitcoinfuzz/pull/520 already being merged.

erickcestari (approved):
Thanks!

ACK 283d5364b8f51130e8822119e7269c7da4aa7b2c

Issue #489, opened 2026-03-25
Add pycoin module


quapka:
https://github.com/richardkiss/pycoin

brunoerg:
Please, when suggesting a new target, you could tell why this target worth. We want to avoid targets that are no longer maintained or looks abandoned as well as not used in practice.

quapka:
My bad, I went by the number of GitHub stars and recent commits. But looking at commits more closely, the last non-Copilot one is from Dec 2024. And the mentioned pycoinnet is stale for the last 5 years.

PR #496, opened 2026-04-01
feat: add Pycoin module

kuliq23:
- Added a popular Python library `Pycoin`
- https://github.com/richardkiss/pycoin
- So far for `bip32_master_keygen` target only as proof of concept.
---
Notice in [`module.cpp L40`](https://github.com/bitcoinfuzz/bitcoinfuzz/compare/v2...kuliq23:add-pycoin?expand=1#diff-61e4bbf19489d0562766a77f510779d7beab7aab84a468ec4b449b393e407832R40) I used `pycoin_main` as the main `.py` file to which the fuzz entrypoints are copied to. Remaining Python modules copy their entrypoints into `main.py`, leading to an overwrite of `main.py` if more than 1 Python modules are built. I wonder if we should use similar naming approach for both `embit` and `pybitcoinkernel` to avoid such issues. Let me know and I'll open a PR:)


brunoerg (approved):
ACK 579adc5d67e64df7a258b3e6819044127dd9928a

Built and ran successfully, thank you.

quapka:
As mentioned in https://github.com/bitcoinfuzz/bitcoinfuzz/issues/489, pycoin appears to be no longer actively maintained. This [search](https://github.com/search?q=lang%3Apython++AND+path%3A*.py+AND+%28%22import+pycoin%22+OR+%22from+pycoin%22%29&type=code) returns quite several results showing that it's used across Github, and based on [pypistats](https://pypistats.org/packages/pycoin), it's not completely dead. Not saying these links prove something, but it's OK for now to include pycoin in Bitcoinfuzz. It's criticality score is `0.44259`.

PR #501, opened 2026-04-01
feat(bitcoin): add context and test data initializer helpers

KY-U:
Implemented TODOs related to encapsulating miniscript targets initializing logic.

I've implemented it in the most simple and "easy to read" way. For more strict multi-thread correctness, [`std::call_once`](https://en.cppreference.com/w/cpp/thread/call_once.html) could be utilized, but since libfuzzer's `fork` feature creates multiple processes, and not multiple threads in the same process, It would probably be overengineering.

### DIFF
--- a/modules/bitcoin/module.cpp
+++ b/modules/bitcoin/module.cpp
@@ -306,10 +306,27 @@ Bitcoin::verify_script(const std::vector<uint8_t> &script_sig,
                       FuzzedSignatureChecker(), nullptr);
 }
 
-std::optional<bool> Bitcoin::descriptor_parse(std::string str) const {
-  // TODO: Move it to a constructor
+namespace {
+void EnsureECCContextInitialized() {
+  static bool ecc_context_initialized = false;
   static ECC_Context ecc_context{};
-  SelectParams(ChainType::MAIN);
+  if (!ecc_context_initialized) {
+    SelectParams(ChainType::MAIN);
+    ecc_context_initialized = true;
+  }
+}
+
+void EnsureTestDataInitialized() {
+  static bool test_data_initialized = false;
+  if (!test_data_initialized) {
+    TEST_DATA.Init();
+    test_data_initialized = true;
+  }
+}
+} // namespace
+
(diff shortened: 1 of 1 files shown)


modules/bitcoin/module.cpp
```diff
@@ -306,10 +306,26 @@ Bitcoin::verify_script(const std::vector<uint8_t> &script_sig,
                       FuzzedSignatureChecker(), nullptr);
 }
 
+namespace {
+void EnsureECCContextInitialized() {
+  static bool ecc_context_initialized = false;
```
brunoerg, line 311: Why do we need this `ecc_context_initialized` flag?  `static ECC_Context ecc_context{}` inside the if block is only constructed on first entry, but since it's static, it will never be destroyed or re-constructed, right?

modules/bitcoin/module.cpp
```diff
@@ -306,10 +306,26 @@ Bitcoin::verify_script(const std::vector<uint8_t> &script_sig,
...
+  if (!ecc_context_initialized) {
+    static ECC_Context ecc_context{};
+    SelectParams(ChainType::MAIN);
+    ecc_context_initialized = true;
+  }
+}
+void EnsureTestDataInitialized() {
```
brunoerg, line 319: nit: it's missing blank line before `EnsureTestDataInitialized`.

modules/bitcoin/module.cpp
KY-U: The `ecc_context_initialized` flag is there to guarantee that `SelectParams(ChainType::MAIN);` is called once, since there is no reason to globally set the same chain on every iteration.

`static ECC_Context ecc_context{}` is not strictly necessary to be inside the if block, you're right. It won't change the behaviour. I've put it inside the if to make it explicit that it will only be called upon initialization, but I'll put it outside to avoid the possible interpretation that it is necessary to be there.

brunoerg (approved):
code review ACK 793328186d64d3d73b6d93a53cae9a686bca1779

PR #503, opened 2026-04-02
fix(bitcoin): use CheckBlock in deserialize_block

brunoerg:
Aligns deserialize_block with btcd's BTCDDesBlock, which calls CheckBlockSanity (coinbase checks, per-tx sanity) and ValidateWitnessCommitment. CheckBlock with fCheckPOW=false covers the former; IsBlockMutated(block, true) covers the latter.

### DIFF
--- a/modules/bitcoin/module.cpp
+++ b/modules/bitcoin/module.cpp
@@ -352,10 +352,16 @@ Bitcoin::deserialize_block(std::span<const uint8_t> buffer) const {
   } catch (const std::ios_base::failure &) {
     return std::nullopt;
   }
-  if (block.vtx.empty()) {
+  static bool initialized = false;
+  if (!initialized) {
+    SelectParams(ChainType::MAIN);
+    initialized = true;
+  }
+  BlockValidationState state;
+  if (!CheckBlock(block, state, Params().GetConsensus(), /*fCheckPOW=*/false)) {
     return "0";
   }
-  if (IsBlockMutated(block, true)) {
+  if (IsBlockMutated(block, /*check_witness_root=*/true)) {
     return "0";
   }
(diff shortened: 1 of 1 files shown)


modules/bitcoin/module.cpp
```diff
@@ -352,10 +352,12 @@ Bitcoin::deserialize_block(std::span<const uint8_t> buffer) const {
   } catch (const std::ios_base::failure &) {
     return std::nullopt;
   }
-  if (block.vtx.empty()) {
+  SelectParams(ChainType::MAIN);
```
KY-U: `SelectParams(ChainType::MAIN);` only needs to be executed once per process, right? Maybe it's call could be substituted by a simple helper function that instantiates it only once, similar to [501](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/501):
```c++
void EnsureChainTypeSelected() {
  static bool chain_type_selected = false;
  if (!chain_type_selected) {
    SelectParams(ChainType::MAIN);;
    chain_type_initialized = true;
  }
}
...

EnsureChainTypeSelected();
BlockValidationState state;
...
```
Or an inline lambda function call to a static flag variable.
```c++
static const bool chaintype_selected = [] {
  SelectParams(ChainType::MAIN);
  return true;
}();
BlockValidationState state;
...
```

brunoerg: Oh, thanks, I forgot to do it!

brunoerg:
Force-pushed resolving https://github.com/bitcoinfuzz/bitcoinfuzz/pull/503#discussion_r3029732205.

brunoerg:
Left it running overnight, looks good, will merge for now.

PR #504, opened 2026-04-03
custommutator: add extended_key property tests and Makefile target

morningstarxcdcode:
Adds competency-test scope for extended_key custom mutator: property checks for max_size, base58check decode to 78 bytes, and 0-byte input safety. Also adds testextendedkey target and runs it from make test.

### DIFF
--- a/custommutator/Makefile
+++ b/custommutator/Makefile
@@ -13,7 +13,7 @@ MAKE = make
 AR := ar
 
 # Automatically find all .cpp files
-BITCOIN_SRCS := $(shell find . -name "*.cpp" ! -name "bech32.cpp" ! -name "test_onion_main.cpp" ! -path "./univalue/*" ! -path "./secp256k1/*")
+BITCOIN_SRCS := $(shell find . -name "*.cpp" ! -name "bech32.cpp" ! -path "./tests/*" ! -path "./univalue/*" ! -path "./secp256k1/*")
 BITCOIN_OBJS := $(BITCOIN_SRCS:.cpp=.o)
 
 # Build libsecp256k1 with sanitizer flags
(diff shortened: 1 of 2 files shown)


morningstarxcdcode:
First, I cleaned the build and rebuilt the testextendedkey target from scratch to make sure the binary compiles correctly.

image



After that, I ran ./testextendedkey and confirmed it passes all required checks: every output stays within max_size, every output decodes to exactly 78 bytes, and zero-byte input completes safely without crashing.

image



Next, I ran make test to confirm the full test workflow works end to end. That run executes test_onion first and then testextendedkey, and both pass.

image



Finally, I checked that the new test file exists and that the Makefile includes the testextendedkey target, which confirms the project is wired up correctly.

image

morningstarxcdcode:
I completed and verified the extended-key competency test. I rebuilt the target cleanly, ran the dedicated test binary, and confirmed it passes all three required behaviors. I also verified the combined make test target runs both test_onion and testextendedkey, and I confirmed the new test file and Makefile target are present.

morningstarxcdcode:
I approached this as a property-based test rather than a one-off check, focusing on the key invariants the mutator needs to satisfy (max_size bounds, valid 78-byte decoding, and safe handling of edge-case inputs).
The harness is deterministic using fixed seeds, so results are reproducible and easier to debug. While testing, I also caught and fixed a buffer sizing issue in the harness by ensuring the allocation accounts for both input_size and max_size.
The test is intentionally minimal and self-contained, and the same structure should scale cleanly to the other mutators in Idea 3 in sob 2026 ...

morningstarxcdcode:
Let me know if you'd like me to extend this further. I’m planning to apply the same property-based approach to the remaining mutators (e.g. bolt11/bolt12), and also start building out a mock-based driver harness to systematically test the Driver targets and their cross-module consistency.

morningstarxcdcode:
The goal is to make the test structure consistent across both mutators and driver logic so we can validate correctness and invariants end-to-end.

morningstarxcdcode:
This mirrors the kind of differential correctness guarantees the driver enforces, so testing both layers together should improve overall reliability.

brunoerg:
NACK. Do not open PR for competency test.

morningstarxcdcode:
> NACK. Do not open PR for competency test.

But in sob in your org idea it is written submit a small pr with the proposal
@brunoerg

PR #506, opened 2026-04-03
fix: init secp256k1 submodule for custom mutators

wolgwang1729:
### Description

This PR addresses local build failures for targets utilizing custom mutators and introduces a dynamic smoke testing script to validate the local Docker setup for all fuzz targets.

**1. Fix `auto_build.py` Submodule Initialization**

Certain fuzz targets were failing to build locally via Docker, specifically during the `auto_build.py` execution phase. The build would terminate with the following error:

```

failed to solve: process "/bin/sh -c export JAVA_HOME=/usr/lib/jvm/java-21-openjdk-$(dpkg --print-architecture) &&     /build/auto_build.py" did not complete successfully: exit code: 1

error: Recipe `run` failed on line 15 with exit code 17

```

This occurred because the `CUSTOM_MUTATOR_*` flags (e.g., `CUSTOM_MUTATOR_ONION` or `CUSTOM_MUTATOR_P2P_MESSAGE`) were not being recognized by `auto_build.py` as triggers to initialize the `external/secp256k1` submodule.

* **Change:** Updated `ensure_submodules_for` in `auto_build.py` to append and deduplicate the `secp256k1` path for any flag starting with the `CUSTOM_MUTATOR_` prefix. This systemic fix ensures all current and future mutator-based targets initialize their dependencies correctly.

### Motivation

A user(@mowgli_2002) recently reported hitting the `secp256k1` error on Discord when trying to run `decode_onion` on `v2`. I wrote the smoke test to verify repository health and immediately caught the exact same regression on `parse_p2p_message`. Combining the test and the fix ensures no custom mutator targets silently break local development environments going forward.

### Testing Done

- [x] Verified `parse_p2p_message` and `decode_onion` now successfully build, start, and run until the timeout without submodule errors.

### DIFF
--- a/auto_build.py
+++ b/auto_build.py
@@ -75,9 +75,17 @@ SUBMODULES_BY_FLAG = {
 
 def ensure_submodules_for_flags(flags, quiet: bool):
     paths = []
+    has_custom_mutator = False
     for flag in flags:
         paths += SUBMODULES_BY_FLAG.get(flag, [])
-
+        if flag.startswith("CUSTOM_MUTATOR_"):
+            has_custom_mutator = True
+            
+    if has_custom_mutator:
+        paths.append("external/secp256k1")
+            
+    # Deduplicate paths and preserve order
+    paths = list(dict.fromkeys(paths))
     if not paths:
         return
     if not quiet:


brunoerg:
Can you please rebase it? I think it solves the issue on CI.

wolgwang1729:
@brunoerg I have rebased onto the latest v2 and squashed the commits. The CI shows "All checks have passed
31 successful checks"

brunoerg:
While I'm not sure about this dynamic smoke test, I think we can move faster if you limit the scope of this PR to just fix the secp256k1 submodule init.

wolgwang1729:
@brunoerg I've removed the smoke test and restricted the scope of this PR entirely to the `auto_build.py` submodule fix.

wolgwang1729:
I've updated the PR to resolve the merge conflicts caused by recent changes in `auto_build.py`.

brunoerg:
Please, squash your commits.

auto_build.py
```diff
@@ -77,7 +77,11 @@ def ensure_submodules_for_flags(flags, quiet: bool):
     paths = []
     for flag in flags:
         paths += SUBMODULES_BY_FLAG.get(flag, [])
-
+        if flag.startswith("CUSTOM_MUTATOR_"):
```
brunoerg, line 81: When we have more than one `CUSTOM_MUTATOR_`, shouldn't this append "external/secp256k1" just one time?

wolgwang1729: Yes, it should append just one time. I am de duplicating in the end. If it's inefficient, I can check after all the flags are added and append it once after the loop.

wolgwang1729: Done.

brunoerg (approved):
code review ACK 28ac87aba94cdce692623cebe08adcc220bee81f

PR #507, opened 2026-04-03
feat: Add pywallet

### LINKED ISSUE #491: Add pywallet module
https://github.com/ranaroussi/pywallet

Omswastik-11:
closes #491


modules/bitcoinj/Makefile
```diff
@@ -21,6 +21,7 @@ lib: $(BUILD_DIR)/app.zip
 	cp -r bitcoinj_extracted/*/lib/* lib/
 
 $(BUILD_DIR)/app.zip:
+	chmod +x ./gradlew
```
quapka: Changes to `bitcoinj` should not be added as part of this PR.

.github/workflows/workflow.yml
```diff
@@ -73,6 +75,7 @@ jobs:
           pip install -r modules/embit/requirements.txt
           pip install -r modules/pybitcoinkernel/requirements.txt
           pip install -r modules/pycoin/requirements.txt
+          pip install -r modules/pywallet/requirements.txt
```
kuliq23:
```suggestion
          pip install -r --no-deps modules/pywallet/requirements.txt
```
Add `--no-deps` to avoid dependency problems (as you did in Dockerfile#L73)

This file is probably the reason the CI pipeline fails in your fork.
Continued below to mirror your setup workaround.

.github/workflows/workflow.yml
```diff
@@ -73,6 +75,7 @@ jobs:
...
           pip install mako
```
kuliq23, line 81:
```suggestion
- name: Patch pywallet
  run: |
    python -c "from pathlib import Path; import importlib.util; spec = importlib.util.find_spec('pywallet.utils'); Path(spec.origin).write_text('from .bip32 import Wallet\n', encoding='utf-8')"
    
```

Mirror the setup to be consistent with Dockerfile#74.

modules/pywallet/requirements.txt
```diff
@@ -0,0 +1,4 @@
+pywallet==0.1.0
+base58>=2.1.1
+ecdsa>=0.19.0
+six>=1.17.0
```
kuliq23, line 4:
```suggestion
```
I did not need this dependency to run the fuzzing. Please double check if this line is really necessary.

auto_build.py
```diff
@@ -15,6 +15,9 @@
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from pathlib import Path
 import platform
+import threading
+
+SUBMODULE_UPDATE_LOCK = threading.Lock()
```
kuliq23:
```suggestion
```
I'd suggest opening a separate PR for this type of change, where it can be further discussed. This clear separation of issues helps with orientation and backwards code inspection. I suggest sticking to the target only for now.

auto_build.py
```diff
@@ -91,8 +94,9 @@ def ensure_submodules_for(flag: str, quiet: bool):
         return
     if not quiet:
         print(f"Ensuring submodules for {flag}: {', '.join(paths)}")
-    for path in paths:
-        run(f"git submodule update --init --recursive {path}", quiet=quiet)
+    with SUBMODULE_UPDATE_LOCK:
```
kuliq23:
```suggestion
```
I'd suggest opening a separate PR for this type of change, where it can be further discussed. This clear separation of issues helps with orientation and backwards code inspection. I suggest sticking to the target only for now.

Dockerfile
```diff
@@ -62,13 +62,16 @@ ENV PATH="/venv/bin:$PATH" \
...
+COPY modules/pywallet/requirements.txt /tmp/requirements-pywallet.txt
 RUN --mount=type=cache,target=/root/.cache/pip,id=fuzz-pip \
     python3 -m venv /venv && \
     python3 -m ensurepip && \
     python3 -m pip install --upgrade pip && \
     python3 -m pip install mako setuptools && \
-    python3 -m pip install -r /tmp/requirements.txt
+    python3 -m pip install -r /tmp/requirements-embit.txt && \
+    python3 -m pip install --no-deps -r /tmp/requirements-pywallet.txt && \
+    python3 -c "from pathlib import Path; import importlib.util; spec = importlib.util.find_spec('pywallet.utils'); Path(spec.origin).write_text('from .bip32 import Wallet\\n', encoding='utf-8')"
```
kuliq23, line 74: This is a simple workaround around the `two1` module  & `Python3.12+` incompatibility issue.
It allows the `master_keygen` target to run with the `pywallet` module successfully.

However, this approach modifies installed package code after installation, which can affect reproducibility and may hide potential issues. If `pywallet.utils` is used elsewhere, this patch will change the behavior.

Either way, we should very explicitly document this patch.

I'd like to hear your opinions on this, @brunoerg , @erickcestari, @quapka.

Dockerfile
```diff
@@ -124,6 +127,11 @@ COPY --from=builder \
     /build/modules/*/lib /
 COPY --from=builder /build/*.so .
 
+# Keep Python dependencies installed in the builder venv available at runtime
+# for embedded Python modules (pywallet, pycoin, embit, etc).
+COPY --from=builder /venv /venv
+ENV PYTHONPATH=/venv/lib/python3.12/site-packages
+
```
kuliq23:
```suggestion
```
This idea is nice and would probably improve performance. However, I'd suggest to open a separate PR for this change.

README.md
```diff
@@ -268,6 +268,22 @@ If you prefer, you can still build the modules manually. Below are the steps for
...
 
+- ### [pywallet](https://github.com/ranaroussi/pywallet)
+
+    To run the fuzzer with `pywallet` module, you need to install the `pywallet` library.
+
+    To install the `pywallet` library, you can use the following command:
+    ```bash
+    cd modules/pywallet
+    pip install --no-deps -r ./requirements.txt
+    ```
```
kuliq23, line 280:
```suggestion
    To install the `pywallet` library, you can use the following command:
    ```bash
    cd modules/pywallet
    pip install --no-deps -r ./requirements.txt
    python3 -c "from pathlib import Path; import importlib.util; spec = importlib.util.find_spec('pywallet.utils'); Path(spec.origin).write_text('from .bip32 import Wallet\\n', encoding='utf-8')"
    ```
```
Add the `pywallet.utils` patch to avoid dependency errors in consistency with Dockerfile#L74.

kuliq23:
Nice work!
Thank you for working on this.

I’ve left a few comments in the code with some suggested changes.

The main concern I see is the [workaround](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/507/changes#r3035625094) around the `two1` dependency. I know dependency issues can be frustrating and finding a clean fix is hard but hopefully we will find an acceptable solution through discussion.

The steps you can take now is going through the recommended changes and improving your code to get the GitHub CI/CD pipeline to run successfully on your fork.

Thank you again and don't be afraid to reach out!

Dockerfile
quapka: I worry a bit that the issue stems from the fact that `pywallet` does not seem to be actively developed - aka, I'd expect the upstream to fix the root cause. This goes to me; it might not be a good target.

Anyways, overriding the sources might be okayish in the Dockerfile, but it's a worse solution for the local installation, where the library could be installed in an environment shared across multiple projects.

I am questioning the addition of this module altogether. I've pinged the original maintainer to see if there are any plans to continue.

modules/pywallet/requirements.txt
Omswastik-11: I re-tested in an isolated virtualenv: without six, pywallet import fails through ecdsa (ModuleNotFoundError: six). So I kept six because it is required for this no-deps installation path.

Omswastik-11:
Hi @kuliq23 , 

I've updated the PR to address the Docker stage copies and reverted the unrelated changes in auto_build.py and `bitcoinj/Makefile` as requested.

Regarding the `pywallet` dependency, there is a significant hurdle: the upstream `pywallet` package (which is largely unmaintained) has a hard dependency on `two1`, which is broken on Python 3.12+ (it relies on the removed `distutils` and `pipes` modules). This causes any standard `pip install pywallet` to fail.

To get around this while keeping the Docker image slim, I am currently using this approach:
1. `pip install pywallet --no-deps` along with its working sub-dependencies (`ecdsa`, `base58`, `six`).
2. A surgical source-patch via `python -c` that removes the `import two1` line inside `pywallet/utils/__init__.py` (which isn't actually needed for the functions we use).

**Could you let me know if you are comfortable with this patching approach in the Dockerfile and CI?** 

Alternatively:
- We could vendor the specific parts of `pywallet` we need into `modules/pywallet/utils/`, but that might be more bloat in the repo.
- We could try to find a more modern fork, though most appear to be equally outdated.

What would you prefer here? or any suggestion what we can do ?

brunoerg:
Thanks for your contibution but since pywallet has not been actively developed, we don't think worth having it here.

PR #508, opened 2026-04-03
Add pywallet module

### LINKED ISSUE #491: Add pywallet module

codemaster1104:
closes #491


modules/pywallet/pywallet_lib.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class _Two1Finder(importlib.abc.MetaPathFinder):
+    def find_spec(self, name, path, target=None):
+        if name == "two1" or name.startswith("two1."):
+            return importlib.machinery.ModuleSpec(name, _Two1Loader())
+        return None
+
+
+sys.meta_path.insert(0, _Two1Finder())
```
quapka, line 34: What is all of this supposed to be? It looks some kind of backdoor attempt.

codemaster1104: Hey this was done to resolve the two1 dependency issue for pywallet, as the dockerfile encouraged python 3.12 but  a plain pip install pywallet breaks on the project's Ubuntu 24 environment.
So with this when something tries to import two1, give it this dummy empty module instead of going to disk.

I get the concern , what could be a better solution?

codemaster1104: and two1 was generally used for eth utilities, and in our usecase it didnt affect the fuctionality that's why i left it this way initially

brunoerg:
This is a duplicate of #507. Also, the commit organization and PR description are terrible, if you don't take the care to organize commits and write a proper PR description, I don't think it deserves our review time.

Issue #334, opened 2025-11-20
Add bitcoin-s module


brunoerg:
https://bitcoin-s.org/

Divyansh200102:
Can I work on this are you a maintainer of this project? @brunoerg

brunoerg:
> Can I work on this are you a maintainer of this project? @brunoerg 

You can work on it, no need to ask for permission.

codemaster1104:
Hey, was working on it I think it's implementation would be similar to bitcoin j , will make a pr shortly

codemaster1104:
hey , initially planning to make a PR for only master keygen target and will add other targets with different PR's once its reviewed

PR #514, opened 2026-04-05
add bitcoin-s

guptapratykshh:
- Adds a new `modules/bitcoins/` module wrapping the [bitcoin-s](https://github.com/bitcoin-s/bitcoin-s) library (v1.9.9, Scala/JVM)
- Implements three fuzzing targets: `bip32_master_keygen`, `bip32_deserialize_extended_key`, and `psbt_parse`
- Wires the module into all five integration points: root `Makefile`, `helpers/jvmloader.cpp`, `module_defs.h`, `module_loader.h`, and the module's own `Makefile`

PR #515, opened 2026-04-06
Add bitcoin-s module for bip32_master_keygen

codemaster1104:
work on #334


modules/bitcoins/app/src/main/scala/wrapper/Wrapper.scala
```diff
@@ -0,0 +1,26 @@
...
+
+  def createMasterKey(seed: Array[Byte]): String = {
+    try {
+      val xprv: ExtPrivateKey = ExtPrivateKey(
+        ExtKeyVersion.LegacyMainNetPriv,
+        Some(ByteVector(seed)),
+        BIP32Path.empty)
+      xprv.toString
+    } catch {
+      case _: Exception => "skip error"
```
kuliq23:
```suggestion
      case _: Exception => ""
```
We return `""` (empty string) when the key generation fails. Skipping is only used in cases where a specific library enforces e.g. seed length limitations, to avoid crashes.

I recommend running the target locally first, so we're sure it works as intended.

brunoerg:
Please, squash your commits.

modules/bitcoins/app/src/main/scala/wrapper/Wrapper.scala
codemaster1104: Hey, i think i have corrected this now and have also added length checks for skip error

brunoerg:
This a duplication of #514, I will close it for now. Please, help us to review that PR since you got experience here. Just in case, if #514 gets stuck or inactive, we can reopen this one again. Thanks.

PR #516, opened 2026-04-07
fix(stump_modified_add): skip all 0's input

KY-U:
While `utreexo`'s Go implementation considers a hash consisting of all 0 bytes as empty, `rustreexo` has no such restriction, considering it a valid hash, which means that an all 0's input is outside the shared domain of both implementations and should be skipped.

Added a lambda function in `driver.cpp`'s `stump_modified_add` target to skip the case where a hash is composed of all zeros.

### DIFF
--- a/driver.cpp
+++ b/driver.cpp
@@ -1,3 +1,4 @@
+#include <algorithm>
 #include <cassert>
 #include <fuzzer/FuzzedDataProvider.h>
 #include <iostream>
@@ -855,6 +856,13 @@ void Driver::StumpModifyAddTarget(std::span<const uint8_t> buffer) const {
     auto hash = provider.ConsumeBytes<uint8_t>(32);
     if (hash.size() < 32) {
       break;
+    }
+    // Skip all-zero hashes because Utreexod's dynamic accumulator
+    // implementation treats them as a special case (empty tree), but other
+    // implementations may handle that differently.
+    if (std::all_of(hash.begin(), hash.end(),
+                    [](uint8_t byte) { return byte == 0; })) {
+      continue;
     }
     add_hashes.push_back(std::move(hash));
   }


driver.cpp
```diff
@@ -856,6 +857,10 @@ void Driver::StumpModifyAddTarget(std::span<const uint8_t> buffer) const {
     if (hash.size() < 32) {
       break;
     }
+    if (std::all_of(hash.begin(), hash.end(),
```
brunoerg, line 863: Please, add a comment saying this is a workaround because there is a bug in one of the implementations.

KY-U:
Force pushed solving https://github.com/bitcoinfuzz/bitcoinfuzz/pull/516#discussion_r3045133853.

brunoerg (changes requested):
> Force pushed solving [#516 (comment)](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/516#discussion_r3045133853).

I don't see the comment in the code.

KY-U:
> > Force pushed solving [#516 (comment)](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/516#discussion_r3045133853).
> 
> I don't see the comment in the code.

Forgot to save the change before amending, skill issue.

driver.cpp
```diff
@@ -856,6 +857,12 @@ void Driver::StumpModifyAddTarget(std::span<const uint8_t> buffer) const {
     if (hash.size() < 32) {
       break;
     }
+    // Skip all-zero hashes because Utreexo treats them as a special case (empty
```
brunoerg: Utreexo or Utreexod implementation?

```suggestion
    // Skip all-zero hashes because Utreexod treats them as a special case (empty
```

KY-U: Utreexod is the full node implementation. The  dynamic acumulator repo is indeed called [utreexo](https://github.com/utreexo/utreexo). But it probably isn't clear for a user that "utreexo" is refering to the specific go implementation. I'll change that.

KY-U: Changed comment to better specify reference to utreexod's specific accumulator implementation.

brunoerg (approved):
code review ACK 3290a29ca58a4d513bf38c9e916dbee3293cc5f8

PR #517, opened 2026-04-07
fix(just): remove legacy docker-compose requirement

mahmoudnasser1561:
This removes the `docker-compose` binary requirement from the `justfile` and aligns the setup docs with the commands the repo already uses.

The `just` recipes invoke `docker compose`, but `required_bins` still checked for the standalone `docker-compose` executable. This could block users with a working Compose v2 plugin setup.

for verification I did the following :
- ran `just list-targets`
- verified `docker compose` config parsing still works

This is a small consistency fix that removes an unnecessary prerequisite and makes the documented setup match the actual `just` workflow.

### DIFF
--- a/RUNNING.md
+++ b/RUNNING.md
@@ -2,18 +2,15 @@
 
 ## Environment Preparation
 
-Install Docker, Docker Compose, jq, and just before interacting with this repository:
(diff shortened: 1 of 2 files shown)


erickcestari (approved):
Thanks! Nice catch!

ACK 197898166e3aba19a818e1e8f395ce45934034fa

PR #520, opened 2026-04-07
refactor(driver): deduplicate repeated response verification logic

mahmoudnasser1561:
This refactors `driver.cpp` by extracting the repeated response verification pattern used across multiple targets into a small helper.

#### Why this is useful

A large part of `driver.cpp` repeated the same verification block across many targets. That made the file noisier to maintain and increased the chance of mismatch-reporting logic drifting over time.

This change reduces that duplication while keeping target-specific parsing and custom mismatch handling untouched.

This is a cleanup-only change and is not intended to change fuzzing behavior.

### DIFF
--- a/driver.cpp
+++ b/driver.cpp
@@ -2,11 +2,34 @@
 #include <fuzzer/FuzzedDataProvider.h>
 #include <iostream>
 #include <string>
+#include <string_view>
 #include <unistd.h>
 
 #include "driver.h"
 #include <bitcoinfuzz/basemodule.h>
 #include <bitcoinfuzz/module_registry.h>
+
+namespace {
+template <typename T>
+void VerifyMatchingResponse(std::optional<T> &last_response,
+                            std::string &last_module_name,
+                            const std::string &module_name, const T &response,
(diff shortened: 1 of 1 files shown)


KY-U:
You can run `make format` from root to fix the format check CI error

brunoerg:
Concept ACK

KY-U:
Tested a few targets locally, did not run into any issue, lgtm.

erickcestari (changes requested):
Concept ACK

There are more targets that can be refactored to use the new function:

  - `DescriptorParseTarget`                                                                                                                                     
  - `MiniscriptParseTarget`                                                                                                                                    
  - `InvoiceDeserializationTarget`                                                                                                                                 
  - `OfferDeserializationTarget`                                                                                                                               
  - `TransactionEvalTarget`

erickcestari (approved):
Thanks!

ACK f92d7b49cb8534789abbb4732e85d272978c62a1

brunoerg (approved):
code review ACK f92d7b49cb8534789abbb4732e85d272978c62a1

Thanks.

Issue #479, opened 2026-03-11
musig2 differential fuzzing


luisschwab:
Add targets for MuSig2 libraries:

- [ ] [`storopoli/musig2`](https://github.com/storopoli/musig2) (Haskell λ)
- [ ] [`conduition/musig2`](https://github.com/conduition/musig2) (Rust)
- [ ] [`cmdruid/musig2`](https://github.com/cmdruid/musig2) (TypeScript)
- [ ] [`bted/btcec`](https://github.com/bitweb-project/bted/tree/master/btcec) (Go)

Vedd-Patel:
hi! i've recently set up the bitcoinfuzz Docker environment locally and would love to help add MuSig2 targets!
to keep the PR focused, could i claim just the Go (`bted/btcec`) or Rust (`conduition/musig2`) target to start with?
would love any guidance over the same!!

Justxd22:
WIP [`conduition/musig2`](https://github.com/conduition/musig2) Rust: #522

brunoerg:
We have a rule of only integrating libraries that are actively maintained. e.g. https://github.com/cmdruid/musig2 last commit was 2 years ago.

PR #522, opened 2026-04-08
feat: add MuSig2 key aggregation differential fuzzing target (#479)

Justxd22:
Impl first stage of MuSig2 differential fuzzing #479, i added `musig2_key_agg` target and compares the results to `libsecp256k1`.

- Added **`modules/rustmusig2`** wrapping `conduition/rust-musig2` via FFI.
- Added `musig2_key_agg` target to `BaseModule` and `Driver`.
- Updated **`modules/secp256k1`** to include MuSig2 support.
- Output format: **33-byte hex string** (32-byte X-only pubkey + 1-byte Y-parity) to ensure full mathematical consistency across implementations.

Im planning a follow-up PR to implement the remaining MuSig2 protocol stages (Nonce Aggregation, Partial Signing, and Signature Aggregation) + add `custommutator/` to generate valid curve points

#### Test with 
```bash
CXX=clang++ CXXFLAGS="-DRUST_MUSIG2 -DSECP256K1" ./auto_build.py
```
 then 
```bash
FUZZ=musig2_key_agg ./bitcoinfuzz
```


brunoerg (changes requested):
Please integrate it in CI. Otherwise, running the CI for this PR will have no effect.

Justxd22:
@brunoerg CI will need a corpus in bitcoinfuzz/corpora - I can open a PR there with seed inputs, or what do you prefer?

brunoerg:
> @brunoerg CI will need a corpus in bitcoinfuzz/corpora - I can open a PR there with seed inputs, or what do you prefer?

We don't need a corpus to verify that the modules successfully builds - which is very important. I appreciate if you open a PR on corpora to add inputs for this target but we will only merge that after this.

Justxd22:
@brunoerg ready

brunoerg:
Needs rebase.

Justxd22:
@brunoerg ready

modules/secp256k1/module.cpp
```diff
@@ -214,6 +215,59 @@ secp256k1_schnorr_verify(std::span<const uint8_t> privkey,
   return "VALID";
 }
 
+std::optional<std::string>
+secp256k1_musig2_key_agg(std::span<const uint8_t> buffer) {
+  // Input format: [1 byte: num_keys] [33 bytes * num_keys: pubkeys]
+  if (buffer.size() < 1)
+    return std::nullopt;
+
+  size_t num_keys = buffer[0];
```
brunoerg, line 224: I don't think num_keys should be `buffer[0]`. It's not a good practice to use the whole buffer and let the modules do the stuff. Instead, `Musig2KeyAggTarget` could have more parameters (from driver.cpp) and you can use the `FuzzedDataProvider` to define them.

brunoerg (changes requested):
Concept ACK but Approach NACK - please take a look at https://github.com/bitcoinfuzz/bitcoinfuzz/pull/522/changes#r3161612593, it's a blocker to me.

brunoerg:
Closing due to lack of activity

PR #523, opened 2026-04-08
add addrman_serialize differential fuzz target

guptapratykshh:
- Rust addrman binary parser (modules/bitcoinaddrman/)
- All C++ integration points wired: basemodule.h/cpp, driver.h/cpp, module_defs.h, module_loader.h, docker-compose.yml, auto_build.py


brunoerg:
Thanks but NACK - We don't have any interest on this harness anymore.

PR #524, opened 2026-04-09
add property tests for extendedkey custom mutator

guptapratykshh:
adds property based tests for the extended_key custom mutator covering three invariants: output always fits within `max_size`, output `base58check-decodes` to exactly 78 bytes (`BIP32 payload`), and a 0-byte input never crashes.

### DIFF
--- a/custommutator/Makefile
+++ b/custommutator/Makefile
@@ -13,7 +13,7 @@ MAKE = make
 AR := ar
 
 # Automatically find all .cpp files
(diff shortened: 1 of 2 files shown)


kuliq23:
See [PR504](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/504)

guptapratykshh:
ok, I think the SOB portal states a PR regarding that project is needed for the extendedkey custom mutator. Any updates on how we will submit the competency test for that project?

brunoerg:
No need to open PR, can attach your solution in your proposal.

PR #526, opened 2026-04-09
auto_build.py: accept -D flags in CLEAN_BUILD

Shreyas2004wagh:
## Summary

Normalize custom `CLEAN_BUILD` values before passing them to `clean_by_flags()`.

This makes the script accept both of these forms:

- `CLEAN_BUILD="LDK BTCD"`
- `CLEAN_BUILD="-DLDK -DBTCD"`

## Why

The README currently documents the `-D...` form for selective clean:

```bash
CLEAN_BUILD="-DLDK -DBTCD" CXXFLAGS="-DLDK -DLND" ./auto_build.py
```

But in the custom clean path, `auto_build.py` was splitting raw tokens and using them directly as flags, which produces invalid module paths for values like `-DLDK`.

## Change

- add a small parser for custom `CLEAN_BUILD` values
- strip the `-D` prefix when present
- keep plain flag names working as before

## Verification

- `python -m py_compile auto_build.py`
- checked that both `"-DLDK -DBTCD"` and `"LDK BTCD"` normalize to `["LDK", "BTCD"]`

### DIFF
--- a/auto_build.py
+++ b/auto_build.py
@@ -76,6 +76,15 @@ def should_build_sequentially(flag: str) -> bool:
 def get_flags(cxxflags: str):
     return re.findall(r"-D([A-Z0-9_]+)", cxxflags)
 
+def parse_clean_flags(clean_build: str):
+    flags = []
+    for token in clean_build.split():
+        if token.startswith("-D"):
+            token = token[2:]
+        if token:
+            flags.append(token)
+    return flags
+
 # Maps module flags to required git submodule paths defined in .gitmodules.
 SUBMODULES_BY_FLAG = {
     "CLIGHTNING": ["external/lightning"],
@@ -142,7 +151,7 @@ def main():
         elif clean_build == "CLEAN":
             clean_by_flags(get_flags(cxxflags))
         else:
-            clean_by_flags(clean_build.split())
+            clean_by_flags(parse_clean_flags(clean_build))
     else:
         print("No CLEAN_BUILD option specified. Skipping clean step.")
 


auto_build.py
```diff
@@ -76,6 +76,15 @@ def should_build_sequentially(flag: str) -> bool:
...
     return re.findall(r"-D([A-Z0-9_]+)", cxxflags)
 
+def parse_clean_flags(clean_build: str):
+    flags = []
+    for token in clean_build.split():
+        if token.startswith("-D"):
+            token = token[2:]
+        if token:
+            flags.append(token)
+    return flags
```
KY-U, line 86: You could reutilize the already existing parsing logic from `get_flags()`, for example:

```python
def parse_clean_flags(clean_build: str):
    flags = []
    for token in clean_build.split():
        parsed_flag = get_flags(token)
        if parsed:
            flags.extend(parsed_flag)
        else:
            flags.append(token)
    return flags
   ```
It  wouldn't make practical difference tho, its functional the way it is.
   
 Or, if we were to not allow mixing flags with plain module names in the `CLEAN_BUILD` variable:
 ```python
def parse_clean_flags(clean_build: str):
    flags = get_flags(clean_build)
    if flags:
        return flags
    return [token for token in clean_build.split()]
 ```

KY-U:
Tested and it's working. The only nit is I'm not sure if it makes sense to allow mixing compiler flags `-DFLAG` and plain module names `FLAG` in the same environment variable, e.g., `CLEAN_BUILD="-DLDK BTCD"`, since they are conceptually different.

Shreyas2004wagh:
@KY-U ,that makes sense. I kept the parser permissive so both the documented `-D...` form and plain module names work, including mixed input, but I agree that mixed values are a bit conceptually odd.Will tighten it if you'd prefer `CLEAN_BUILD` to accept only one style at a time.

KY-U:
> @KY-U ,that makes sense. I kept the parser permissive so both the documented `-D...` form and plain module names work, including mixed input, but I agree that mixed values are a bit conceptually odd.Will tighten it if you'd prefer `CLEAN_BUILD` to accept only one style at a time.

It's working the way it is, so it isn't strictly necessary. Do you think it makes sense to enforce not mixing compiler flags with module names in  the flag @brunoerg?

brunoerg:
Concept NACK

Thanks for noticing but if "The README currently documents the -D... form for selective clean" so the documentation is wrong. `-D` form should not be part of this script, it's a build/c++ thing.

PR #527, opened 2026-04-09
main: reject missing FUZZ target during init

Shreyas2004wagh:
## Summary

Fail fast when `FUZZ` is missing or empty during fuzzer initialization.

## Why

`bitcoinfuzz` requires `FUZZ` to select the target to run. Previously, `main.cpp` read it with `std::getenv("FUZZ")` and stored the raw pointer directly. If `FUZZ` was unset, that left a null value flowing into target dispatch.

## Change

- store the target as a `std::string`
- validate that `FUZZ` is present and non-empty in `LLVMFuzzerInitialize()`
- print a clear error and exit early when it is missing

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -2,12 +2,17 @@
 
 static std::shared_ptr<bitcoinfuzz::Driver> driver = nullptr;
 static bitcoinfuzz::ModuleLogger module_logger;
-static const char *g_target = nullptr;
+static std::string g_target;
 
 extern "C" int LLVMFuzzerInitialize([[maybe_unused]] int *argc,
                                     [[maybe_unused]] char ***argv) {
   bitcoinfuzz::InitializeRegistry();
-  g_target = std::getenv("FUZZ");
+  const char *target = std::getenv("FUZZ");
+  if (target == nullptr || target[0] == '\0') {
+    std::cerr << "FUZZ environment variable must be set" << std::endl;
+    std::exit(1);
+  }
+  g_target = target;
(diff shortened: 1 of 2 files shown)


main.cpp
```diff
@@ -1,13 +1,22 @@
 #include <bitcoinfuzz/module_loader.h>
 
+#include <cstdlib>
+#include <iostream>
+#include <string>
+
```
erickcestari: We don't needs these imports, since we already importing it on `module_loader.h`

```suggestion
```

erickcestari (changes requested):
Concept ACK

Just some nits:

We can change the message error from `driver.cpp` since now it will only detect unknowns target

https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/driver.cpp#L954

```cpp
std::cerr << "Unknown target: " << target << std::endl;
```

Shreyas2004wagh:
Addressed, thanks.

> Concept ACK
> 
> Just some nits:
> 
> We can change the message error from `driver.cpp` since now it will only detect unknowns target
> 
> https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/driver.cpp#L954
> 
> ```c++
> std::cerr << "Unknown target: " << target << std::endl;
> ```

erickcestari (changes requested):
Could you squash the commits to be only one?

erickcestari (approved):
Thanks!

ACK e58e2451ec54cef748740ffb387498ff3fc93cbf

PR #532, opened 2026-04-12
adding cgroup support

codemaster1104:
This was listed as one of the TODO items in codebase 
This implementation assumes cgroup v2 and has a fallback to nproc

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -141,7 +141,7 @@ COPY --from=builder --parents \
 # from the website https://llvm.org/docs/LibFuzzer.html#options
 # mkdir to init and make sure we have write permissions
 #
-# TODO: Fix the use of nproc to actually use cgroup based "cpus"
+
 ENTRYPOINT mkdir -p $FUZZ_DATADIR/crash \
     $FUZZ_DATADIR/corpus \
     && exec /app/bitcoinfuzz \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -161,7 +161,7 @@ ENTRYPOINT mkdir -p $FUZZ_DATADIR/crash \
     $( [ -n "${LIBFUZZ_MINIMIZE_CRASH}" ] && echo "-minimize_crash=${LIBFUZZ_MINIMIZE_CRASH}" ) \
     -reload=${LIBFUZZ_RELOAD:-1} \
     -jobs=${LIBFUZZ_JOBS:-0} \
-    -fork=${LIBFUZZ_FORKS:-$(nproc)} \
+    -fork=${LIBFUZZ_FORKS:-$(awk 'NR==1 {print int($1/$2)}' /sys/fs/cgroup/cpu.max 2>/dev/null || nproc)} \
```
KY-U: Nit: extra empty spaces after `\`

codemaster1104: cool :'( , correcting it

KY-U:
You can remove the extraspaces in the first commit, rebase the branch and then force push to avoid an extra fix commit.

Dockerfile
```diff
@@ -161,7 +161,7 @@ ENTRYPOINT mkdir -p $FUZZ_DATADIR/crash \
...
```
KY-U: In the current docker image, the value of the first column (cpu quota) of `/sys/fs/cgroup/cpu.max ` is `"max"` which is converted to 0 in `int($1/$2)`. And since the `awk` command doesn't fail, it doesn't default to `nproc`, resulting in `LIBFUZZ_FORKS=0`. You can force the default to `nproc` in the case of `int($1/$2) == 0`  by forcing 1(error) as the `stdout` return:
```suggestion
    -fork=${LIBFUZZ_FORKS:-$(awk 'NR==1 { n=int($1/$2); if (n < 1) exit 1; print n }' /sys/fs/cgroup/cpu.max 2>/dev/null || nproc)} \
```

codemaster1104: Thanks for the insight

brunoerg (approved):
ACK 1dddbeab9a78541cd1441cf03f58f71cff5ce70e

PR #533, opened 2026-04-13
driver: scope NBitcoin PSBT workaround and avoid early target exit

rocketpowerkille:
This PR fixes a control-flow issue in `Driver::PSBTParseTarget` (`driver.cpp`).

Before this change, when NBitcoin-specific PSBT output contained `in=0` or `out=0`, the code used `return;` inside the module loop. That exited the whole target early and skipped comparisons for remaining loaded modules.

Changes:
- scope the workaround to NBitcoin module output (`module.first == "NBitcoin"`).
- replace `return` with `continue` so only that module result is skipped while other module comparisons continue.

How this might help:
early target exit can hide downstream mismatches, reducing differential fuzzing effectiveness by giving false negatives.

Validation (WSL/Linux):
- A/B check with old logic reproduced failure:
  `Assertion 'reference->psbt_calls == 1' failed.`
- with this fix:
  `driver psbt nbitcoin regression tests: PASS`

attached `test_driver_psbt_nbitcoin_regression.cpp` to lock this behavior.
[test_driver_psbt_nbitcoin_regression.txt](https://github.com/user-attachments/files/26665711/test_driver_psbt_nbitcoin_regression.txt)

The test uses a minimal `CountingPSBTModule` mock and checks:
1. NBitcoin `in=0` output is skipped **without** aborting comparisons for later modules.
2. Non-NBitcoin modules returning `in=0` are **not** filtered by the NBitcoin workaround.

This catches the old `return` behavior (loop abort) and validates the new `continue` behavior.

### DIFF
--- a/driver.cpp
+++ b/driver.cpp
@@ -73,8 +73,10 @@ void Driver::ScriptEvalTarget(std::span<const uint8_t> buffer) const {
 #endif
 
 #ifdef NBITCOIN
-  if (std::ranges::find(input_data, 0xB2) != input_data.end())
-    return;
+  if (ModuleRegistry::instance().isEnabled("NBITCOIN")) {
+    if (std::ranges::find(input_data, 0xB2) != input_data.end())
+      return;
+  }
 #endif
 
   auto flags = provider.ConsumeIntegral<unsigned int>();
@@ -263,10 +265,11 @@ void Driver::PSBTParseTarget(std::span<const uint8_t> buffer) const {
       continue;
 
 #ifdef NBITCOIN
-    if (ModuleRegistry::instance().isEnabled("NBITCOIN")) {
+    if (module.first == "NBitcoin" &&
+        ModuleRegistry::instance().isEnabled("NBITCOIN")) {
       if (res->find("in=0") != std::string::npos ||
           res->find("out=0") != std::string::npos)
-        return;
+        continue;
     }
 #endif
 


KY-U:
The change preserves comparisons among the remainer modules after skipping nbitcoin. However when only two modules are enabled, one of them being nbitcoin, skipping it leaves no differential comparison for that input, and no warning that it is the case. Is that tradeoff intentional? Of course, it won't matter if you're comparing >= 3 modules at the same time.

rocketpowerkille:
fair point. My change just stops the whole target from bailing out early. If only two modules are enabled and one is NBitcoin, skipping NBitcoin means there’s nothing left to compare, so that input is effectively dropped. I think that’s okay because with only one remaining module there’s no meaningful differential result anyway. If you’d prefer a warning or counter for that case I’m happy to add it, but I avoided extra fuzz log noise by default.

brunoerg (approved):
code review ACK f6caaa701e0078031ca029e45cd11a7263c8feda

wolgwang1729:
@rocketpowerkille @brunoerg I was investigating the test failures that started occurring after this was merged, and I think I've found the root cause.

The change in this PR from `return;` to `continue;` was totally correct for fixing the control-flow bug (avoiding the early target exit). However, that old `return;` statement had a side effect: it was acting as a global silencer. Because of the way the old `if (ModuleRegistry::instance().isEnabled("NBITCOIN"))` check worked, *any* module that returned `in=0` or `out=0` caused the loop to instantly abort, skipping all further differential comparisons for that fuzzer input.

By scoping the `continue` strictly to `module.first == "NBitcoin"`, this PR successfully removed that global silencer. But as a result, the fuzzer is now properly cross-checking the *other* parsed modules (like `rust-bitcoin`, `libwally`, etc.) on these 0-input/0-output PSBT edge cases. 

It turns out the pipeline is failing now because those other libraries **actually disagree** with each other on how to handle 0-input/0-output PSBTs. 

### The Dilemma
I initially wrote a [fix](https://github.com/wolgwang1729/bitcoinfuzz/commit/c8ebeccfaac065dbe0025d10da6e497e7d938ef3) to broaden the `continue;` skip so it applies to *all* modules returning `in=0` or `out=0` (which gets the tests to pass again). However, I realized doing so might go against the spirit of `bitcoinfuzz`. If two libraries disagree on a 0-input PSBT, shouldn't the fuzzer flag that so we can figure out which one is violating BIP 174?

### I am wondering
1. Should we expand the `continue;` to mask all 0-input/0-output PSBTs across all modules (treating these edge cases as known/ignored discrepancies)?
2. Or should we dig into the specific mismatches now being exposed, and see if we need targeted upstream bug reports or specific string-parsing workarounds for the diverging libraries?

brunoerg:
> I initially wrote a [fix](https://github.com/wolgwang1729/bitcoinfuzz/commit/c8ebeccfaac065dbe0025d10da6e497e7d938ef3) to broaden the continue; skip so it applies to all modules returning in=0 or out=0 (which gets the tests to pass again). However, I realized doing so might go against the spirit of bitcoinfuzz. If two libraries disagree on a 0-input PSBT, shouldn't the fuzzer flag that so we can figure out which one is violating BIP 174?

To not block the fuzzing process, we can simply skip this 0-0 PSBTs, no problem. We're aware of the discrepancy, not super relevant and we should not let this block us.

PR #534, opened 2026-04-14
add libbitcoin-system module

guptapratykshh:
- added libbitcoin-system(v4) as differential fuzzing module with 8 targets: script, transaction_eval, deserialize_block, address_parse, bip32_master_keygen, bip32_deserialize_extended_key, script_eval, and verify_script
- script consensus verification, which includes script_eval and verify_script, is implemented directly using libbitcoin-system's machine::interpreter


brunoerg (changes requested):
Please, add this new module and their targets on CI.

brunoerg (changes requested):
From CI:

```sh
CMake Warning:
  Manually-specified variables were not used by the project:

    CMAKE_CXX_COMPILER


CMake Error at CMakeLists.txt:44 (find_package):
  Could not find a configuration file for package "Boost" that is compatible
  with requested version "1.86.0".

  The following configuration files were considered but not accepted:

    /usr/lib/x86_64-linux-gnu/cmake/Boost-1.83.0/BoostConfig.cmake, version: 1.83.0
    /lib/x86_64-linux-gnu/cmake/Boost-1.83.0/BoostConfig.cmake, version: 1.83.0



make: *** [Makefile:47: ../../external/libbitcoin-system/build_fuzz/libbitcoin-system.a] Error 1
```

quapka:
When I was working with `libbitcoin-system` about a year ago, I had to use Boost version < 1.76, [see](https://github.com/libbitcoin/libbitcoin-system/blob/version3/README.md?plain=1#L94-L100) in the `v3` version of libbitcoin-system. When looking through the source code [this line](https://github.com/libbitcoin/libbitcoin-system/blob/4f211adf2a2900ead138754af2252a7620ed099e/m4/ax_boost_url.m4#L21) seems to suggest that it should build with versions > 1.74, but apparently it doesn't?

brunoerg:
> When I was working with libbitcoin-system about a year ago, I had to use Boost version < 1.76, [see](https://github.com/libbitcoin/libbitcoin-system/blob/version3/README.md?plain=1#L94-L100) in the v3 version of libbitcoin-system. When looking through the source code [this line](https://github.com/libbitcoin/libbitcoin-system/blob/4f211adf2a2900ead138754af2252a7620ed099e/m4/ax_boost_url.m4#L21) seems to suggest that it should build with versions > 1.74, but apparently it doesn't?

Perhaps it depends on which version of libbitcoin you're trying to build?

KY-U:
It might be good to update the readme and include instructions to initialize the required  submodules.

modules/libbitcoinsystem/Makefile
```diff
@@ -0,0 +1,91 @@
...
+	  -DCMAKE_BUILD_TYPE=Release \
+	  -DCMAKE_INSTALL_PREFIX=$(SECP_INSTALL) \
+	  -DCMAKE_C_FLAGS="-fsanitize=address,fuzzer-no-link" \
+	  -DSECP256K1_BUILD_TESTS=OFF \
+	  -DSECP256K1_BUILD_EXAMPLES=OFF \
+	  -DSECP256K1_ENABLE_MODULE_RECOVERY=ON \
+	  -DSECP256K1_ENABLE_MODULE_EXTRAKEYS=ON \
+	  -DSECP256K1_ENABLE_MODULE_SCHNORRSIG=ON \
+	  -DSECP256K1_ENABLE_MODULE_ECDH=ON
+	cmake --build $(LIB_DIR)/build_secp --target install --parallel
```
KY-U: The `--parallel` flag has no `job` count, so the Make build was running too many jobs for my local machine, and the OOM killer was terminating clang++ when trying to compile the module.

Adding a job count so no unbound parallelism happens can help prevent that problem. Maybe bound by `nproc`? It doesn't necessarily means it's a safe number of jobs, just a possible option.

modules/libbitcoinsystem/Makefile
```diff
@@ -0,0 +1,91 @@
...
+
+UNAME_S := $(shell uname -s)
+ifeq ($(UNAME_S), Darwin)
+  BOOST_ROOT ?= /opt/homebrew
+else
+  BOOST_ROOT ?= /usr
+endif
+
+CXX ?= c++
+
```
KY-U, line 22:
```suggestion
CLANG_WARNING_FLAGS = -Wno-deprecated-literal-operator -Wno-macro-redefined
```

When compiling the module, I was getting a lot of `macro-redefined` warnings such as
```
/home/ted/github/bitcoinfuzz/external/libbitcoin-system/builds/cmake/../../include/bitcoin/system/boost.hpp:33:9: warning: 'BOOST_CONTAINER_NO_LIB' macro
        redefined [-Wmacro-redefined]
     33 | #define BOOST_CONTAINER_NO_LIB
        |         ^
  <command line>:4:9: note: previous definition is here
      4 | #define BOOST_CONTAINER_NO_LIB 1
        |         ^
```
The first definition of `BOOST_CONTAINER_NO_LIB` is generated by CMake after linking against Boost imported targets. You can see the `-DBOOST_CONTAINER_NO_LIB` definition in `external/libbitcoin-system/build_fuzz/CMakeFiles/libbitcoin-system.dir/flags.make:12`:
```
CXX_DEFINES = -DBOOST_ATOMIC_NO_LIB -DBOOST_CHARCONV_NO_LIB -DBOOST_CHRONO_NO_LIB -DBOOST_CONTAINER_NO_LIB -DBOOST_DATE_TIME_NO_LIB -DBOOST_IOSTREAMS_NO_LIB -DBOOST_LOCALE_NO_LIB -DBOOST_PROGRAM_OPTIONS_NO_LIB -DBOOST_RANDOM_NO_LIB -DBOOST_REGEX_NO_LIB -DBOOST_THREAD_NO_LIB -DBOOST_URL_NO_LIB
```
Which is treated as `#define BOOST_CONTAINER_NO_LIB 1`.

Then, the same variable is defined as empty in `external/libbitcoin-system/include/bitcoin/system/boost.hpp`:
```
// Headers only containers.
#define BOOST_CONTAINER_NO_LIB
```
Which causes the  macro redefined warning. Adding the `-Wno-macro-redefined` flag gets rid of that.

modules/libbitcoinsystem/Makefile
```diff
@@ -0,0 +1,91 @@
...
+	@echo "Building libbitcoin-system..."
+	sed -i.bak -E "s/(find_package\( *Boost )1\.86\.0/\1$(BOOST_MIN_VERSION)/" \
+	  $(LIB_DIR)/builds/cmake/CMakeLists.txt
+	cmake -B $(LIB_DIR)/build_fuzz \
+	  -S $(abspath $(LIB_DIR)/builds/cmake) \
+	  -DCMAKE_C_COMPILER="$(CC)" \
+	  -DCMAKE_CXX_COMPILER="$(CXX)" \
+	  -DCMAKE_BUILD_TYPE=Release \
+	  -DBUILD_SHARED_LIBS=OFF \
+	  -DCMAKE_CXX_FLAGS="-fsanitize=address,fuzzer-no-link -O1" \
```
KY-U:
```suggestion
	  -DCMAKE_CXX_FLAGS="-fsanitize=address,fuzzer-no-link -O1 $(CLANG_WARNING_FLAGS)" \
```

Adding the previously defined warning flags to the libbitcoin-system build step

modules/libbitcoinsystem/Makefile
KY-U: Also, there a lot of `deprecated-literal-operator` warnings such as:
```
/home/ted/github/bitcoinfuzz/external/libbitcoin-system/builds/cmake/../../include/bitcoin/system/literals.hpp:117:28: warning: identifier '_a' preceded
        by whitespace in a literal operator declaration is deprecated [-Wdeprecated-literal-operator]
    117 | consteval auto operator "" _a() noexcept
        |                ~~~~~~~~~~~~^~
        |                operator""_a
  [...]
  /home/ted/github/bitcoinfuzz/external/libbitcoin-system/builds/cmake/../../include/bitcoin/system/literals.hpp:178:25: warning: identifier '_size'
        preceded by whitespace in a literal operator declaration is deprecated [-Wdeprecated-literal-operator]
    178 | DECLARE_INTEGER_LITERAL(_size, positive, size_t)
        |                         ^
  /home/ted/github/bitcoinfuzz/external/libbitcoin-system/builds/cmake/../../include/bitcoin/system/literals.hpp:179:25: warning: identifier '_nsize'
        preceded by whitespace in a literal operator declaration is deprecated [-Wdeprecated-literal-operator]
    179 | DECLARE_INTEGER_LITERAL(_nsize, negative, signed_size_t)
        |
```
Which can be hidden with the `-Wno-deprecated-literal-operator`

It is possible that those warnings are specific to my clang version, mine is `22.1.3`. If those warnings don't appear to you and are specific to my config, do let me know!

modules/libbitcoinsystem/Makefile
```diff
@@ -0,0 +1,91 @@
...
+UNAME_S := $(shell uname -s)
+ifeq ($(UNAME_S), Darwin)
+  BOOST_ROOT ?= /opt/homebrew
+else
+  BOOST_ROOT ?= /usr
+endif
+
+CXX ?= c++
+
+CXXFLAGS += -Wall -Wextra -std=c++20 \
```
KY-U:
```suggestion
CXXFLAGS += $(CLANG_WARNING_FLAGS) \
            -Wall -Wextra -std=c++20 \
```
Adding warnings flags to `CXXFLAGS` in order to avoid warnings during `module.o` build

KY-U:
Although the module compiles individually, currently, it isn't possible to do differential fuzzing with `btccore` module due to both modules utilizing the same `secp256k1` symbols.

When trying to compile bitcoinfuzz with both modules enabled, I get a lot of multiple definition errors due to libsecp256k1:
```
/usr/bin/ld: modules/libbitcoinsystem/module.a(secp256k1.c.o):(.bss+0x0): multiple definition of `__odr_asan_gen_secp256k1_context_static'; modules/bitcoin/module.a(libsecp256k1_la-secp256k1.o):(.bss+0x0): first defined here
/usr/bin/ld: modules/libbitcoinsystem/module.a(secp256k1.c.o):(.bss+0x1): multiple definition of `__odr_asan_gen_secp256k1_context_no_precomp'; modules/bitcoin/module.a(libsecp256k1_la-secp256k1.o):(.bss+0x1): first defined here
/usr/bin/ld: modules/libbitcoinsystem/module.a(secp256k1.c.o):(.bss+0x2): multiple definition of `__odr_asan_gen_secp256k1_nonce_function_rfc6979'; modules/bitcoin/module.a(libsecp256k1_la-secp256k1.o):(.bss+0x2): first defined here
/usr/bin/ld: modules/libbitcoinsystem/module.a(secp256k1.c.o):(.bss+0x3): multiple definition of `__odr_asan_gen_secp256k1_nonce_function_default'; modules/bitcoin/module.a(libsecp256k1_la-secp256k1.o):(.bss+0x3): first defined here
/usr/bin/ld: modules/libbitcoinsystem/module.a(secp256k1.c.o):(.bss+0x4): multiple definition of `__odr_asan_gen_secp256k1_ecdh_hash_function_sha256'; modules/bitcoin/module.a(libsecp256k1_la-secp256k1.o):(.bss+0x4): first defined here
/usr/bin/ld: modules/libbitcoinsystem/module.a(secp256k1.c.o):(.bss+0x5): multiple definition of `__odr_asan_gen_secp256k1_ecdh_hash_function_default'; modules/bitcoin/module.a(libsecp256k1_la-secp256k1.o):(.bss+0x5): first defined here
/usr/bin/ld: modules/libbitcoinsystem/module.a(secp256k1.c.o):(.bss+0x6): multiple definition of `__odr_asan_gen_secp256k1_nonce_function_bip340'; modules/bitcoin/module.a(libsecp256k1_la-secp256k1.o):(.bss+0x6): first defined here
/usr/bin/ld: modules/libbitcoinsystem/module.a(secp256k1.c.o):(.bss+0x7): multiple definition of `__odr_asan_gen_secp256k1_ellswift_xdh_hash_function_prefix'; modules/bitcoin/module.a(libsecp256k1_la-secp256k1.o):(.bss+0x7): first defined here
/usr/bin/ld: modules/libbitcoinsystem/module.a(secp256k1.c.o):(.bss+0x8): multiple definition of `__odr_asan_gen_secp256k1_ellswift_xdh_hash_function_bip324'; modules/bitcoin/module.a(libsecp256k1_la-secp256k1.o):(.bss+0x8): first defined here
```
This is a recurring problem with differential fuzzing and has been solved before by remapping conflicting symbols. One example of such logic is in the [module for differential fuzzing of libbitcoinkernel](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/modules/bitcoinkernelvariant/Makefile#L22-L54).

Using that remapping as a guide, I've added a symbol remapping logic as well as my other suggested changes to this diff file in order to address those issues.
  
[libbitcoinsystem-makefile.txt](https://github.com/user-attachments/files/27185079/libbitcoinsystem-makefile.txt)

Update: since we aren't supporting macOS anymore, you can ignore those parts of my suggestion!

README.md
```diff
@@ -260,6 +260,17 @@ If you prefer, you can still build the modules manually. Below are the steps for
     export CXXFLAGS="$CXXFLAGS -DLIBWALLY_CORE"
     ```
 
+- ### [libbitcoin-system](https://github.com/libbitcoin/libbitcoin-system)
+
+    Requires Boost (>= 1.86) and CMake. On Debian/Ubuntu: `apt-get install libboost-all-dev cmake binutils`. On macOS: `brew install boost cmake`.
```
brunoerg: You can remove this "macOS" instruction since we do not support macOS anymore.

brunoerg:
> Although the module compiles individually, currently, it isn't possible to do differential fuzzing with `btccore` module due to both modules utilizing the same `secp256k1` symbols.
> 
> When trying to compile bitcoinfuzz with both modules enabled, I get a lot of multiple definition errors due to libsecp256k1:
> 
> ```
> /usr/bin/ld: modules/libbitcoinsystem/module.a(secp256k1.c.o):(.bss+0x0): multiple definition of `__odr_asan_gen_secp256k1_context_static'; modules/bitcoin/module.a(libsecp256k1_la-secp256k1.o):(.bss+0x0): first defined here
> /usr/bin/ld: modules/libbitcoinsystem/module.a(secp256k1.c.o):(.bss+0x1): multiple definition of `__odr_asan_gen_secp256k1_context_no_precomp'; modules/bitcoin/module.a(libsecp256k1_la-secp256k1.o):(.bss+0x1): first defined here
> /usr/bin/ld: modules/libbitcoinsystem/module.a(secp256k1.c.o):(.bss+0x2): multiple definition of `__odr_asan_gen_secp256k1_nonce_function_rfc6979'; modules/bitcoin/module.a(libsecp256k1_la-secp256k1.o):(.bss+0x2): first defined here
> /usr/bin/ld: modules/libbitcoinsystem/module.a(secp256k1.c.o):(.bss+0x3): multiple definition of `__odr_asan_gen_secp256k1_nonce_function_default'; modules/bitcoin/module.a(libsecp256k1_la-secp256k1.o):(.bss+0x3): first defined here
> /usr/bin/ld: modules/libbitcoinsystem/module.a(secp256k1.c.o):(.bss+0x4): multiple definition of `__odr_asan_gen_secp256k1_ecdh_hash_function_sha256'; modules/bitcoin/module.a(libsecp256k1_la-secp256k1.o):(.bss+0x4): first defined here
> /usr/bin/ld: modules/libbitcoinsystem/module.a(secp256k1.c.o):(.bss+0x5): multiple definition of `__odr_asan_gen_secp256k1_ecdh_hash_function_default'; modules/bitcoin/module.a(libsecp256k1_la-secp256k1.o):(.bss+0x5): first defined here
> /usr/bin/ld: modules/libbitcoinsystem/module.a(secp256k1.c.o):(.bss+0x6): multiple definition of `__odr_asan_gen_secp256k1_nonce_function_bip340'; modules/bitcoin/module.a(libsecp256k1_la-secp256k1.o):(.bss+0x6): first defined here
> /usr/bin/ld: modules/libbitcoinsystem/module.a(secp256k1.c.o):(.bss+0x7): multiple definition of `__odr_asan_gen_secp256k1_ellswift_xdh_hash_function_prefix'; modules/bitcoin/module.a(libsecp256k1_la-secp256k1.o):(.bss+0x7): first defined here
> /usr/bin/ld: modules/libbitcoinsystem/module.a(secp256k1.c.o):(.bss+0x8): multiple definition of `__odr_asan_gen_secp256k1_ellswift_xdh_hash_function_bip324'; modules/bitcoin/module.a(libsecp256k1_la-secp256k1.o):(.bss+0x8): first defined here
> ```
> 
> This is a recurring problem with differential fuzzing and has been solved before by remapping conflicting symbols. One example of such logic is in the [module for differential fuzzing of libbitcoinkernel](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/modules/bitcoinkernelvariant/Makefile#L22-L54).
> 
> Using that remapping as a guide, I've added a symbol remapping logic as well as my other suggested changes to this diff file in order to address those issues.
> 
> [libbitcoinsystem-makefile.txt](https://github.com/user-attachments/files/27185079/libbitcoinsystem-makefile.txt)

@guptapratykshh I think you're going well, and adding libbitcoin here is very valuable. I think that fixing these symbol conflicts is basically what it's missing to go. Take a look on this suggestion from @KY-U.

modules/libbitcoinsystem/Makefile
```diff
@@ -0,0 +1,125 @@
...
+	# libbitcoin-system.o files) using objcopy --redefine-syms. This is what
+	# allows linking together with bitcoin core's secp256k1 without symbol
+	# collisions during differential fuzzing builds.
+	rm -rf tmp_lbs tmp_secp
+	mkdir -p tmp_lbs tmp_secp
+	cd tmp_lbs && $(AR) x $(abspath $(LIBBITCOIN_LIB))
+	cd tmp_secp && $(AR) x $(abspath $(SECP_STATIC_LIB))
+	@echo "Renaming secp256k1 symbols in libbitcoin-system and secp256k1 objects..."
+	@for obj in tmp_lbs/*.o tmp_secp/*.o; do \
+		$(OBJCOPY) --redefine-syms=$(abspath secp256k1_symbols.map) $$obj 2>/dev/null || true; \
```
KY-U: I think it would be desirable to have the build stop on a failed objcopy operation, instead of continuing even with a partially mapped object due to `|| true`.

Also, if it does fails, it would be valuable to have the diagnostic included in the ci instead of redirecting it with `2>/dev/null`.

modules/libbitcoinsystem/Makefile
```diff
@@ -0,0 +1,125 @@
...
+	@echo "Generated $$(wc -l < $@) symbol mappings"
+
+module.a: $(LIBBITCOIN_LIB) $(SECP_STATIC_LIB) module.o secp256k1_symbols.map
+	rm -f module.a
+ifeq ($(UNAME_S), Darwin)
+	libtool -static -o module.a \
+	  $(LIBBITCOIN_LIB) \
+	  $(SECP_STATIC_LIB) \
+	  module.o
+else
```
KY-U: Since mac isn't supported anymore, this section can be removed.

KY-U (commented):
The object-level remapping approach makes sense. But I think it would be good to fail loudly if remapping fails, since its a crucial part of the build process.

brunoerg:
Cool, now it's building, we just need to investigate if the targets are properly written since there are crashes for basically every single one.

brunoerg:
Since basically all the touched targets are failed, it's probably some missing stuff on libbitcoin side when adding support for these targets. I've asked claude about it and here are some concerns that I got (have not verified, please do not consider it a validated thing from me):

```
  Bug 1 — script_parse: wrong return value format + missing unspendable check

  - "000" for invalid/unparseable scripts conflicted with Bitcoin Core's "0". The cross-check in ScriptTarget would immediately assert when Bitcoin
  Core returned "0" and libbitcoin-system returned "000" for the same input.
  - Bitcoin Core's IsUnspendable() returns "0" for any script beginning with OP_RETURN (0x6a) or larger than 10 000 bytes. Without that check,
  libbitcoin-system would return "000" (0 sigops + not-witness + not-pushonly) for OP_RETURN scripts while Bitcoin Core returned "0" → assert.

  Bug 2 — address_parse: "INVALID" for bech32 addresses

  wallet::payment_address only handles legacy P2PKH/P2SH. Any bech32/segwit address (P2WPKH, P2WSH, P2TR) made it return "INVALID" while Bitcoin Core
  / rust-bitcoin / btcd return "WPKH:...", "WSH:...", "TR:..." → assert. Returning std::nullopt makes libbitcoin-system abstain for those cases
  (nullopt is skipped by the driver). It still votes on P2PKH/P2SH addresses where it has real coverage.

  Bug 3 — script_eval: wrong scriptPubKey causes inverted semantics

  Using OP_1 (0x51) as the scriptPubKey meant connect() always pushed a truthy value last, so the result was "did scriptSig execute without error?" —
  not "did scriptSig leave a truthy value on the stack?". Concrete example: OP_0 → Bitcoin Core's EvalScript returns false (empty byte vector on stack
   is falsy), but libbitcoin's connect() returned true (OP_0 runs fine, then OP_1 is pushed on top → truthy). An empty scriptPubKey makes connect()
  check whatever scriptSig actually left, matching Bitcoin Core's semantics.

  ---
  One remaining concern (not fixed, worth investigating): verify_script uses chain::flags::no_rules but doesn't have the equivalent of Bitcoin Core's
  FuzzedSignatureChecker. If libbitcoin-system still performs real cryptographic signature verification under no_rules, scripts containing
  OP_CHECKSIG/OP_CHECKMULTISIG in the scriptPubKey would diverge (libbitcoin fails on random signatures, Bitcoin Core always passes via
  FuzzedSignatureChecker). This would affect the verify_script target specifically.
```

Suggested improvements:

```diff
diff --git a/modules/libbitcoinsystem/module.cpp b/modules/libbitcoinsystem/module.cpp
index d559180c..e2d49478 100644
--- a/modules/libbitcoinsystem/module.cpp
+++ b/modules/libbitcoinsystem/module.cpp
@@ -19,10 +19,19 @@ std::optional<std::string>
 LibbitcoinSystem::script_parse(std::span<const uint8_t> buffer) const {
   try {
     data_chunk data(buffer.begin(), buffer.end());
+
+    // Match Bitcoin Core's IsUnspendable(): OP_RETURN prefix or oversize script.
+    if (!data.empty() && data[0] == 0x6a) {
+      return "0";
+    }
+    if (data.size() > 10000) {
+      return "0";
+    }
+
     chain::script script{data, false};

     if (!script.is_valid()) {
-      return "000";
+      return "0";
     }

     const auto &ops = script.ops();
@@ -102,7 +111,11 @@ LibbitcoinSystem::address_parse(std::string str) const {
     wallet::payment_address addr(str);

     if (!addr) {
-      return "INVALID";
+      // libbitcoin-system's payment_address only handles legacy P2PKH/P2SH.
+      // For bech32/segwit addresses (P2WPKH, P2WSH, P2TR) it returns an
+      // invalid addr while other modules return "WPKH:", "WSH:", "TR:" etc.
+      // Returning nullopt makes us abstain rather than falsely report "INVALID".
+      return std::nullopt;
     }

     std::string prefix;
@@ -216,8 +229,12 @@ LibbitcoinSystem::script_eval(const std::vector<uint8_t> &input_data,
       return std::nullopt;
     }

-    const data_chunk op_true_bytes{0x51};
-    const chain::script out_script{op_true_bytes, false};
+    // Use an empty scriptPubKey so connect() checks whether scriptSig leaves
+    // a truthy value on the stack — matching Bitcoin Core's EvalScript semantics.
+    // An OP_1 scriptPubKey would always push a truthy value last, making the
+    // result "did scriptSig execute without error?" instead of "did it succeed?".
+    const data_chunk empty_bytes{};
+    const chain::script out_script{empty_bytes, false};

     const chain::transaction tx{
         1u,

```

brunoerg:
We'll have to fork libbitcoin-system and mock functions (from `system/impl/machine/interpreter.ipp`) such as `verify_signature`, `op_check_locktime_verify` and `op_check_sequence_verify`, unless there is a way to mock them without having to modify the code. 

It can be done for this PR or we can move without it, and remove transaction/scripts targets from this PR and address them in a follow-up.

brunoerg:
Needs rebase

modules/libbitcoinsystem/module.cpp
```diff
@@ -0,0 +1,303 @@
...
+    data_chunk data(buffer.begin(), buffer.end());
+    chain::block blk{data, true};
+
+    // Same rationale as transaction_eval: is_valid() only checks
+    // deserialization. check() applies the consensus-level checks that match
+    // bitcoin core's CheckBlock, so we don't return a hash for blocks the
+    // reference rejects.
+    if (!blk.is_valid() || blk.check()) {
+      return std::nullopt;
+    }
```
KY-U, line 102: If the condition ```if (!blk.is_valid() || blk.check())``` is met, doesn't it mean the block is invalid? Why skip that input instead of signaling as failure with `return "0"`?

modules/libbitcoinsystem/module.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+    return encode_hash(blk.hash());
+  } catch (...) {
+    return std::nullopt;
```
KY-U, line 106: Instead of assuming every input that captured an exception should be skipped, it would be good to capture that failure to investigate the cause.

[Here](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/modules/bitcoinerlabminiscript/module.cpp#L97-L101) is an example of that rationale, if the exception is an already discussed "known case", the input is skipped, otherwise, it fails.

modules/libbitcoinsystem/module.cpp
```diff
@@ -0,0 +1,303 @@
...
+      prefix = "PKH:";
+    } else if (addr.prefix() == wallet::payment_address::mainnet_p2sh) {
+      prefix = "SH:";
+    } else {
+      prefix = "UNK:";
+    }
+
+    return prefix + addr.encoded();
+  } catch (...) {
+    return std::nullopt;
```
KY-U, line 134: Same rationale from the `deserialize_block` [comment](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/534#discussion_r3210976015)

modules/libbitcoinsystem/module.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+    bool is_witness = chain::script::is_pay_witness_key_hash_pattern(ops) ||
+                      chain::script::is_pay_witness_script_hash_pattern(ops) ||
+                      chain::script::is_pay_witness_taproot_pattern(ops);
+    result += is_witness ? "1" : "0";
+    result += chain::script::is_push_only_pattern(ops) ? "1" : "0";
+
+    return result;
+  } catch (...) {
+    return std::nullopt;
```
KY-U, line 50: Same rationale as `deserialize_block` comment

modules/libbitcoinsystem/Makefile
```diff
@@ -0,0 +1,114 @@
...
+
+format:
+	clang-format -i ./module.cpp ./module.h
+
+check-format:
+	clang-format -Werror --fail-on-incomplete-format -n ./module.cpp ./module.h
+
+clean:
+	rm -f *.o *.a secp256k1_symbols.map
+	rm -rf $(LIB_DIR)/build_fuzz $(LIB_DIR)/build_secp $(SECP_INSTALL) 2>/dev/null || true
```
KY-U, line 112:
```suggestion
clean:
        rm -f *.o *.a
	rm -rf $(LIB_DIR)/build_fuzz 2>/dev/null || true

clean-all:
	rm -f *.o *.a secp256k1_symbols.map
	rm -rf $(LIB_DIR)/build_fuzz $(LIB_DIR)/build_secp $(SECP_INSTALL) 2>/dev/null || true
```
suggestion / nit. Recompiling secp on every new iteration might be unnecessary when testing changes that are only related to libbitcoin-system.

KY-U:
With regards to the current failing CI tests:
- `deserialize_block` target failing was related to a [bug ](https://github.com/libbitcoin/libbitcoin-system/pull/1829)in libbitcoin-system. Also, it was found thanks to your PR @guptapratykshh !
- `psbt_parse` target [has been fixed](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/561).
- about the script / transaction targets, I've [forked](https://github.com/KY-U/libbitcoin-system/commit/8dcbb4ed4285049c222ee34368902811fcb40771) libbitcoin-system and added an injectable script checker that allows for mocking signature and context-reliant verifications such as `op_check_locktime_verify` so you can be sure that such verifications don't cause targets to fail. Currently it isn't passing CI, but I've successfully compiled and tested it locally.

<details>

<summary>Here is an example on how I've used it, I hope it can be of help!</summary>

```
  diff --git a/modules/libbitcoinsystem/module.cpp b/modules/libbitcoinsystem/module.cpp
  index 66f3288..b1f67dc 100644
  --- a/modules/libbitcoinsystem/module.cpp
  +++ b/modules/libbitcoinsystem/module.cpp
  @@ -5,11 +5,49 @@
   
   #include <bitcoin/system.hpp>
   #include <bitcoin/system/machine/interpreter.hpp>
  -
   #include "module.h"
   
   using namespace libbitcoin::system;
   
  +namespace {
  +class mock_signature_checker : public libbitcoin::system::machine::script_checker {
  +public:
  +  bool verify_ecdsa_signature(
  +    const data_slice& point,
  +    const hash_digest& hash,
  +    const ec_signature& signature) const NOEXCEPT override
  +  {
  +    return true;
  +  }
  +
  +  bool verify_schnorr_signature(
  +    const data_chunk& x_point,
  +    const hash_digest& hash,
  +    const ec_signature& signature) const NOEXCEPT override
  +  {
  +    return true;
  +  }
  +
  +  error::op_error_t verify_locktime(
  +    bool input_final,
  +    uint64_t stack_locktime,
  +    uint32_t transaction_locktime) const NOEXCEPT override
  +  {
  +    return error::op_success;
  +  }
  +
  +  error::op_error_t verify_sequence(
  +    uint32_t transaction_version,
  +    uint32_t stack_sequence,
  +    uint32_t input_sequence) const NOEXCEPT
  +  {
  +    return error::op_success;
  +  }
  +
  +  ~mock_signature_checker() override = default;  
  +};
  +}
  +
   namespace bitcoinfuzz {
   namespace module {
   
  @@ -254,7 +292,7 @@ LibbitcoinSystem::script_eval(const std::vector<uint8_t> &input_data,
       const chain::context ctx{chain::flags::no_rules, 0u, 0u, 0u, 0u, 0u};
   
       using interp = machine::interpreter<machine::contiguous_stack>;
  -    const auto ec = interp::connect(ctx, tx, 0u);
  +    const auto ec = interp::connect(ctx, tx, 0u, mock_signature_checker());
       return !ec;
     } catch (...) {
       return std::nullopt;
```

</details>

brunoerg:
Nice work, @KY-U! @guptapratykshh easy to move on.

brunoerg:
Closing due to lack of activity. @KY-U will move on this thing.

PR #535, opened 2026-04-15
refactor(driver): replace target dispatch chain with static table

mahmoudnasser1561:
This refactors `Driver::Run()` by replacing the long target-dispatch `if/else if` chain with a static dispatch table.

#### Why this is useful
The previous implementation encoded target dispatch as a long sequence of string comparisons and direct calls in one function. 
This change centralizes the target-to-handler mapping in one place while keeping the runtime behavior the same.

### DIFF
--- a/driver.cpp
+++ b/driver.cpp
@@ -1,4 +1,5 @@
 #include <algorithm>
+#include <array>
 #include <cassert>
 #include <fuzzer/FuzzedDataProvider.h>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


quapka:
As someone who does not see an immediate benefit (maybe due to my lack of practice in C++), can you elaborate on what the benefit is? Looks like a cosmetic change to me.

> This change centralizes the target-to-handler mapping in one place

Well, it's all in the same file and place anyway; it's not like the code was all over the place before.

Are there any concrete benefits, e.g., performance-wise? Also, is there any difference when this gets compiled at all? Possibly the compiler optimises things into similar code?

mahmoudnasser1561:
@quapka 
That is fair. I agree this change is mainly a style and maintainability refactor rather than something that should bring a noticeable performance improvement.

My motivation here was mostly code clarity. In many open source projects, long "if/else" dispatch chains tend to be viewed as harder to read and less pleasant to extend, especially in our case once the "number of targets" grows. What I was aiming for was a cleaner table-driven dispatch approach where the target-to-handler relationship is expressed more uniformly.

I hesitated a bit before opening the PR for the same reason, since I knew the benefit was not dramatic. Still, I felt it was a harmless cleanup that could make this part of the code a bit easier to follow and maintain, even if the practical gain is modest.

quapka:
As indicated in the other pull request mentioned, there might be a useful outcome from this. However, as suggested there as well, please try to identify valuable targets and high-impact modules, and rather focus your attention on them.

While I wouldn't call it harm, reviewing PRs takes non-trivial effort.

brunoerg:
> This refactors Driver::Run() by replacing the long target-dispatch if/else if chain with a static dispatch table

NACK. It is not a useful use of reviewers time.

mahmoudnasser1561:
@brunoerg 
If you happen to add support to a new target you would need to add another branch in the if-else chain
the current approach adopted in the code is not future proof.
my code changes solves this problem as you can compress the target table and the code style will be better.

quapka:
@mahmoudnasser1561, are you arguing that writing

```cpp
else if (target == "target_name") {
    this->TargetFunction(buffer);
  }
```
is _not future proof_ and more difficult than writing
```cpp
 {"target_name", &bitcoinfuzz::Driver::TargetName},
```
? I don't think your code solves a problem.

What do you mean by "you can compress the target table"? I hope you see that, compared to adding a new target (which could include adding a new module altogether), writing these couple of lines is not an issue.

mahmoudnasser1561:
@quapka

Yes, that’s fair at first glance, so let me explain my point a bit more clearly.

> What do you mean by "you can compress the target table"? I hope you see that, compared to adding a new target (which could include adding a new module altogether), writing these couple of lines is not an issue.

What I meant is not that adding one new target is hard by itself. I agree that adding a single target is just a few lines either way.

My point is more about maintenance.

With the current `if / else if` chain, each dispatch case has to live as its own control-flow branch, so as more targets get added over time, that section keeps getting longer and noisier. Right now it already takes a fairly large block of `driver.cpp`  (takes up the lines from 680 - 745).

With the target table approach, the dispatch becomes data-driven instead of branch-driven. The target-to-handler mapping is centralized in one place, and adding a new target is reduced to adding one entry like:

```cpp
{"target_name", &Driver::TargetName},
```
So the main benefit is not just “saving a couple of lines” for one target. It is that the dispatch path becomes simpler, easier to scan, and easier to extend as the number of targets grows.

For example, the table can be represented compactly as just a list of mappings:
```cpp
static constexpr std::array<TargetDispatchEntry, 29> kTargetDispatchTable{{
    {"script", &Driver::ScriptTarget}, {"deserialize_block", &Driver::BlockDeserializationTarget},
    {"script_eval", &Driver::ScriptEvalTarget}, {"verify_script", &Driver::VerifyScriptTarget},
    {"descriptor_parse", &Driver::DescriptorParseTarget}, {"miniscript_parse", &Driver::MiniscriptParseTarget},
    {"deserialize_invoice", &Driver::InvoiceDeserializationTarget}, {"address_parse", &Driver::AddressParseTarget},
    {"psbt_parse", &Driver::PSBTParseTarget}, {"addrv2", &Driver::AddrV2Target},
    {"deserialize_offer", &Driver::OfferDeserializationTarget}, {"cmpctblocks_parse", &Driver::CompactBlocksTarget},
    {"parse_p2p_message", &Driver::ParseP2PMessageTarget}, {"parse_p2p_lightning_message", &Driver::ParseLightningP2pMessageTarget},
    {"transaction_eval", &Driver::TransactionEvalTarget}, {"bip32_master_keygen", &Driver::Bip32MasterKeygenTarget},
    {"kernel_block", &Driver::KernelBlockTarget}, {"kernel_transaction", &Driver::KernelTransactionTarget},
    {"private_to_public_key", &Driver::PrivateToPublicKeyTarget}, {"sign_compact", &Driver::SignCompactTarget},
    {"sign_der", &Driver::SignDerTarget}, {"sign_verify", &Driver::SignVerifyTarget},
    {"ecdh", &Driver::ECDHTarget}, {"sign_schnorr", &Driver::SignSchnorrTarget},
    {"bip32_deserialize_extended_key", &Driver::Bip32DeserializeExtendedKeyTarget},
    {"decode_ellswift", &Driver::DecodeEllswiftTarget}, {"schnorr_verify", &Driver::SchnorrVerifyTarget},
    {"decode_onion", &Driver::DecodeOnionTarget}, {"stump_modify_add", &Driver::StumpModifyAddTarget},
}};
```

So for me the value is mainly in making the dispatch structure cleaner and more maintainable if more targets are added in the future, not in claiming that adding one target today is difficult.

It’s totally okay if you still don’t find that valuable. No problem at all :)

quapka:
> I agree that adding a single target is just a few lines either way

Yes, it's easy to add new targets.

> so as more targets get added over time, that section keeps getting longer and noisier

The noise level is just your subjective feeling. To me the "compressed" table is much, much messier.

> making the dispatch structure cleaner and more maintainable

There is little to no maintenance - just add a target, and that's it.

> if more targets are added in the future

We wish! Adding (high-impact) targets and (high-impact) modules IS the valuable part. Please, focus on that.

PR #536, opened 2026-04-15
Add listtargets helper script

mansi104-ai:
This PR adds a standalone helper script at tools/listtargets.py that parses the Driver::Run() dispatch block in driver.cpp and outputs the available fuzz targets.

It includes:

--format=json support for machine-readable output
a reusable parsing function for future tooling
a pytest test covering both direct parsing and CLI JSON output

### DIFF
--- a/tests/test_listtargets.py
+++ b/tests/test_listtargets.py
@@ -1,0 +1,58 @@
+import json
+import subprocess
+import sys
+from pathlib import Path
+
+from tools.listtargets import extract_targets
+
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+SCRIPT_PATH = REPO_ROOT / "tools" / "listtargets.py"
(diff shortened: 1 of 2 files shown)


quapka:
Listing existing targets is rather a nice-to-have functionality, but not that much missing. Either way, I don't agree with this _hacky_ approach of using RegEx to parse the `driver.cpp` file. Possibly, this could be combined with https://github.com/bitcoinfuzz/bitcoinfuzz/pull/535 to provide a richer CLI interface, allowing `./bitcoinfuzz --list-targets` to output the targets table.

In general, however, I suggest focusing on which valuable targets or modules could be added for the actual fuzzing in a PR, rather than adding extra scripts that don't do much and whose implementation is rather fragile. Bitcoinfuzz is a _meta_ tool; its value comes from the fuzzing identifying discrepancies among implementations and then analysing them.

I am expecting this to get closed.

brunoerg:
NACK

PR #541, opened 2026-04-17
java modules: harden JVM init failure paths

Shreyas2004wagh:
## Summary

Harden JVM/JNI initialization in the Java-backed modules.

## Why

`bitcoinj`, `eclair`, and `lightningkmp` were all exposing partially initialized JNI state during setup.

In the current flow, a module could:
- create a global class reference
- fail a later `GetStaticMethodID(...)`
- return failure while leaving partially initialized state behind

`eclair` also aborted the whole process when JVM initialization failed, and `eclair` / `lightningkmp` were using `GetEnv(...)` results without validating success.

## Change

- build JNI init state in local temporaries first
- assign static class/method handles only after the full init path succeeds
- delete global refs on partial init failure
- clear pending JNI exceptions after failed class/method lookups
- validate `GetEnv(...)` before using `JNIEnv*`
- return normal module failure for `eclair` init errors instead of aborting the process

## Scope

This change is limited to:
- `modules/bitcoinj/module.cpp`
- `modules/eclair/module.cpp`
- `modules/lightningkmp/module.cpp`

### DIFF
--- a/modules/bitcoinj/module.cpp
+++ b/modules/bitcoinj/module.cpp
@@ -17,6 +17,12 @@ static jclass bitcoinJWrapperClass = nullptr;
 static jmethodID createMasterKeyMethod = nullptr;
 static jmethodID deserializeExtendedKeyMethod = nullptr;
 
+static void clear_pending_exception(JNIEnv *env) {
+  if (env->ExceptionCheck()) {
+    env->ExceptionClear();
+  }
+}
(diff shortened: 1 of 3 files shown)


erickcestari (approved):
Thanks!

ACK 72ab3f488e41050ee4357bf8a657c01755ea02ce

PR #542, opened 2026-04-18
build scripts: avoid changing the global Rust toolchain

Shreyas2004wagh:
## Summary

Stop mutating the global Rust default toolchain from the helper build scripts.

## Why

`auto_build.py` and `scripts/build-all.sh` were calling `rustup default nightly` before building Rust-backed modules.

That changes the user's global Rust toolchain outside the repo, which is an unnecessary side effect for project-local build helpers.

The Rust module Makefiles already use scoped nightly invocations via `cargo +nightly`, so these scripts do not need to switch the global default toolchain.

## Change

- remove `rustup default nightly` from `auto_build.py`
- remove `rustup default nightly` from `scripts/build-all.sh`

## Scope

This change is limited to:
- `auto_build.py`
- `scripts/build-all.sh`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -92,7 +92,6 @@ RUN \
     --mount=type=cache,target=/root/.gradle,id=fuzz-gradle \
     --mount=type=cache,target=/root/.m2,id=fuzz-maven \
     --mount=type=cache,target=/root/.nuget/packages,id=fuzz-nuget-build \
-    --mount=type=cache,target=/root/.rustup,id=fuzz-rustup \
     --mount=type=cache,target=/root/go/pkg/mod,id=fuzz-go-mod \
     export JAVA_HOME=/usr/lib/jvm/java-21-openjdk-$(dpkg --print-architecture) && \
     /build/auto_build.py
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,6 +49,9 @@ RUN --mount=type=cache,target=/var/cache/apt,id=fuzz-apt-cache-builder \
     rustup \
     unzip
 
+# Keep Rust nightly scoped to the builder image instead of mutating a host toolchain.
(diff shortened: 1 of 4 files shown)


Dockerfile
erickcestari: Removing the line: `-mount=type=cache,target=/root/.rustup,id=fuzz-rustup` should fix the CI docker build image check 

```suggestion
```

erickcestari (changes requested):

Dockerfile
Shreyas2004wagh: Works.

erickcestari (approved):
Thanks!

ACK 9ecfef0c887cf65e8eb9bb19a590a2226482ed4e

quapka:
@erickcestari doesn't this "break" local, non-CI build? It makes the CI and docker image default to nightly, but what about running `auto_build.py` locally when nightly is not set?

Nonetheless, if `rustup default nightly` was indeed updating the users global toolchain, then building the nightly targets would cause subsequent, non-nightly builds to also default to nightly.

Could we use `rust-toolchain.toml`, as described [here](https://stackoverflow.com/a/64701936) on a per-project basis?

erickcestari:
> @erickcestari doesn't this "break" local, non-CI build? It makes the CI and docker image default to nightly, but what about running `auto_build.py` locally when nightly is not set?
> 
> Nonetheless, if `rustup default nightly` was indeed updating the users global toolchain, then building the nightly targets would cause subsequent, non-nightly builds to also default to nightly.
> 
> Could we use `rust-toolchain.toml`, as described [here](https://stackoverflow.com/a/64701936) on a per-project basis?

We could use the `rust-toolchain.toml`. However, it's not necessary since all of the Makefiles related to rust in our project use `cargo +nightly.` Therefore, defaulting to Rust Nightly wasn't necessary.

PR #544, opened 2026-04-19
Optimize memory deallocation in LockedPool using a map for arena lookup

kartikangiras:
Implemented a Performance Optimization marked TODO in lockedpool.


kartikangiras:
@brunoerg PTAL into this PR.

erickcestari:
Hi @kartikangiras,

That code comes directly from Bitcoin Core. It was copied over as-is. We use those utils for custom mutations, since they provide useful helper functions for implementing the mutators.

I don’t think changes are needed on our side. If this is truly an optimization worth making, it would probably be better to open a PR against Bitcoin Core.

PR #546, opened 2026-04-19
Add SecureRPCParameters type for locking RPC parameter data in memory

kartikangiras:
Implement RPC request parameters use mlock() for memory protection TODO


erickcestari:
Closed by the same reason as: #544

PR #549, opened 2026-04-21
driver: add LOG_OUTPUTS env var to log every module response

erickcestari:
When set, each accepted module response is printed as "Module: X / Result: Y" alongside the existing mismatch reporting. Useful for debugging replays where you want to see what every module returned for a given input.

The flag is read once in LLVMFuzzerInitialize (alongside FUZZ) and plumbed through the Driver constructor. LogResponse and VerifyMatchingResponse are private member templates so call sites don't need to thread the flag around.

### DIFF
--- a/driver.cpp
+++ b/driver.cpp
@@ -10,29 +10,36 @@
 #include <bitcoinfuzz/basemodule.h>
 #include <bitcoinfuzz/module_registry.h>
 
-namespace {
+namespace bitcoinfuzz {
 template <typename T>
-void VerifyMatchingResponse(std::optional<T> &last_response,
-                            std::string &last_module_name,
-                            const std::string &module_name, const T &response,
-                            std::string_view failure_message) {
-  if (last_response.has_value()) {
-    if (response != *last_response) {
-      std::cout << failure_message << std::endl;
-      std::cout << "Module: " << module_name << std::endl;
-      std::cout << "Result: " << response << std::endl;
-      std::cout << "Module: " << last_module_name << std::endl;
-      std::cout << "Result: " << *last_response << std::endl;
-    }
-    assert(response == *last_response);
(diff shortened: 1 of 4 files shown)


driver.cpp
```diff
@@ -282,6 +295,7 @@ void Driver::AddrV2Target(std::span<const uint8_t> buffer) const {
     std::optional<std::string> res{module.second->addrv2_parse(buffer)};
     if (!res.has_value())
       continue;
+    LogResponse(module.first, *res);
```
brunoerg, line 297: nit: miss blank line after `LogResponse`.

driver.cpp
```diff
@@ -10,12 +10,23 @@
...
+  std::cout << "Module: " << module_name << std::endl;
+  std::cout << "Result: " << response << std::endl;
+}
+
 template <typename T>
-void VerifyMatchingResponse(std::optional<T> &last_response,
-                            std::string &last_module_name,
-                            const std::string &module_name, const T &response,
-                            std::string_view failure_message) {
+void Driver::VerifyMatchingResponse(std::optional<T> &last_response,
```
brunoerg, line 24: `VerifyMatchingResponse` calls `LogResponse` at the top, then if a mismatch is found it also prints the same "Module: X\nResult: Y" inline. When LOG_OUTPUTS=1 and a mismatch fires, the current module's response appears twice in sequence. It won't cause a wrong result but it's confusing during triage.

brunoerg:
Concept ACK, left a comment.

driver.cpp
erickcestari: Fixed!

brunoerg (approved):
ACK 2b1fe8a33987d5bf8f625303c80cc82fde45a2bb

Issue #288, opened 2025-10-06
descriptor_parse: parse failed for `raw(bc)`


brunoerg:
```sh
Descriptor parse failed for raw(bc)
Module: NBitcoin
Result: 1
Module: Bitcoin
Result: 0
```

brunoerg:
NBitcoin doesn't reject this descriptor because `bc` is a hex thing. However, when parsing this descriptor Bitcoin Core also parses the script which is invalid and then reject it.

brunoerg:
```sh
Descriptor parse failed for raw(77227225555755575555755575)
Module: NBitcoin
Result: 1
Module: Bitcoin
Result: 0
```

PR #550, opened 2026-04-23
Fix/descriptor raw nbitcoin known diff combined with PR #533 for nbitcoin PSBT

rocketpowerkille:
this change addresses false positives in descriptor parsing fuzzing where NBitcoin module produces different results than other modules for raw() descriptors,which are known implementation differences.

- added IsKnownNBitcoinRawDescriptorDifference helper function
- modified DescriptorParseTarget to skip known NBitcoin differences under NBITCOIN flag

this is combined with the pr #533 and in accordance with issue #288 and helps reduce noise in fuzzing results by ignoring expected variations between descriptor parser implementations while preserving detection of genuine issues.


brunoerg:
Needs rebase.

rocketpowerkille:
@brunoerg ready

brunoerg:
NACK 

I think we cannot simply skip it without understanding the issue.

PR #556, opened 2026-05-03
driver: scope script opcode workarounds to affected modules

Shreyas2004wagh:
## Summary

Scope existing module-specific opcode workarounds in `script_eval` and `verify_script` to the affected modules only, instead of returning from the whole fuzz target before comparisons begin.

## Problem

`Driver::ScriptEvalTarget()` and `Driver::VerifyScriptTarget()` currently apply some known implementation-specific workarounds with an early `return`.

That means when one enabled module has a known discrepancy-triggering input, the entire target exits and we skip running the same input against other loaded modules that are not affected by that workaround.

In practice this reduces differential fuzzing coverage and can hide mismatches in unrelated modules.

## Change

- keep the existing workaround conditions
- move them inside the per-module loop
- skip only the affected module(s) for those inputs instead of skipping the whole target

Specifically:
- skip `Btcd` in `script_eval` for inputs containing `0xAC`, `0xAE`, or `0xAF`
- skip `NBitcoin` in `script_eval` for inputs containing `0xB2`
- skip `Btcd` and `Gocoin` in `verify_script` for the known `0xAC..0xAF` discrepancy case

## Why this is better

This preserves the existing compatibility workaround behavior while allowing unaffected modules to continue participating in the same fuzz run.

That improves target usefulness by avoiding false negatives caused by whole-target early exits.

## Scope

This change is limited to:
- `driver.cpp`

## Validation

- reviewed the current workaround logic and confirmed the previous control flow returned before the module comparison loop
- verified the updated logic now skips only the affected module names while preserving the rest of the target execution

### DIFF
--- a/driver.cpp
+++ b/driver.cpp
@@ -115,7 +115,9 @@ void Driver::VerifyScriptTarget(std::span<const uint8_t> buffer) const {
 
   // Skip these opcodes only for implementations that disagree with Core's
   // FindAndDelete handling, while continuing to compare all other modules.
-  auto opcodes_to_skip = [](unsigned char op) { return op >= 0xAC && op <= 0xAF; };
+  auto opcodes_to_skip = [](unsigned char op) {
+    return op >= 0xAC && op <= 0xAF;
+  };
   bool has_opcode_mismatch =
       std::ranges::any_of(script_sig, opcodes_to_skip) ||
       std::ranges::any_of(script_pubkey, opcodes_to_skip);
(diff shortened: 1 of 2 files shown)


brunoerg:
Needs rebase

Shreyas2004wagh:
Hey @brunoerg , Rebased onto the latest v2 and force-pushed.

brunoerg:
Concept ACK

driver.cpp
```diff
@@ -115,7 +115,9 @@ void Driver::VerifyScriptTarget(std::span<const uint8_t> buffer) const {
 
   // Skip these opcodes only for implementations that disagree with Core's
   // FindAndDelete handling, while continuing to compare all other modules.
-  auto opcodes_to_skip = [](unsigned char op) { return op >= 0xAC && op <= 0xAF; };
+  auto opcodes_to_skip = [](unsigned char op) {
+    return op >= 0xAC && op <= 0xAF;
+  };
```
KY-U, line 120: Nit: this formatting-only change belongs more naturally to the [previous commit](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/556/changes/a2c402c451e1b50d9cbeaf12526be9d2debea20f) as opposed to the [current one](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/556/changes/a4ecc948575b0acb37724da822bec8c1c154b33c), since it's that commit that introduces this lambda function and touches on `VerifyScriptTarget`.

KY-U:
Code Review ACK. My previous comment is mainly a commit hygiene nit.

PR #561, opened 2026-05-07
Fix: skip completely on 0-input/0-output PSBT match

wolgwang1729:
# Title
Fix: skip completely on 0-input/0-output PSBT match

# Description
This PR resolves the test failures introduced after PR #533.

PR #533 changed an early `return` to a `continue` in `Driver::PSBTParseTarget` and scoped it to NBitcoin. However, the other libraries (like `rust-bitcoin`, `libwally`) also disagree on how to handle non-standard 0-input / 0-output PSBTs. 

Since these edge-case discrepancies are known and we do not want them blocking the fuzzing process (as discussed with  brunoerg), this PR:
1. Removes the `module.first == "NBitcoin"` restriction.
2. Reverts the logic to `return;` so that if *any* module encounters an `in=0` or `out=0` PSBT, the entire fuzz target gracefully exits. This guarantees no partial, mismatching comparisons are logged for this edge case.

### DIFF
--- a/driver.cpp
+++ b/driver.cpp
@@ -232,14 +232,11 @@ void Driver::PSBTParseTarget(std::span<const uint8_t> buffer) const {
     if (!res.has_value())
       continue;
 
-#ifdef NBITCOIN
-    if (module.first == "NBitcoin" &&
-        ModuleRegistry::instance().isEnabled("NBITCOIN")) {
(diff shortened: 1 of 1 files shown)


brunoerg (commented):
ACK 3b9ac80ef79de38b5d815f80342b12f7c52aa089 - thank you

PR #562, opened 2026-05-14
fix: libwally pubkey result fix in bip32_deserialize_extended_key

kuliq23:
Noticed this off-by-one bug when going through the BIP32 targets... we should only skip for private keys.

### DIFF
--- a/modules/libwallycore/module.cpp
+++ b/modules/libwallycore/module.cpp
@@ -154,7 +154,7 @@ std::optional<std::string> LibwallyCore::bip32_deserialize_extended_key(
       result << std::setw(2) << std::setfill('0')
              << static_cast<int>(key.priv_key[i]);
   } else {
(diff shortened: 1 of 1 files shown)


erickcestari:
> Noticed this off-by-one bug when going through the BIP32 targets... we should only skip for private keys.

I'm curious to know how we didn't catch this earlier if it's a off by one in the returned string. Shouldn't a simple run already caught this issue?

kuliq23:
@erickcestari Good question. Since libwally-core has the missing TV5 validation problem, the fuzzer manages to stop before this comes into play. I tested the target with hardcoded xpubs and this change is correct.

erickcestari (approved):
Makes sense! Thanks.

ACK 87d9fa51f910e7f9320f25fed53ebb65881d5df8

PR #572, opened 2026-06-05
Add libbitcoin trophies

KY-U:
Adding some consensus-related findings on libbitcoin


brunoerg (changes requested):
Can we put the new ones at the end of the list? We usually order it by date.

brunoerg (approved):
ACK b062b0bd274983b7be110ef20f55c8897eeaac2b

PR #575, opened 2026-06-10
Add libbitcoin-system integration

KY-U:
Continuing a previously opened [PR](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/534#issuecomment-4578054115) that adds a libbitcoin module.


modules/libbitcoinsystem/module.cpp
```diff
@@ -0,0 +1,320 @@
...
+    return std::nullopt;
+  }
+  data_chunk script_bytes(input_data.begin(), input_data.end());
+
+  const chain::script in_script{script_bytes, false};
+  if (!in_script.is_valid()) {
+    return false;
+  }
+
+  // Uses an 0P_1 scripPubkey to alway push truthy, so that script_eval tests
```
brunoerg, line 267:
```suggestion
  // Uses an 0P_1 scripPubkey to always push truthy, so that script_eval tests
```

modules/libbitcoinsystem/module.cpp
```diff
@@ -0,0 +1,320 @@
...
+  chain::points points;
+  const auto inputs_ptr = tx.inputs_ptr();
+  for (const auto &input_ptr : *inputs_ptr) {
+    points.push_back(input_ptr->point());
+  }
+  // Check for double spend, CVE-2018-17144
+  if (!is_distinct(points))
+    return "0";
+
+  // Check for spend overflow, CVE-2010-5139.
```
brunoerg, line 102: libbitcoin by itself doesn't check it?

brunoerg (approved):
ACK 6a81982fef4ff157a18a24f0098b1dfd559667ae

modules/libbitcoinsystem/module.cpp
evoskuil: Not in .check(), as from a consensus standpoint that would be redundant with block internal double spend checks. This type of check is only a DoS guard, and those are placed independently (into "guards") so that they do not waste CPU when performing actual validation:
```cpp
// Guards (for tx pool without compact blocks).
// ----------------------------------------------------------------------------

// Pools do not have coinbases.
// Redundant with block is_internal_double_spend check.
// Redundant with block max_block_size check.
code transaction::check_guard() const NOEXCEPT
{
    if (is_coinbase())
        return error::coinbase_transaction;
    if (is_internal_double_spend())
        return error::transaction_internal_double_spend;
    if (is_oversized())
        return error::transaction_size_limit;

    return error::transaction_success;
}

// Redundant with block max_block_weight accept.
code transaction::check_guard(const context& ctx) const NOEXCEPT
{
    const auto bip141 = ctx.is_enabled(flags::bip141_rule);

     if (!bip141 && is_segregated())
        return error::unexpected_witness_transaction;
     if (bip141 && is_overweight())
        return error::transaction_weight_limit;

    return error::transaction_success;
}

// Redundant with block max_block_sigops accept.
code transaction::accept_guard(const context& ctx) const NOEXCEPT
{
    const auto bip16 = ctx.is_enabled(flags::bip16_rule);
    const auto bip141 = ctx.is_enabled(flags::bip141_rule);

    if (is_missing_prevouts())
        return error::missing_previous_output;
    if (is_signature_operations_limited(bip16, bip141))
        return error::transaction_sigop_limit;

    return error::transaction_success;
}
111

File: modules/libbitcoinsystem/module.cpp [hunk 3]
```diff
@@ -0,0 +1,320 @@
...
+  return encode_hash(blk.hash());
+}
+
+std::optional<std::string>
+LibbitcoinSystem::address_parse(std::string str) const {
+  try {
+    wallet::payment_address addr(str);
+
+    if (!addr) {
+      // libbitcoin-system's payment_address only handles legacy P2PKH/P2SH.
```
evoskuil, line 142: This class is designed specifically for non-segwit addresses. There is another class for that (`witness_address`) as they are a very distinct type. There is also `stealth_address`.

modules/libbitcoinsystem/module.cpp
```diff
@@ -0,0 +1,320 @@
...
+  static constexpr uint64_t max_money =
+      21'000'000ULL * chain::satoshi_per_bitcoin;
+  uint64_t total = 0;
+  // is_valid() only signals deserialization succeeded — it accepts
+  // 0-input/0-output and other consensus-invalid forms that bitcoin core's
+  // CheckTransaction rejects. Use check() (the libbitcoin-system equivalent)
+  // to align acceptance with the reference modules. However, check() only
+  // verifies the transaction's internal consistency, not whether its inputs are
+  // valid or unspent, so we also check for distinct inputs to match bitcoin
+  // core's validation.
```
evoskuil, line 88: Also note that validation is clearly differentiated from DoS protection. See below.

modules/libbitcoinsystem/module.cpp
```diff
@@ -0,0 +1,320 @@
...
+  // to match core's CScript deserialization
+  chain::script script{data, true};
+
+  if (!script.is_valid())
+    return "0";
+
+  // Match bitcoin core's IsUnspendable(): OP_RETURN prefix or oversize
+  // script. Libbitcoin's script::is_unspendable() is broader:
+  // it treats parse-underflow/invalid leading operations as unspendable,
+  // while Core only checks OP_RETURN prefix and MAX_SCRIPT_SIZE here.
```
evoskuil, line 60: reserved leading, invalid anywhere (note that these categories are not consistent in bitcoind, so you have to refer to our classifications, which are strictly behavior-driven).

modules/libbitcoinsystem/module.cpp
```diff
@@ -0,0 +1,320 @@
...
+namespace bitcoinfuzz {
+namespace module {
+
+LibbitcoinSystem::LibbitcoinSystem(void) : BaseModule("LibbitcoinSystem") {}
+
+std::optional<std::string>
+LibbitcoinSystem::script_parse(std::span<const uint8_t> buffer) const {
+  data_chunk data(buffer.begin(), buffer.end());
+
+  // Prefix set to true treats the first byte as lenght
```
evoskuil, line 50: typo: lenght

evoskuil (commented):
Nice! Added some info comments.

modules/libbitcoinsystem/module.cpp
KY-U: Thanks for pointing that out. Adding `witness_address` to the target [now](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/582/changes/8d32119b3083fdee905e9d5b6966706c70dc76cb#diff-b5613606750d29c631dbc82f4ef4b9704fbb17be903f1da809745a118554ade8R159-R178)!

modules/libbitcoinsystem/module.cpp
KY-U: [Changed the comment ](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/582/changes/5a10e630bdcd72ed28dee2e021d06ff188b7a9b3#diff-b5613606750d29c631dbc82f4ef4b9704fbb17be903f1da809745a118554ade8R83-R89)to better reflect that nuance.

KY-U:
> Nice! Added some info comments.

Thanks for the feedback! I've addressed your points [here](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/582)

PR #580, opened 2026-06-10
Fix broken ci build after libbitcoin integration

KY-U:
[Libbitcoin integration ](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/575)introduced some problems during ci build that weren't caught due to `commit-by-commit` workflow not being run on PRs with a single commit. This PR aims to standardize `cmake` and `boost` installations across different workflows. Those changes include:

- Add 'json' to the BOOST_INCLUDE_LIBRARIES variable.
- move cmake installation to the 'Setup environment' step, instead of inside boost installation.
- move boost installation to always be executed after dependencies install.

### DIFF
--- a/.github/actions/setup-test-env/action.yml
+++ b/.github/actions/setup-test-env/action.yml
@@ -36,14 +36,16 @@ runs:
       run: |
         sudo apt-get update
         sudo apt-get install -y lowdown libsodium-dev
-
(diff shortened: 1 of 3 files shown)


brunoerg (approved):
ACK 3e708bf50a575a770976a5094551765597e7e29f

PR #581, opened 2026-06-10
Skip commit-by-commit-build step for documentation only commits

KY-U:
Depends on #580 

`commit-by-commit-build` workflow only skipped execution if every commit in the PR contained `.md` only changes. This PR adjusts the wofklow so that it skips building individual commits that change only `.md` files, even if there are other build-relevant commits in the PR.


brunoerg (approved):
ACK 0747ca9672e7ab0a6f4442b0d8f1df5dec9977ea

PR #582, opened 2026-06-11
integrate witness addresses into libbitcoin module's adress_parse target

KY-U:
- Integrate `WPKH`, `WSH` and `TR` witness addresses into libbitcoin's `address_parse` parse target.
- Clarify comments regarding some architectural differences between Core and Libbitcoin

### DIFF
--- a/modules/libbitcoinsystem/module.cpp
+++ b/modules/libbitcoinsystem/module.cpp
@@ -47,17 +47,20 @@ std::optional<std::string>
 LibbitcoinSystem::script_parse(std::span<const uint8_t> buffer) const {
   data_chunk data(buffer.begin(), buffer.end());
 
-  // Prefix set to true treats the first byte as lenght
+  // Prefix set to true treats the first byte as length
   // to match core's CScript deserialization
   chain::script script{data, true};
 
   if (!script.is_valid())
     return "0";
 
-  // Match bitcoin core's IsUnspendable(): OP_RETURN prefix or oversize
-  // script. Libbitcoin's script::is_unspendable() is broader:
-  // it treats parse-underflow/invalid leading operations as unspendable,
-  // while Core only checks OP_RETURN prefix and MAX_SCRIPT_SIZE here.
+  // Match Bitcoin Core's IsUnspendable(), which checks for OP_RETURN prefix or
+  // oversize script. Libbitcoin's script::is_unspendable() is a separate,
+  // broader unspendability predicate, not the parser's short-circuit logic: it
+  // classifies scripts whose first opcode is reserved or invalid as provably
+  // unspendable. The reserved set includes OP_RETURN, but also other opcodes
+  // that fail when reached as the first executed opcode, so using it here would
+  // reject more scripts than Core's IsUnspendable().
   const auto &ops = script.ops();
   if ((!ops.empty() && ops.front() == chain::opcode::op_return) ||
       script.is_oversized())
@@ -81,18 +84,20 @@ LibbitcoinSystem::transaction_eval(std::span<const uint8_t> buffer) const {
   uint64_t total = 0;
   // is_valid() only signals deserialization succeeded — it accepts
   // 0-input/0-output and other consensus-invalid forms that bitcoin core's
-  // CheckTransaction rejects. Use check() (the libbitcoin-system equivalent)
-  // to align acceptance with the reference modules. However, check() only
-  // verifies the transaction's internal consistency, not whether its inputs are
-  // valid or unspent, so we also check for distinct inputs to match bitcoin
-  // core's validation.
+  // CheckTransaction rejects. Use check() for libbitcoin-system's
+  // context-free transaction checks. Some Core CheckTransaction rules, such as
+  // duplicate inputs and output sum overflow, are placed differently in
+  // libbitcoin: they are DoS guards or block-level redundancy checks rather
+  // than part of transaction::check(). They are applied here explicitly so this
+  // standalone transaction target matches Core's acceptance
+  // surface.
   if (!tx.is_valid() || tx.check()) {
     return "0";
   }
 
   // Check for double spend, CVE-2018-17144. Libbitcoin checks for double spend
   // during block validation at block::is_internal_double_spend() called from
-  // block::check()
+  // block::check().
   chain::points points;
   const auto inputs_ptr = tx.inputs_ptr();
   for (const auto &input_ptr : *inputs_ptr) {
@@ -288,7 +293,7 @@ LibbitcoinSystem::script_eval(const std::vector<uint8_t> &input_data,
 
   // Uses an 0P_1 scripPubkey to always push truthy, so that script_eval tests
   // whether script executes without error, as oposed to whether it leaves
-  // a truth value on stack, which is tested by ggipt target.
+  // a truth value on stack, which is tested by verify_script target.
   // Concrete divergence: scriptSig=OP_0 → bitcoin core false,
   // libbitcoin (with OP_1) returns true.
   const data_chunk out_bytes{0x51}; // OP_1
--- a/modules/libbitcoinsystem/module.cpp
+++ b/modules/libbitcoinsystem/module.cpp
@@ -139,28 +139,47 @@ LibbitcoinSystem::deserialize_block(std::span<const uint8_t> buffer) const {
 std::optional<std::string>
 LibbitcoinSystem::address_parse(std::string str) const {
   try {
-    wallet::payment_address addr(str);
-
-    if (!addr) {
-      // libbitcoin-system's payment_address only handles legacy P2PKH/P2SH.
-      // For bech32/segwit addresses (P2WPKH, P2WSH, P2TR) it returns an
-      // invalid addr while other modules return "WPKH:", "WSH:", "TR:".
-      // Abstain (nullopt) instead of falsely reporting "INVALID".
-      return std::nullopt;
-    }
-
     std::string prefix;
-    if (addr.prefix() == wallet::payment_address::mainnet_p2kh) {
-      prefix = "PKH:";
-    } else if (addr.prefix() == wallet::payment_address::mainnet_p2sh) {
-      prefix = "SH:";
-    } else {
-      prefix = "UNK:";
+
+    wallet::payment_address legacy_addr(str);
+    if (legacy_addr) {
+      switch (legacy_addr.prefix()) {
+      case wallet::payment_address::mainnet_p2kh:
+        prefix = "PKH:";
+        break;
+      case wallet::payment_address::mainnet_p2sh:
+        prefix = "SH:";
+        break;
+      default:
+        prefix = "UNK:";
+      }
+      return prefix + legacy_addr.encoded();
     }
 
-    return prefix + addr.encoded();
-  } catch (...) {
-    return std::nullopt;
+    wallet::witness_address segwit_addr(str, false);
+    if (segwit_addr &&
+        segwit_addr.prefix() == wallet::witness_address::mainnet) {
+      switch (segwit_addr.identifier()) {
+      case wallet::witness_address::program_type::version0_p2kh:
+        prefix = "WPKH:";
+        break;
+      case wallet::witness_address::program_type::version0_p2sh:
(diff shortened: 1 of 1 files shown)


modules/libbitcoinsystem/module.cpp
```diff
@@ -47,17 +47,16 @@ std::optional<std::string>
...
   if (!script.is_valid())
     return "0";
 
-  // Match bitcoin core's IsUnspendable(): OP_RETURN prefix or oversize
-  // script. Libbitcoin's script::is_unspendable() is broader:
-  // it treats parse-underflow/invalid leading operations as unspendable,
-  // while Core only checks OP_RETURN prefix and MAX_SCRIPT_SIZE here.
+  // Match bitcoin core's IsUnspendable(), which checks for OP_RETURN prefix or
+  // oversize script. Libbitcoin's script::is_unspendable() models a different
+  // fast-fail condition based on reserved/invalid leading opcodes
```
evoskuil: I would just clarify for your info that the invalid opcode check is not limited to a “leading” op. Unlike reserved codes who are conditional failures, and invalid code is unconditional (will cause script failure regardless of position or context), so if we see one anywhere in the script it is marked unspendable (fail fast on eval).

KY-U: Removed the "leading opcode" section and focused the explanation on invalid opcodes.

brunoerg:
@evoskuil Could you do a final look here?

evoskuil:
> @evoskuil Could you do a final look here?

Sure, the short-circuit conditions are more extensive than the summary. During script parse each opcode is tested against three criteria. Notably this excludes a short-circuit in the case where the first opcode is `reserved`. Reserved includes `op_return` and a large number of other codes, though this set is reduced (most of it moved into `invalid`) when executing a tapscript. There is no significant performance benefit to short-circuiting the first opcode, since it is the first tested during script execution. However any `reserved` opcode (including `op_return`) in the first position fails the script.

The three conditions we test during parse are:
```cpp
    bool easier{};
    bool failer{};
    bool roller{};

    for (const auto& op: ops)
    {
        easier |= op.is_success();
        failer |= op.is_invalid();
        roller |= op.is_roller();
    }
```
Under `tapscript` a success code anywhere in the script unconditionally succeeds the script. This overrides the second condition, in which any `invalid` code fails the script (unconditional outside of tapscript, and without a success code within tapscript).

The first line of `script::run()` evaluates this parse:
```cpp
TEMPLATE
code CLASS::
run() NOEXCEPT
{
    if (const auto ec = state::initialize())
        return ec;

    for (auto it = state::begin(); it != state::end(); ++it)
    {
        const auto& op = *it;

        if (op.is_oversized())
            return error::invalid_push_data_size;

        if (!state::ops_increment(op))
            return error::invalid_operation_count;

        if (state::if_(op))
        {
            if (const auto ec = run_op(it))
                return ec;

            if (state::is_stack_overflow())
                return error::invalid_stack_size;
        }
    }

    if (!state::is_balanced())
        return error::invalid_stack_scope;

    return error::script_success;
}
```
And is shown here:
```cpp
TEMPLATE
INLINE error::script_error_t CLASS::
initialize() const NOEXCEPT
{
    // TODO: nops rule must first be enabled in tests and config.
    constexpr auto nops = true; ////is_enabled(flags::nops_rule);
    const auto bip141 = is_enabled(flags::bip141_rule);
    const auto bip342 = is_enabled(flags::bip342_rule);

    // Succeed if any success code, overrides all codes and errors [bip342].
    if (bip342 && script_->is_prevalid())
        return error::prevalid_script;

    // Fail if last op is underflow (same behavior as prefail, invalid code).
    if (script_->is_underflow())
        return error::invalid_script;

    // Stack limit (1,000) applied to initial stack [bip342].
    if (bip342 && is_stack_overflow())
        return error::invalid_stack_size;

    // Apply stack element limit (520) to initial witness [bip342|bip141].
    if ((bip342 || bip141) && witness_ && !chain::witness::is_push_size(*witness_))
        return error::invalid_witness_stack;

    // Script size limit (10,000) [0.3.7+], removed [bip342].
    if (!bip342 && nops && script_->is_oversized())
        return error::invalid_script_size;

    // Fail if any op invalid (invalid codes reduced in tapscript).
    // Should be after underflow check since underflow is also an invalid op.
    // Promoted success codes are not reachable here due to is_prevalid above.
    // So only op_verif/op_vernotif (unpromoted invalids) are caught here for
    // tapscript, otherwise is_prevalid bypassed and all invalids caught here.
    if (script_->is_prefail())
        return error::prefail_script;

    return error::script_success;
}
```

These conditions short-circuit the script with success/failure based on the first two accumulated parse states:
```cpp
    if (bip342 && script_->is_prevalid())
        return error::prevalid_script;
        ...
    if (script_->is_prefail())
        return error::prefail_script;
```

This remaining conditions short-circuit the script with failure. This includes `underflow`. This is set during parse of the script buffer into opcodes (prior to the above parse step). The only invalid parse is a script with a final opcode that indicates push data where the remaining buffer is insufficient to satisfy the push size. In this case the script is unconditionally invalid, again excluding the tapscript prevalid state. If the script is not evaluated (e.g. coinbase input) then these conditions are never evaluated.

Note that this short-circuit also captures an initial program stack overflow (`bip342`), and witness stack overflow (`bip342`/`bip141`), and script size limit (`!bip342`).

Incidentally, the `roller` state indicates that the script contains at least one `op_roll` code. This allows the caller to evaluate the script using a `list` stack, as opposed to a `vector` stack. Vector is more efficient when not rolling, but rolling a vector is `O(n)` in stack size, and a `m` instances of `op_roll` produces quadratic evaluation code `O(m * n)`. Given the associated vector memory allocation this is extremely costly. Use of a list stack returns this to linear computational complexity, This condition is evaluated at the instantiation of the script interpreter:

```cpp
code transaction::connect_input(const context& ctx, const input_iterator& it,
    const signatures& capture) const NOEXCEPT
{
    using namespace machine;
    const auto& input = **it;

    if (!input.prevout)
        return error::missing_previous_output;

    // TODO: evaluate performance tradeoff.
    if (input.is_roller() || input.prevout->script().is_roller())
    {
        // Evaluate rolling scripts with linear search but constant erase.
        return interpreter<linked_stack>::connect(ctx, *this, it, capture);
    }

    // Evaluate non-rolling scripts with constant search but linear erase.
    return interpreter<contiguous_stack>::connect(ctx, *this, it, capture);
}
```

KY-U:
> > @evoskuil Could you do a final look here?
> 
> Sure, the short-circuit conditions are more extensive than the summary. During script parse each opcode is tested against three criteria. Notably this excludes a short-circuit in the case where the first opcode is `reserved`. Reserved includes `op_return` and a large number of other codes, though this set is reduced (most of it moved into `invalid`) when executing a tapscript. There is no significant performance benefit to short-circuiting the first opcode, since it is the first tested during script execution. However any `reserved` opcode (including `op_return`) in the first position fails the script.
> 
> The three conditions we test during parse are:
> 
> ```c++
>     bool easier{};
>     bool failer{};
>     bool roller{};
> 
>     for (const auto& op: ops)
>     {
>         easier |= op.is_success();
>         failer |= op.is_invalid();
>         roller |= op.is_roller();
>     }
> ```
> 
> Under `tapscript` a success code anywhere in the script unconditionally succeeds the script. This overrides the second condition, in which any `invalid` code fails the script (unconditional outside of tapscript, and without a success code within tapscript).
> 
> The first line of `script::run()` evaluates this parse:
> 
> ```c++
> TEMPLATE
> code CLASS::
> run() NOEXCEPT
> {
>     if (const auto ec = state::initialize())
>         return ec;
> 
>     for (auto it = state::begin(); it != state::end(); ++it)
>     {
>         const auto& op = *it;
> 
>         if (op.is_oversized())
>             return error::invalid_push_data_size;
> 
>         if (!state::ops_increment(op))
>             return error::invalid_operation_count;
> 
>         if (state::if_(op))
>         {
>             if (const auto ec = run_op(it))
>                 return ec;
> 
>             if (state::is_stack_overflow())
>                 return error::invalid_stack_size;
>         }
>     }
> 
>     if (!state::is_balanced())
>         return error::invalid_stack_scope;
> 
>     return error::script_success;
> }
> ```
> 
> And is shown here:
> 
> ```c++
> TEMPLATE
> INLINE error::script_error_t CLASS::
> initialize() const NOEXCEPT
> {
>     // TODO: nops rule must first be enabled in tests and config.
>     constexpr auto nops = true; ////is_enabled(flags::nops_rule);
>     const auto bip141 = is_enabled(flags::bip141_rule);
>     const auto bip342 = is_enabled(flags::bip342_rule);
> 
>     // Succeed if any success code, overrides all codes and errors [bip342].
>     if (bip342 && script_->is_prevalid())
>         return error::prevalid_script;
> 
>     // Fail if last op is underflow (same behavior as prefail, invalid code).
>     if (script_->is_underflow())
>         return error::invalid_script;
> 
>     // Stack limit (1,000) applied to initial stack [bip342].
>     if (bip342 && is_stack_overflow())
>         return error::invalid_stack_size;
> 
>     // Apply stack element limit (520) to initial witness [bip342|bip141].
>     if ((bip342 || bip141) && witness_ && !chain::witness::is_push_size(*witness_))
>         return error::invalid_witness_stack;
> 
>     // Script size limit (10,000) [0.3.7+], removed [bip342].
>     if (!bip342 && nops && script_->is_oversized())
>         return error::invalid_script_size;
> 
>     // Fail if any op invalid (invalid codes reduced in tapscript).
>     // Should be after underflow check since underflow is also an invalid op.
>     // Promoted success codes are not reachable here due to is_prevalid above.
>     // So only op_verif/op_vernotif (unpromoted invalids) are caught here for
>     // tapscript, otherwise is_prevalid bypassed and all invalids caught here.
>     if (script_->is_prefail())
>         return error::prefail_script;
> 
>     return error::script_success;
> }
> ```
> 
> These conditions short-circuit the script with success/failure based on the first two accumulated parse states:
> 
> ```c++
>     if (bip342 && script_->is_prevalid())
>         return error::prevalid_script;
>         ...
>     if (script_->is_prefail())
>         return error::prefail_script;
> ```
> 
> This remaining conditions short-circuit the script with failure. This includes `underflow`. This is set during parse of the script buffer into opcodes (prior to the above parse step). The only invalid parse is a script with a final opcode that indicates push data where the remaining buffer is insufficient to satisfy the push size. In this case the script is unconditionally invalid, again excluding the tapscript prevalid state. If the script is not evaluated (e.g. coinbase input) then these conditions are never evaluated.
> 
> Note that this short-circuit also captures an initial program stack overflow (`bip342`), and witness stack overflow (`bip342`/`bip141`), and script size limit (`!bip342`).
> 
> Incidentally, the `roller` state indicates that the script contains at least one `op_roll` code. This allows the caller to evaluate the script using a `list` stack, as opposed to a `vector` stack. Vector is more efficient when not rolling, but rolling a vector is `O(n)` in stack size, and a `m` instances of `op_roll` produces quadratic evaluation code `O(m * n)`. Given the associated vector memory allocation this is extremely costly. Use of a list stack returns this to linear computational complexity, This condition is evaluated at the instantiation of the script interpreter:
> 
> ```c++
> code transaction::connect_input(const context& ctx, const input_iterator& it,
>     const signatures& capture) const NOEXCEPT
> {
>     using namespace machine;
>     const auto& input = **it;
> 
>     if (!input.prevout)
>         return error::missing_previous_output;
> 
>     // TODO: evaluate performance tradeoff.
>     if (input.is_roller() || input.prevout->script().is_roller())
>     {
>         // Evaluate rolling scripts with linear search but constant erase.
>         return interpreter<linked_stack>::connect(ctx, *this, it, capture);
>     }
> 
>     // Evaluate non-rolling scripts with constant search but linear erase.
>     return interpreter<contiguous_stack>::connect(ctx, *this, it, capture);
> }
> ```

Thank you for the detailed explanation. I updated the comment to make it explicit that this implementation choice is intended to match Bitcoin Core’s `IsUnspendable()` behaviour for the harness, rather than to describe libbitcoin’s script parsing logic itself.

brunoerg (approved):
ACK 34a9793454e5b68adeb8d78c9dc068fcf078d83b

PR #597, opened 2026-07-02
Use block_view for hash calculation in libbitcoin's block deserialization target

### LINKED ISSUE #577: `deserialize_block` on libbitcoin-system is slow
Running this target with `libbitcoin-system` is considerally slower than Core and btcd, worth investigating why.

KY-U:
Fixes #577 

Add `chain::block_view` for hash calculation to better match libbitcoin's behaviour as mentioned in #577.

### DIFF
--- a/modules/libbitcoinsystem/module.cpp
+++ b/modules/libbitcoinsystem/module.cpp
@@ -128,17 +128,49 @@ LibbitcoinSystem::transaction_eval(std::span<const uint8_t> buffer) const {
 std::optional<std::string>
 LibbitcoinSystem::deserialize_block(std::span<const uint8_t> buffer) const {
   data_chunk data(buffer.begin(), buffer.end());
-  chain::block blk{data, true};
+  chain::block_view blk_view{std::move(data), true};
 
-  // Same rationale as transaction_eval: is_valid() only checks
-  // deserialization. check() applies the consensus-level checks that match
-  // bitcoin core's CheckBlock, so we don't return a hash for blocks the
-  // reference rejects.
+  if (!blk_view.is_valid())
+    return std::nullopt;
+
+  // Checks if block is malleated or merkleroot is invalid.
+  if (blk_view.identify())
+    return "0";
+
+  // Checks witness-commitment under BIP141
+  chain::context ctx{};
+  ctx.flags = chain::flags::bip141_rule;
+  if (blk_view.identify(ctx))
+    return "0";
+
+  // Full object reconstruction after identity checks passed.
+  const data_slice blk_data{buffer.data(), buffer.data() + buffer.size()};
+  chain::block blk{blk_data, true};
+
   if (!blk.is_valid())
     return std::nullopt;
-  if (blk.check())
+
+  // Attach hashes already computed by block_view so block checks do not
+  // recompute them through transaction/header serialization.
+  blk.header_ptr()->set_hash(blk_view.hash());
+
+  const auto &tx_views = blk_view.views();
+  const auto &txs = blk.transactions_ptr();
+  for (size_t i = 0; i < txs->size(); ++i) {
+    const auto &tx = (*txs)[i];
+
+    tx->set_nominal_hash(tx_views[i].hash(true));
+
+    if (tx->is_segregated())
+      tx->set_witness_hash(tx_views[i].hash(true));
+  }
+
+  // Run full block check, skiping identity
+  // checks already performed by block_view.
+  if (blk.check(false))
     return "0";
-  return encode_hash(blk.hash());
+
+  return encode_hash(blk_view.hash());
 }
 
 std::optional<std::string>


modules/libbitcoinsystem/module.cpp
```diff
@@ -128,17 +128,52 @@ LibbitcoinSystem::transaction_eval(std::span<const uint8_t> buffer) const {
...
+    if (tx->is_segregated())
+      tx->set_witness_hash(tx_views[i].hash(true));
+  }
+
+  // Run full block checks, skiping identity
+  // checks already performed by block_view.
+  if (blk.check(false))
+    return "0";
+
+  if (blk.check(ctx, false))
```
brunoerg: Why does it have to call `blk.check` twice here?

modules/libbitcoinsystem/module.cpp
```diff
@@ -128,17 +128,52 @@ LibbitcoinSystem::transaction_eval(std::span<const uint8_t> buffer) const {
 std::optional<std::string>
 LibbitcoinSystem::deserialize_block(std::span<const uint8_t> buffer) const {
   data_chunk data(buffer.begin(), buffer.end());
-  chain::block blk{data, true};
+  chain::block_view blk_view{std::move(data), true};
+
+  if (!blk_view.is_valid())
+    return std::nullopt;
```
erickcestari, line 134: Why are we returning a `nullopt` here? We usually have it well documented so that we know exactly why we are skipping it, because otherwise we might miss bugs.

modules/libbitcoinsystem/module.cpp
KY-U: [`block::check(bool identity)`](https://github.com/libbitcoin/libbitcoin-system/blob/master/src/chain/block.cpp#L895) performs context-free validation such as empty block, size limit and coinbase. Meanwhile, [`block::check(const context &ctx, bol identity)`](https://github.com/libbitcoin/libbitcoin-system/blob/master/src/chain/block.cpp#L924) performs context-dependent validation, it uses fork activation from `ctx` to make validations such as BIP141's weight and witness commitment rules.

modules/libbitcoinsystem/module.cpp
KY-U: I've included nullopt return on invalid block data to match [core's nullopt return](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/modules/bitcoin/module.cpp#L377) on invalid block deserialization. If we consider the core module as the reference, it seems that "0" is only returned when deserialization succeded, but the block failed semantic validation, such as `CheckBlock()`.

It could be interpreted that blocks that failed to deserialize at all provide no real meaningfull result for that input and should be skipped, while "0" is a real result that can be compared against modules, as returning "0" on parse failure would treat malformed bytes the same as well-formed but invalid block.

However:
1. Since there is no documentation on the core module for that particular `return std::nullopt`, I'm not entirely sure my interpretation is correct.
2. It could be argued that, since malformed bytes should be invalid on every module anyway, the target could indeed return "0" on parse failure.

@brunoerg , could you perhaps clarify the undocumented returns of `std::nullopt` on core's module for us? There are also similar cases in [`script_eval`](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/modules/bitcoin/module.cpp#L296), [`verify_script`](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/modules/bitcoin/module.cpp#L311-L315), [`psbt_parse`](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/modules/bitcoin/module.cpp#L512) and [`cmpctblocks_parse`](https://github.com/bitcoinfuzz/bitcoinfuzz/blob/v2/modules/bitcoin/module.cpp#L639-L650) targets.

modules/libbitcoinsystem/module.cpp
brunoerg: I didn't take a look at libbitcoin's code, but passing the `ctx` will make `check` to not perform context-free validation? I mean, you're saying that `check(false)` does A and `check(ctx, false)` does B. I just want to make sure that `check(ctx, false)` doesn't do A + B.

modules/libbitcoinsystem/module.cpp
brunoerg: We usually use `std::nullopt` to skip cases that we don't want comparisons. I don't remember exactly why "0" or `std::nullopt` was chosen in every single case. I think there were some nuances of Core and other implementations during deserialization that were not worth to take in consideration. But would need to check again.

modules/libbitcoinsystem/module.cpp
KY-U: Indeed, `check(false)` and `check(ctx, false)` do not perform redundant operations:
```c++
code block::check(bool identity) const NOEXCEPT
{
    // type64 malleated is a subset of first_not_coinbase.
    // type32 malleated is a subset of is_internal_double_spend.
    if (is_empty())
        return error::empty_block;
    if (is_oversized())
        return error::block_size_limit;
    if (is_first_non_coinbase())
        return malleated_or(error::first_not_coinbase);
    if (is_extra_coinbases())
        return error::extra_coinbases;
    if (is_forward_reference())
        return error::forward_reference;
    if (is_internal_double_spend())
        return malleated_or(error::block_internal_double_spend);
    if (identity && is_invalid_merkle_root())
        return error::invalid_transaction_commitment;

    return check_transactions();
}
```
```c++
code block::check(const context& ctx, bool identity) const NOEXCEPT
{
    const auto bip141 = ctx.is_enabled(bip141_rule);
    const auto bip34 = ctx.is_enabled(bip34_rule);
    const auto bip50 = ctx.is_enabled(bip50_rule);

    if (bip141 && is_overweight())
        return error::block_weight_limit;
    if (bip34 && is_invalid_coinbase_script(ctx.height))
        return error::coinbase_height_mismatch;
    if (bip50 && is_hash_limit_exceeded())
        return error::temporary_hash_limit;
    if (bip141 && identity && is_invalid_witness_commitment())
        return error::invalid_witness_commitment;

    return check_transactions(ctx);
}
```
`check_transactions()` and `check_transactions(ctx)` also differ, they loop trough all transactions calling `tx->check()` and `tx->check(ctx)` respectively, which perform different checks:
```c++
code transaction::check() const NOEXCEPT
{
    const auto coinbase = is_coinbase();

    if (is_empty())
        return error::empty_transaction;
    if (coinbase && is_invalid_coinbase_size())
        return error::invalid_coinbase_script_size;
    if (!coinbase && is_null_non_coinbase())
        return error::previous_output_null;

    return error::transaction_success;
}
```
```c++
code transaction::check(const context& ctx) const NOEXCEPT
{
    const auto bip113 = ctx.is_enabled(bip113_rule);

    if (is_absolute_locked(ctx.height, ctx.timestamp, ctx.median_time_past, bip113))
        return error::absolute_time_locked;

    return error::transaction_success;
}
```
However upon reviewing core's target, I've noticed `ContextualCheckBlock()` isn't called, which performs block weight validation besides witness malleation: `CheckWitnessMalleation()` and `GetBlockWeight(block) > MAX_BLOCK_WEIGHT`. So I've removed `blk.check(ctx, false)` from the libbitcoin target.

I have kept `block_view::identify(ctx)` since core's `IsBlockMutated(block, /*check_witness_root=*/true)` does validates the witness commitment against the witness merkle root, but doesn't rejects overweight blocks, which `blk.check(ctx)` does.

brunoerg (approved):
code review ACK cf531611d7ac99035e8658078af611257fc13dda

PR #599, opened 2026-07-07
feat: add support for psbtv2

brunoerg:


brunoerg:
100% vibe-coded, 0% reviewed.

erickcestari:
I'll start the review today!

modules/rustpsbt/rust_psbt_lib/src/lib.rs
```diff
@@ -22,50 +23,208 @@ pub unsafe extern "C" fn rust_psbt_free_c_string(ptr: *mut c_char) {
...
+    value: i64,
+    script_hex: String,
+}
+
+fn format_result(lock_time: u32, inputs: &[InputSummary], outputs: &[OutputSummary]) -> String {
+    let mut result = String::new();
+
+    result.push_str(&format!("lt={};", lock_time));
+    result.push_str(&format!("in={};", inputs.len()));
+    result.push_str(&format!("out={};", outputs.len()));
```
erickcestari, line 67: 6782ff0718bff6c7ec6c0080f44bc9b6913f1877

It's probably a follow-up. For me, though, these values could be more self-describing.

```suggestion
    result.push_str(&format!("lock_time={};", lock_time));
    result.push_str(&format!("inputs={};", inputs.len()));
    result.push_str(&format!("outputs={};", outputs.len()));
```

modules/rustpsbt/rust_psbt_lib/src/lib.rs
```diff
@@ -22,50 +23,208 @@ pub unsafe extern "C" fn rust_psbt_free_c_string(ptr: *mut c_char) {
...
         }
-        Err(_) => std::ptr::null_mut(),
+    }
+    true
+}
+
+fn try_parse_v2(data: &[u8]) -> Option<String> {
+    if !v2_counts_are_plausible(data) {
+        return None;
+    }
```
erickcestari, line 248: 6782ff0718bff6c7ec6c0080f44bc9b6913f1877

nit: could add an TODO here to remove this in the future.

modules/rustpsbt/rust_psbt_lib/src/lib.rs
```diff
@@ -22,50 +23,197 @@ pub unsafe extern "C" fn rust_psbt_free_c_string(ptr: *mut c_char) {
...
+#[no_mangle]
+pub unsafe extern "C" fn rust_psbt_psbt_parse(data: *const u8, len: usize) -> *mut c_char {
+    let data_slice = slice::from_raw_parts(data, len);
+
+    let result = try_parse_v0(data_slice).or_else(|| try_parse_v2(data_slice));
+
+    match result {
+        Some(result) => str_to_c_string(&result),
+        None => std::ptr::null_mut(),
     }
```
erickcestari: 6782ff0718bff6c7ec6c0080f44bc9b6913f1877

Are we going to skip all the errors possible?

modules/rustpsbt/rust_psbt_lib/src/lib.rs
```diff
@@ -22,50 +23,197 @@ pub unsafe extern "C" fn rust_psbt_free_c_string(ptr: *mut c_char) {
...
+
+fn try_parse_v2(data: &[u8]) -> Option<String> {
+    if !v2_counts_are_plausible(data) {
+        return None;
+    }
+    let psbt = PsbtV2::deserialize(data).ok()?;
+    // Conflicting per-input lock time requirements (BIP-370) are not a
+    // meaningful successful parse; treat like any other parse failure so it
+    // is skipped rather than compared, mirroring the Bitcoin Core module.
+    let lock_time = psbt.determine_lock_time().ok()?.to_consensus_u32();
```
erickcestari: We should probably return an specific error to skip only this case.

modules/rustpsbt/rust_psbt_lib/src/lib.rs
```diff
@@ -22,50 +23,197 @@ pub unsafe extern "C" fn rust_psbt_free_c_string(ptr: *mut c_char) {
...
+        .collect();
+
+    let outputs: Vec<OutputSummary> = psbt
+        .outputs
+        .iter()
+        .map(|psbt_output| OutputSummary {
+            value: psbt_output.amount.to_sat() as i64,
+            script_hex: psbt_output.script_pubkey.to_hex_string(),
+        })
+        .collect();
```
erickcestari, line 305: I feel that we could compare more fields from the PSBT than just some fields from the input, output, and locktime. Maybe also a follow-up.

erickcestari (changes requested):
Nice. I really liked the custom mutator and on my local tests it increased considerably the code coverage.

Theoretically with these checks in the driver.cpp we will never know if an implementation is failing to parse an psbt and other not. Should we keep it?

```cpp
    // Treat malformed parses consistently across modules. For this target, an
    // empty result or a zero-input/zero-output transaction is not a meaningful
    // successful parse and should not participate in differential comparison.
    if (res->empty() || res->find("in=0") != std::string::npos ||
        res->find("out=0") != std::string::npos)
      continue;
```

modules/bitcoin/module.cpp
```diff
@@ -526,10 +570,13 @@ Bitcoin::psbt_parse(std::span<const uint8_t> buffer) const {
...
-    const auto lt{psbt.fallback_locktime.has_value()
-                      ? std::to_string(psbt.fallback_locktime.value())
-                      : ""};
-    result += "lt=" + lt + ";";
+    const std::optional<uint32_t> lt{DeterminePSBTLockTime(psbt)};
+    if (!lt.has_value()) {
+      // Conflicting per-input lock time requirements (BIP-370); not a
+      // meaningful successful parse.
+      return std::string{};
+    }
```
KY-U, line 581: It isn't clear to me why transactions with conflicting per-input lock time requirements would be a "not meaningfull successfull parse". Isn't it valuable to confirm that different modules are all rejecting transactions with conflicting inputs?

KY-U (commented):
While I did come across a point in the PR I would like to discuss, which I've left a comment for, I've noticed that my concerns in regard to the internal logic of which inputs are "meaningfull" enough to include in the comparison process or not, is part of a general lack of explicit semantic contract between the harness and targets. I've opened an [issue](https://github.com/bitcoinfuzz/bitcoinfuzz/issues/603) to discuss that point directly and not clutter the review process.

modules/rustpsbt/rust_psbt_lib/src/lib.rs
brunoerg: Yes, would be better. I think we were trying to avoid a huge final string, but I agree a verbose one would be better.

modules/rustpsbt/rust_psbt_lib/src/lib.rs
brunoerg: Done.

modules/rustpsbt/rust_psbt_lib/src/lib.rs
brunoerg: Done.

modules/rustpsbt/rust_psbt_lib/src/lib.rs
brunoerg: I would do it in a follow-up but decided to extend it here. Done.

modules/rustpsbt/rust_psbt_lib/src/lib.rs
brunoerg: No, just added some cases to not skip.

modules/bitcoin/module.cpp
brunoerg: Good catch, I just fixed by having all PSBTv2 modules return a non-empty sentinel ("CONFLICTING_LOCKTIME") instead of "", since driver.cpp was silently skipping empty results from comparison entirely, so this case is now actually verified across modules.

.github/workflows/workflow.yml
KY-U: Since the `workflow.yml` is already being changed, could we remove this module addition that remained from [#602](https://github.com/bitcoinfuzz/bitcoinfuzz/commit/048384965e270fe98ca6bcef57276c3bc46892f4#diff-126bf89616b7daa3d14ebc882ad18666aaf1c3dae888c4ba306a66ec80758bc1R118)?

KY-U:
Ran for about 1 hour and ran into no issues besides a few memory leaks mentioned on #612, which probably aren't related to this PR, since they also happen on the v2 branch.

.github/workflows/workflow.yml
brunoerg: Done in #614

erickcestari:
Is this ready for review @brunoerg ?

brunoerg:
> Is this ready for review @brunoerg ?

Yes.

erickcestari:
@brunoerg For some reason your commits are being marked as `unverified`

modules/rustpsbt/rust_psbt_lib/src/lib.rs
```diff
@@ -22,50 +23,307 @@ pub unsafe extern "C" fn rust_psbt_free_c_string(ptr: *mut c_char) {
...
+    match try_parse_v2(data_slice) {
+        Ok(result) => str_to_c_string(&result),
+        // Conflicting per-input lock time requirements (BIP-370) is a
+        // well-defined "reject" outcome, not a generic parse failure. Use a
+        // non-empty sentinel so it's actually compared across modules (the
+        // driver's PSBTParseTarget skips empty results from comparison
+        // entirely) rather than silently opted out, mirroring the other
+        // PSBTv2-aware modules.
+        Err(TryParseV2Error::ConflictingLockTime) => str_to_c_string("CONFLICTING_LOCKTIME"),
+        Err(TryParseV2Error::Invalid) => std::ptr::null_mut(),
```
erickcestari, line 327: For a followup: We are skipping the cases where a implementations successfully decoded an PSBT and others not. We might want to crash those scenarios to find missing/wrong verification against each other.

erickcestari (approved):
ACK e12fbb70ca40ca0acc174acd20a8a1e3899382ef

PR #600, opened 2026-07-07
feat: add AES-256-CBC differential target

### LINKED ISSUE #321: Add target for AES
For reference: https://github.com/bitcoin/bitcoin/issues/27548

Opening this issue to discuss if we could/should have this target here.

erickcestari:
Adds an `aes256_cbc` target that compares AES-256-CBC encryption and decryption between:

- **bitcoin** — Bitcoin Core's `src/crypto/aes.{h,cpp}` (ctaes)
- **rustcryptoaes** (new module) — the RustCrypto `aes` + `cbc` crates behind a C FFI shim (uses by default AES-NI if the CPU supports it)

Input is split as `key(32) || iv(16) || pad flag(1) || data(≥1)`. Each module encrypts and decrypts the data and returns `enc=<hex|ERR> dec=<hex|ERR>`; the driver asserts the outputs match. Decrypting raw fuzzer data exercises Core's constant-time PKCS#7 padding validation against RustCrypto's.

closes #321


modules/rustcryptoaes/aes_lib/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+type Aes256CbcEnc = cbc::Encryptor<aes::Aes256>;
+type Aes256CbcDec = cbc::Decryptor<aes::Aes256>;
+
+const BLOCK_SIZE: usize = 16;
+const KEY_SIZE: usize = 32;
+
+fn cbc_encrypt(key: &[u8], iv: &[u8], pad: bool, data: &[u8]) -> String {
+    // Mirror Bitcoin Core's CBCEncrypt: empty input always fails, and without
+    // padding the input must be a whole number of blocks.
+    if data.is_empty() || (!pad && data.len() % BLOCK_SIZE != 0) {
```
brunoerg: nit: I think `data` will never be empty because we avoid it in `driver.cpp`.

driver.cpp
```diff
@@ -769,6 +769,35 @@ void Driver::Musig2KeyAggTarget(std::span<const uint8_t> buffer) const {
...
 
+void Driver::Aes256CbcTarget(std::span<const uint8_t> buffer) const {
+  FuzzedDataProvider provider(buffer.data(), buffer.size());
+  // key(32) + iv(16) + pad flag(1) + at least one data byte. Empty data is
+  // excluded because Bitcoin Core's CBC routines reject it while other
+  // implementations accept it, which is not an interesting difference.
+  if (buffer.size() < 50)
+    return;
+
+  std::vector<uint8_t> key = provider.ConsumeBytes<uint8_t>(32);
```
brunoerg: nit: Bitcoin Core uses a function called `ConsumeFixedLengthByteVector` to create the key and iv. I think it's a function from Core itself, not from FuzzedDataProvider. We could keep as is for now, but would be interesting to have a similar one in bitcoinfuzz. See:

```cpp
/**
 * Returns a byte vector of specified size regardless of the number of remaining bytes available
 * from the fuzzer. Pads with zero value bytes if needed to achieve the specified size.
 */
template<typename B = uint8_t>
[[nodiscard]] inline std::vector<B> ConsumeFixedLengthByteVector(FuzzedDataProvider& fuzzed_data_provider, const size_t length) noexcept
{
    static_assert(sizeof(B) == 1);
    auto random_bytes = fuzzed_data_provider.ConsumeBytes<B>(length);
    random_bytes.resize(length);
    return random_bytes;
}
```

brunoerg: It's basically a resize after consuming bytes.

brunoerg (approved):
lgtm ACK d7f06f0dbb6e133530966a3bf3b471d9b04fadeb

any nits can be addressed in a follow-up PR, if so.

erickcestari:
I've added the util and added a small commit to remove unused variables from scope. This can be merged after CI success.

brunoerg (approved):
reACK 91d9774e387d6939fd664475d1c86274ce8d27f7

PR #602, opened 2026-07-14
feat: add kernel_block_check differential fuzz target

brunoerg:
Runs btck::Block::Check (context-free block validation: size limits, coinbase structure, sigops, and optional PoW/merkle-root checks) across bitcoinkernel, bitcoinkernelvariant, rustbitcoinkernel, and pybitcoinkernel, comparing validation mode and result codes to catch consensus-logic divergence between bindings, not just parsing bugs.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -115,6 +115,7 @@ jobs:
           -DEMBIT \
           -DPYBITCOINKERNEL \
           -DBITCOINKERNEL \
+          -DBITCOINKERNEL_VARIANT \
           -DLND \
           -DLDK \
           -DNLIGHTNING \
@@ -198,6 +199,10 @@ jobs:
           - corpus: "kernel_transaction"
             target: "kernel_transaction"
             cxxflags: "-DPYBITCOINKERNEL -DRUSTBITCOINKERNEL -DBITCOINKERNEL"
+            asan_options: "detect_leaks=0:detect_container_overflow=0"
+          - corpus: "kernel_block_check"
+            target: "kernel_block_check"
+            cxxflags: "-DBITCOINKERNEL -DPYBITCOINKERNEL -DRUSTBITCOINKERNEL"
             asan_options: "detect_leaks=0:detect_container_overflow=0"
           - corpus: "cmpctblocks_parse"
             target: "cmpctblocks_parse"
--- a/modules/bitcoinkernel/module.cpp
+++ b/modules/bitcoinkernel/module.cpp
@@ -1,6 +1,9 @@
 #include "module.h"
 #include <kernel/bitcoinkernel_wrapper.h>
 
+#include <array>
+#include <cassert>
+#include <cstddef>
 #include <cstdint>
 #include <cstring>
 #include <iomanip>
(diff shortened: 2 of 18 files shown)


modules/bitcoinkernel/module.cpp
```diff
@@ -31,6 +33,40 @@ std::string hash_bytes_to_hex(const uint8_t *bytes, int length) {
...
+    return "testnet";
+  case btck::ChainType::TESTNET_4:
+    return "testnet4";
+  case btck::ChainType::SIGNET:
+    return "signet";
+  case btck::ChainType::REGTEST:
+    return "regtest";
+  }
+
+  return "unknown";
```
KY-U: If `decode_chain_type` always returns a valid chain with `chain_types[ value % chain_types.size()]`, then `chain_type_to_string`'s `unknown` default will never be returned. Is that case desirable?

modules/bitcoinkernel/module.cpp
```diff
@@ -99,6 +135,51 @@ char *libbitcoinkernel_block(std::span<const uint8_t> buffer) {
...
+    result.append(hash_bytes_to_hex((const uint8_t *)block_hash_bytes.data(),
+                                    static_cast<int>(block_hash_bytes.size())));
+    result.append(";txs=");
+    result.append(std::to_string(block.CountTransactions()));
+    result.push_back(';');
+    return strdup(result.c_str());
+  } catch (...) {
+    result.append("err=exception;");
+    return strdup(result.c_str());
+  }
```
KY-U, line 181: The current expected return for exceptions on the other kernel targets is `strdup("0");`. I think it would be good to standardize that if possible, or is there a specific reason to change that for this target?

modules/bitcoinkernel/module.cpp
brunoerg: Yes, I don't it's super useful to choose an invalid one. Perhaps I could change this return to an `assert(false)` since we're expecting to always have a valid chain type.

modules/bitcoinkernel/module.cpp
brunoerg: Hmm, interesting! I don't  think "0" is great to be a standard. For example, "0" works well for `kernel_transaction` because the entire input buffer is the thing being parsed. So, on failure there's nothing else to say, so a bare "0"  is sufficient. However, when the input is structured ,and for `kernel_block_check` the first two bytes are decoded by decode_chain_type/decode_block_check_flags before the try block even starts, showing something like "chain=signet;flags=2;err=exception;" would be better for triage.

KY-U:
Will re-review today!

.github/workflows/workflow.yml
```diff
@@ -199,6 +200,10 @@ jobs:
             target: "kernel_transaction"
             cxxflags: "-DPYBITCOINKERNEL -DRUSTBITCOINKERNEL -DBITCOINKERNEL"
             asan_options: "detect_leaks=0:detect_container_overflow=0"
+          - corpus: "kernel_block_check"
+            target: "kernel_block_check"
+            cxxflags: "-DBITCOINKERNEL -DBITCOINKERNEL_VARIANT -DPYBITCOINKERNEL -DRUSTBITCOINKERNEL"
```
KY-U: When the bitcoinkernel modules were added, the variant version was intentionally left out of CI workflow executions. Since it's mostly a mirror of the standard version, I believe there isn't any great bonus gained from adding it to the CI, as it just increases its time by compiling core again.

KY-U (commented):
Found no blocking change. The added target seems consistent with existing ones for all modules. Just the addition of the `bitcoinkernel_variant`module to the CI workflow seems unnecessary to me, since both versions checkout to master.

.github/workflows/workflow.yml
brunoerg: Ah nice catch, it makes sense.

brunoerg:
>  Just the addition of the bitcoinkernel_variantmodule to the CI workflow seems unnecessary to me, since both versions checkout to master.

Fixed it.

PR #615, opened 2026-07-27
feat: add multi_index target (Boost.MultiIndex vs tmi2)

brunoerg:
Add a new "multi_index" fuzz target that runs an identical container op sequence (insert, emplace, erase, find, count, bounds, modify, clear) against every loaded multi-index implementation and compares a trace of the observable outcomes. The trace only records implementation- independent behavior: hashed-index iteration order and tie order among equivalent non-unique keys are canonicalized or omitted.

The op decoding and trace logic live in a shared template harness (include/bitcoinfuzz/multiindex_harness.h); the two new modules only instantiate their container type with the shared element/key extractors, which satisfy both Boost's KeyFromValue concept and tmi's:

- modules/boostmultiindex: Boost.MultiIndex, pinned via the external/multi_index submodule. external/boost-mp11 is also pinned because multi_index master requires a newer mp11 than system Boost ships; the remaining support headers come from BOOST_ROOT.
- modules/tmi2: tmi2 (theuni/tmi2), a reimplementation intended as a near drop-in replacement for Bitcoin Core's Boost.MultiIndex usage, pinned via the external/tmi2 submodule.


include/bitcoinfuzz/multiindex_harness.h
```diff
@@ -0,0 +1,256 @@
...
+  auto &hashed_unique_view = c.template get<1>();
+  auto &ordered_nonunique_view = c.template get<2>();
+  auto &hashed_nonunique_view = c.template get<3>();
+
+  std::string trace;
+
+  while (fdp.remaining_bytes() > 0) {
+    uint8_t op = fdp.ConsumeIntegralInRange<uint8_t>(0, 9);
+    trace += '|';
+    switch (op) {
```
brunoerg, line 122: We could have `CallOneOf` like Bitcoin Core.

brunoerg:
Ran it overnight and didn't get any other bug expect by three ones I've reported.

brunoerg:
cc @theuni

brunoerg:
Just pushed (ebd2f850a5271def6468318b3e42c7860e4e41f8) adding some single index targets. Still figuring out if I "mitigated" scenarios that would case timeouts.

theuni:
Concept ACK. Thanks for working on this!

I would ask that you hold off until things stabilize a bit, though. tmi2 is not really ready for eyes yet, it's still very much in the hacking/experimenting stage. So fuzzing it is not (yet) a good use of resources.

PR #618, opened 2026-07-28
feat(utreexo/rustreexo): add `stump_update` target

brunoerg:
Drive the full accumulator update cycle across utreexod and rustreexo: add a batch of leaves, prove a fuzzer-chosen subset of them with the full accumulator (Pollard / MemForest), then apply the deletion plus a second add batch to a Stump via that proof. The serialized final stump and accumulator roots are compared across modules, covering proof verification and root recomputation during deletion, which the add-only stump_modify_add target does not reach.

The driver constrains leaf generation to keep known false-positive classes unreachable:
- leaves are 16-byte seeds mirrored to 32 bytes, so libFuzzer's CMP feedback cannot reinject learned parentHash values as leaves (duplicate node hashes break the two libraries differently)
- leaves sharing a 12-byte prefix are skipped because utreexod's position map is keyed by the first 12 bytes of the leaf hash
- all-zero leaves are skipped as before (empty-node marker)

Proof bytes themselves are not compared: utreexod expresses proof targets in a fixed 63-row position space while rustreexo uses tree_rows(num_leaves) space, so only the resulting state is comparable.


brunoerg:
cc @Davidson-Souza @luisschwab

luisschwab:
I'll take a look soon

Davidson-Souza:
> Proof bytes themselves are not compared: utreexod expresses proof targets in a fixed 63-row position space while rustreexo uses tree_rows(num_leaves) space, so only the resulting state is comparable.

I can fix that upstream, we always communicate them as 63-row positions.

brunoerg:
> > Proof bytes themselves are not compared: utreexod expresses proof targets in a fixed 63-row position space while rustreexo uses tree_rows(num_leaves) space, so only the resulting state is comparable.
> 
> I can fix that upstream, we always communicate them as 63-row positions.

Let me know when it's done.

brunoerg:
Any update here @Davidson-Souza ?

Issue #612, opened 2026-07-27
`Bitcoin-S` and `NBitcoin` modules leaking memory on `psbt_parse` target


KY-U:

brunoerg:
What was the command you used to build and run it? It would be easier to reproduce the issue.

brunoerg:
Sorry, it's only for NBitcoin and Bitcoin-S so it's a known issue. It's a false positive coming from LeakSanitizer's interaction with the two managed runtimes (JNI and the .NET one). When both are initialized the memory that is allocated once, lives for the whole process, and is intentionally never freed, so the LSan reports it as "leaked".

KY-U:
> What was the command you used to build and run it? It would be easier to reproduce the issue.

I've built each module individually simply running `make` inside each module. Built bitcoinfuzz with all psbt_parse modules with:
```shell
CC=clang CXX=clang++ \
CXXFLAGS="-DBITCOIN_CORE -DBITCOINS -DBTCD -DEMBIT -DLIBWALLY_CORE -DNBITCOIN -DRUST_PSBT" \
make
```
Then ran into leak memory when running with any of those 2 specific modules:
```shell
MODULES="BITCOINS,NBITCOIN" FUZZ=psbt_parse ./bitcoinfuzz \
  -artifact_prefix=<artifacts-folder> \
  <corpora-folder>
```

PR #619, opened 2026-07-28
Suppress known false-positive leaks regarding nbitcoin and bitcoin-s.

KY-U:
Adds documentation and lsan rules for false-positive leaks mentioned on [#612](https://github.com/bitcoinfuzz/bitcoinfuzz/issues/612#issuecomment-5102252160)

### DIFF
--- a/lsan.supp
+++ b/lsan.supp
@@ -44,3 +44,5 @@ leak:Unsafe_AllocateMemory0
 # JFR (Java Flight Recorder) instruments classes during loading; the transformed
 # bytecode buffer is owned by the JVM class subsystem — not a real leak.
 leak:JfrEventClassTransformer::
+# .NET NativeAOT process-lifetime WaitSubsystem monitor
+leak:S_P_CoreLib_System_Threading_WaitSubsystem___cctor
--- a/lsan.supp
+++ b/lsan.supp
@@ -46,3 +46,6 @@ leak:Unsafe_AllocateMemory0
 leak:JfrEventClassTransformer::
 # .NET NativeAOT process-lifetime WaitSubsystem monitor
 leak:S_P_CoreLib_System_Threading_WaitSubsystem___cctor
+# BitcoinS secp256k1-zkp global verification/signing context, initialized once
+# and retained for the lifetime of the JVM process.
+leak:secp256k1_context_create
--- a/modules/bitcoins/README.md
+++ b/modules/bitcoins/README.md
@@ -9,3 +9,13 @@ cd modules/bitcoins
 make
 export CXXFLAGS="$CXXFLAGS -DBITCOINS"
 ```
+
+## Notes
+
+When fuzzing with BitcoinS, JVM initialization can cause LSan to report
+process-lifetime allocations as leaks. If this happens, use the provided
+suppression file:
+
+```bash
+LSAN_OPTIONS="suppressions=$(pwd)/lsan.supp" MODULES="BITCOINS" FUZZ=<target> ./bitcoinfuzz
+```
--- a/modules/nbitcoin/README.md
+++ b/modules/nbitcoin/README.md
@@ -9,3 +9,13 @@ cd modules/nbitcoin
 make
 export CXXFLAGS="$CXXFLAGS -DNBITCOIN"
 ```
+
+## Notes
+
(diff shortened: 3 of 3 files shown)


lsan.supp
```diff
@@ -44,3 +44,13 @@ leak:Unsafe_AllocateMemory0
 # JFR (Java Flight Recorder) instruments classes during loading; the transformed
 # bytecode buffer is owned by the JVM class subsystem — not a real leak.
 leak:JfrEventClassTransformer::
+# .NET NativeAOT process-lifetime WaitSubsystem monitor
```
brunoerg, line 47: Which OS did you test these suppressions?

KY-U: Arch linux. Did the rules not work in your machine?

brunoerg: Can you test on Ubuntu? This is the main environment of our servers.

erickcestari:
I usually run those target with `ASAN_OPTIONS=detect_leaks=0` (our CI also run in this way) to avoid leaks errors. We should probably document it 😅.

I'm also in arch linux and received this error when trying to run on `bip32_master_keygen` target:

```
bitcoinfuzz on  pr-619 [$?] via C v16.1.1-gcc via 🐍 v3.14.6 
❯ LSAN_OPTIONS="suppressions=$(pwd)/lsan.supp" FUZZ=bip32_master_keygen ./bitcoinfuzz
=== BitcoinFuzz Module Activation Log ===
Target: bip32_master_keygen
Loaded Modules (1):
  [1] BitcoinS
=========================================
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 659544986
INFO: Loaded 1 modules   (2252 inline 8-bit counters): 2252 [0x5650cdf419e8, 0x5650cdf422b4), 
INFO: Loaded 1 PC tables (2252 PCs): 2252 [0x5650cdf422b8,0x5650cdf4af78), 
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
INFO: A corpus is not provided, starting from an empty corpus
#2      INITED cov: 145 ft: 146 corp: 1/1b exec/s: 0 rss: 269Mb

=================================================================
==175534==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 1048800 byte(s) in 1 object(s) allocated from:
    #0 0x5650cde1fe51  (/home/erick/open-source/bitcoinfuzz/bitcoinfuzz+0x294e51) (BuildId: 5d2526b5787aca722dc9d748dc15e2574db163e7)
    #1 0x7b2958411ab4  (/tmp/nativelib-loader_8567205440101336170/libsecp256k1.so+0x11ab4) (BuildId: 2d44b3a1d2843b92cfd8caebba4ea191bf3b45e1)

-----------------------------------------------------
Suppressions used:
  count      bytes template
   2941     859447 libjvm.so
-----------------------------------------------------

SUMMARY: AddressSanitizer: 1048800 byte(s) leaked in 1 allocation(s).
INFO: to ignore leaks on libFuzzer side use -detect_leaks=0.

MS: 1 InsertByte-; base unit: adc83b19e793491b1c6ea0fd8b46cd9f32e592fc
0xeb,0xa,
\353\012
artifact_prefix='./'; Test unit written to ./leak-5cc5204461646a2220432ebc463142545637e033
Base64: 6wo=
```

lsan.supp
KY-U: Ran on Ubuntu 26.04. The suppressions worked, but It is worth noting that is necessary to have the LLVM-symbolizer on PATH or pass it explicitly when running.

In my case, since I'm using clang 21.1.8:
```shell
apt install llvm-21
```
And then add `/usr/lib/llvm-21/bin` to path or run with:
```
ASAN_SYMBOLIZER_PATH="/usr/lib/llvm-21/bin/llvm-symbolizer"
```

KY-U:
I did not get the error when running `bip32_master_keygen` on ubuntu.

brunoerg (approved):
code review ACK be622b75973589ca120ac4a54e9e5e74ded8711c

PR #629, opened 2026-08-11
feat: add new target merkle_root_compute

brunoerg:
Differential fuzzing of merkle root computation over a list of raw 32-byte hashes (internal byte order), comparing both the resulting root (display byte order) and CVE-2012-2459 duplicated-subtree detection.

Module contract: "<root_hex>;mutated=0|1", or the sentinel "REJECTED" when the library refuses to compute a root for the list (rust-bitcoin's TxMerkleNode::calculate_root returns None for mutated lists). The driver compares roots across all modules that produce one and compares mutation detection across every responding module, so "mutated=1" and "REJECTED" count as agreement while still catching a module that silently accepts a mutated list (or rejects a clean one).

Modules: BITCOIN_CORE (ComputeMerkleRoot + mutated out-param), GOCOIN (btc.CalcMerkle, root + mutated), RUST_BITCOIN (TxMerkleNode::calculate_root, rejects mutated lists), BTCD (tree reduction driven wrapper-side via the exported HashMerkleBranches pair combiner, since btcd's merkle entry points only accept []*btcutil.Tx).

Empty lists are skipped: the empty-tree result differs by design (Core returns the zero hash, rust-bitcoin returns None, gocoin panics) and an empty transaction list is consensus-invalid anyway.

- [x] Add it to CI


include/bitcoinfuzz/basemodule.h
```diff
@@ -137,6 +137,15 @@ class BaseModule {
   virtual std::optional<std::string>
   stump_modify_add(const std::vector<std::vector<uint8_t>> &add_hashes) const;
 
+  // Computes the merkle root over a list of raw 32-byte hashes (given in
+  // internal byte order, as serialized in a block). The response is either
+  // "<root_hex>;mutated=0|1" (root in display byte order, plus whether a
+  // CVE-2012-2459-style duplicated subtree was detected) or the sentinel
+  // "REJECTED" when the library refuses to compute a root for the list at
+  // all (e.g. rust-bitcoin returns None for mutated lists).
```
erickcestari: nit:

```suggestion
```

erickcestari (approved):
ACK 8f020d51ca93ed4be9be9ed1b3793901bfcc979f

PR #630, opened 2026-08-11
feat: add new target sighash_compute

brunoerg:
Differential fuzzing of the signature hash computation for one transaction input, covering both legacy (SigVersion::BASE) and segwit v0 (BIP143), as each module's interpreter would compute it for an OP_CHECKSIG execution.

Structured input (FuzzedDataProvider): serialized tx, scriptCode with OP_CODESEPARATORs intact, which separator to truncate after (module-side, using each library's own script tokenizer), a signature blob to remove via FindAndDelete (legacy only), amount, raw unfiltered sighash type, and the sigversion. The driver optionally splices a canonical push of the signature blob into the script so the FindAndDelete path is actually exercised.

Module contract: hex digest in display order, or skip (nullopt) for input classes the library cannot handle. Modules: BITCOIN_CORE (SignatureHash), BTCD (CalcSignatureHash/CalcWitnessSigHash), GOCOIN (Tx.SignatureHash/Tx.WitnessSigHash + AllocVerVars fix), RUST_BITCOIN (SighashCache legacy_signature_hash/p2wsh_signature_hash).

Documented scope decisions:
- btcd's exported sighash API rejects unparseable scripts (checkScriptParses); those inputs skip the module.
- rust-bitcoin's segwit-v0 API normalizes the sighash type through EcdsaSighashType, while Core hashes the raw u32: non-standard sighash bytes are skipped for rust-bitcoin on the v0 axis only (legacy takes a raw u32 there).
- Legacy inputs with unparseable scripts are gated driver-side: libraries knowingly differ on the malformed-tail serialization (Core's serializer emits the truncated push header, gocoin drops it, btcd rejects), a separate divergence class that would drown this target's signal.
- btcd/gocoin/rust-bitcoin do not export their FindAndDelete equivalents (removeOpcodeByData/delSig/none), so the wrappers port Core's exact FindAndDelete semantics; gocoin's native (buggy) delSig is not reachable through its exported API and is therefore out of scope here.
- gocoin wrapper preflights the serialized tx before NewTx, which allocates slices straight from untrusted CompactSize counts (libFuzzer found an OOM via a huge input count).


erickcestari (commented):
Needs rebase

erickcestari:
Needs rebase. 
image

modules/gocoin/gocoin_wrapper/wrapper.go
```diff
@@ -172,6 +173,252 @@ func GocoinMerkleRootCompute(data C.ByteArray) *C.char {
...
+//
+//export GocoinSighashCompute
+func GocoinSighashCompute(txData C.ByteArray, scriptData C.ByteArray, sigData C.ByteArray, inputIndex C.uint32_t, nCodesep C.uint32_t, amount C.uint64_t, sighashType C.uint32_t, isV0 C.int) (res *C.char) {
+	// gocoin can panic on malformed structures; treat a panic as "unsupported
+	// input" so the driver simply skips this module.
+	defer func() {
+		if r := recover(); r != nil {
+			res = nil
+		}
+	}()
```
erickcestari, line 386: Shouldn't we report these crashes?

modules/gocoin/gocoin_wrapper/wrapper.go
```diff
@@ -172,6 +173,252 @@ func GocoinMerkleRootCompute(data C.ByteArray) *C.char {
...
+	var digest []byte
+	if isV0 == 0 {
+		script = gocoinFindAndDelete(script, sig)
+		digest = tx.SignatureHash(script, idx, int32(sighashType))
+	} else {
+		digest = tx.WitnessSigHash(script, uint64(amount), idx, int32(sighashType))
+	}
+	if len(digest) != 32 {
+		return nil
+	}
```
erickcestari, line 415: Is possible that the digest be smaller than 32 bytes? 

```suggestion

```

erickcestari (changes requested):

modules/gocoin/gocoin_wrapper/wrapper.go
brunoerg: I hope not, I can change it to panic in case it happens.

modules/gocoin/gocoin_wrapper/wrapper.go
brunoerg: Yes, better to report.

brunoerg:
Addressed @erickcestari two comments.

erickcestari (approved):
ACK 57737055dbb2915e57887d2fed8f2270157ea0d3

Needs rebase

brunoerg:
Rebased

PR #633, opened 2026-08-11
feat: add new target partial_merkle_tree

brunoerg:
Differential fuzzing of BIP37 partial Merkle tree extraction: parse a serialized CPartialMerkleTree body (nTransactions | hashes | flag bytes) and compare the reconstructed root plus the matched txids and their leaf indices. Rejection is normalized to "REJECT" (Core collapses all failure modes into a zero-hash return) and stream-level malformation to "PARSE_ERR"; trailing garbage is tolerated on both sides (matches Core's DataStream behavior).

This covers the CVE-2012-2459 duplicated-branch check at the SPV-proof layer (the haskoin-core finding class), which merkle_root_compute does not reach: Core rejects when a descended branch's children are identical, and never descends into fully-unmatched subtrees (so duplicated leaves that are never traversed still verify).

Modules: BITCOIN_CORE (CPartialMerkleTree::ExtractMatches; merkleblock.cpp added to the module build) and RUST_BITCOIN (p2p crate's PartialMerkleTree, which ports Core's checks one-to-one including IdenticalHashesFound). btcd and gocoin have no native partial-tree extraction API, so they are intentionally out of scope (a wrapper-side traversal would only test our own reimplementation).

Fuzzing immediately found a harness-level ambiguity, fixed before commit: a 1-transaction tree whose only hash is all-zero yields a legitimate zero root. Core cannot represent that as success (zero hash is its failure sentinel), so rust-bitcoin's Ok(zero) is normalized to "REJECT" as well.

Validated in Docker (linux/amd64): 46 known-answer vectors against an independent Python port of Core's build/extract (valid trees for 1-8 txids with assorted match masks, matched/unmatched duplicated branches, zero nTransactions, too many hashes, unconsumed bits/hashes, truncated and non-canonical CompactSize, padding-bit tolerance, zero-root regression); plus a seeded 300s run: ~895k execs, no mismatches.

### DIFF
--- a/include/bitcoinfuzz/basemodule.h
+++ b/include/bitcoinfuzz/basemodule.h
@@ -145,6 +145,21 @@ public:
   // all (e.g. rust-bitcoin returns None for mutated lists).
   virtual std::optional<std::string>
   merkle_root_compute(const std::vector<std::vector<uint8_t>> &hashes) const;
+  // Parses a BIP37 partial Merkle tree (nTransactions u32 LE | CompactSize n
+  // | n*32-byte hashes | CompactSize m | m flag bytes) and extracts the
+  // matched transactions, emulating CMerkleBlock extraction semantics.
+  // Returns one of:
+  //   "PARSE_ERR"  — the byte stream is malformed (truncated, non-canonical
+  //                  CompactSize, ...)
+  //   "REJECT"     — parsed but the tree is invalid (Core's ExtractMatches
+  //                  returning the zero hash; covers nTransactions==0,
+  //                  too many transactions/hashes, bit/hash underflow during
+  //                  traversal, CVE-2012-2459 duplicated branches, trailing
+  //                  unconsumed bits or hashes)
+  //   "<root_hex>;m=<txid>@<idx>,..." — root and matched txids with their
+  //                  leaf indices, in traversal order (display byte order)
+  virtual std::optional<std::string>
+  partial_merkle_tree(std::span<const uint8_t> buffer) const;
 
   virtual std::optional<std::string>
   bip32_derive_from_path(std::span<const uint8_t> buffer) const;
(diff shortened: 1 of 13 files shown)


include/bitcoinfuzz/basemodule.h
```diff
@@ -137,6 +137,22 @@ class BaseModule {
...
+  // Returns one of:
+  //   "PARSE_ERR"  — the byte stream is malformed (truncated, non-canonical
+  //                  CompactSize, ...)
+  //   "REJECT"     — parsed but the tree is invalid (Core's ExtractMatches
+  //                  returning the zero hash; covers nTransactions==0,
+  //                  too many transactions/hashes, bit/hash underflow during
+  //                  traversal, CVE-2012-2459 duplicated branches, trailing
+  //                  unconsumed bits or hashes)
+  //   "<root_hex>;m=<txid>@<idx>,..." — root and matched txids with their
+  //                  leaf indices, in traversal order (display byte order)
```
erickcestari, line 160: nit:
```suggestion
```

modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
```diff
@@ -370,6 +370,48 @@ pub unsafe extern "C" fn rust_bitcoin_cmpctblocks_parse(
...
+    match pmt.extract_matches(&mut matches, &mut indexes) {
+        Err(_) => str_to_c_string("REJECT"),
+        Ok(root) => {
+            // Core signals every extraction failure with the zero hash, so a
+            // zero root is indistinguishable from rejection in its API (and
+            // is treated as rejection by all its callers); map to the same
+            // sentinel for comparability.
+            if root == bitcoin::TxMerkleNode::from_byte_array([0u8; 32]) {
+                return str_to_c_string("REJECT");
+            }
```
erickcestari, line 431: Is it an issue that should be fixed?

erickcestari (changes requested):
Overall LGTM!

modules/rustbitcoin/rust_bitcoin_lib/src/lib.rs
brunoerg: No, only a mismatch.

erickcestari (approved):
ACK 9824507a0c0c0da81ef9f28cff98cc2928e6c815

Issue #574, opened 2026-06-09
Add Differential Fuzzing Support for Silent Payments (BIP-352)


erickcestari:
Introduce differential fuzzing for Silent Payments by comparing multiple independent BIP-352 implementations.

We can see all the development progress and wallet that implement SP here:

https://github.com/orgs/silent-payments/projects/2/views/5

brunoerg:
What specifically do you think would be great as a target?

luisschwab:
https://github.com/bitcoindevkit/bdk-sp
https://github.com/cygnet3/spdk
https://github.com/BlueWallet/SilentPayments

erickcestari:
secp256k1, when this PR gets merged: https://github.com/bitcoin-core/secp256k1/pull/1765

brunoerg:
@yan-pi what do you think would be great as a target?

yan-pi:
Exploring the Silent Payments Discord, I discovered [this website](https://silentpayments.xyz/docs/wallets/), which tracks the Silent Payments implementations across multiple wallets, and also [other projects](https://silentpayments.xyz/docs/developers/).

was speaking with @IsaqueFranklin, and perhaps he can offer another point of view on what aspects the implementations should cover or not

KY-U:
@eunovo

Eunovo:
A good place to start is fuzzing SP transaction scanning. Some features I'd expect to see in a differential harness:

- **Random input UTXO outpoints**
- **Random input types**, mixing BIP352-eligible inputs (P2TR, P2WPKH, P2SH-P2WPKH, legacy P2PKH-style eligible inputs) with non-eligible ones.
- **Random `k` values** — normally each `k` is used once per output, but repeated `k` values should be fuzzed deliberately. There was a [recent discovery](https://github.com/bitcoin-core/secp256k1/pull/1765#issuecomment-5020395054) in the secp256k1 silentpayments module review about exactly this: if `k` isn't properly incremented across multiple outputs to the same recipient, implementations can silently diverge or produce colliding outputs. This is a great differential-fuzzing target because a single implementation's unit tests won't surface disagreement between implementations on how `k` is supposed to behave.
- **Random labels**, biased toward a low number of distinct labels (0-3, say) rather than uniformly random, since label support is thin/inconsistent across implementations.

All of the above apply equally to fuzzing SP transaction *creation*, not just scanning.

For implementations with sending capability, two additional target shapes are useful:

[...]

PR #634, opened 2026-08-12
Add silentpayments_create_outputs target with bdk-sp, BlueWallet and spdk modules

erickcestari:
Adds a differential target for the BIP-352 sender side, comparing libsecp256k1's
`silentpayments` module against three implementations:

- `bdksp` ([bdk-sp](https://github.com/bitcoindevkit/bdk-sp)), Rust
- `bluewalletsp` ([BlueWallet/SilentPayments](https://github.com/BlueWallet/SilentPayments)), TypeScript, run on node over a pipe
- `spdk` ([cygnet3/spdk](https://github.com/cygnet3/spdk)), Rust

The target fuzzes 1-8 input keys with a taproot flag each, and 1-8 recipients
whose scan keys are drawn from a smaller pool so recipients group and the k
counter increments within a group. Modules return the recipient-ordered outputs
as hex, or an `INVALID_SECKEY` / `CREATE_FAIL` sentinel, so an accept-vs-reject
disagreement is compared rather than skipped.

All three implementations reject an input key sum whose intermediate is zero
while the total is not, where BIP-352 constrains only the total. bdk-sp tracks
it upstream with an ignored test vector; the other two are also reported as github issues.
Each wrapper detects that shape and skips it, marked for removal once fixed.

partially address: #574


brunoerg:
I'm going to review it tomorrow.

brunoerg (approved):
ACK 09d59432ca50ee81e94db5c7c83d709ed3a806fe

brunoerg:
Minor fixes and improvements can be done in a follow-up PR.

PR #635, opened 2026-08-12
feat: add new target pubkey_parse

brunoerg:
Differential fuzzing of SEC1 public key parsing: each module parses the input bytes as a secp256k1 public key and returns "ERR" (reject) or "OK:<canonical 33-byte compressed encoding>" (accept), so both accept-vs-reject and accepted-point divergences surface. This is the class of the audited-fork findings (x/y >= p aliased encodings verifying, hybrid-pubkey acceptance, off-curve acceptance): an input like 03 || (p+1) is rejected by the reference parser (fe_set_b32_limit) but accepted by forks that drop the range check.


driver.cpp
```diff
@@ -503,6 +503,33 @@ void Driver::PrivateToPublicKeyTarget(std::span<const uint8_t> buffer) const {
...
 
+void Driver::PubkeyParseTarget(std::span<const uint8_t> buffer) const {
+  std::optional<std::string> last_response{std::nullopt};
+  std::string last_module_name;
+
+  // Known SEC1 encoding-policy splits among the reference implementations,
+  // skipped per-module to keep the differential meaningful:
+  const bool is_hybrid =
+      buffer.size() == 65 && (buffer[0] == 0x06 || buffer[0] == 0x07);
+  const bool is_compact = buffer.size() == 33 && buffer[0] == 0x05;
```
KY-U: Since the decision of skipping hybrid and compact inputs is necessary only for `RUST_K256` currently, could this segment be moved inside the `#ifdef RUST_K256` directive?

modules/bitcoin/module.cpp
```diff
@@ -393,6 +393,25 @@ Bitcoin::deserialize_block(std::span<const uint8_t> buffer) const {
...
+                                 buffer.data(), buffer.size())) {
+    return "ERR";
+  }
+  size_t out_len = 33;
+  std::vector<uint8_t> out(out_len);
+  if (!secp256k1_ec_pubkey_serialize(secp256k1_context_static, out.data(),
+                                     &out_len, &pubkey,
+                                     SECP256K1_EC_COMPRESSED)) {
+    return std::nullopt;
+  }
```
KY-U: I understand that, after the input has been successfully parsed, an error on this step would mean an internal serialization error, not an invalid input, hence the `nullopt` return instead of `ERR`. Still, wouldn't it be desirable to identify (maybe with a distinct internal error flag) and investigate when an error does happen on this stage? Or document the justification for the `nullopt` return?

modules/btcd/btcd_wrapper/libbtcd_wrapper.h
```diff
@@ -99,6 +99,12 @@ extern char* BTCDAddrv2(ByteArray addrv2Data);
 extern char* BTCDScriptAsm(ByteArray scriptData);
 extern char* BTCDDesBlock(ByteArray scriptData);
 extern void BTCDFreeString(char* ptr);
+
+// BTCDPubkeyParse parses a SEC1-encoded public key via btcec.ParsePubKey and
+// returns the canonical compressed encoding ("OK:<hex>"), or "ERR" on
+// rejection.
+//
```
KY-U: This same comment is present in `modules/btcd/btcd_wrapper/wrapper.go`. Perhaps this one can be removed?

modules/btcd/module.cpp
```diff
@@ -114,6 +114,21 @@ Btcd::transaction_eval(std::span<const uint8_t> buffer) const {
...
+
+std::optional<std::string>
+Btcd::pubkey_parse(std::span<const uint8_t> buffer) const {
+  ByteArray data{
+      .data = reinterpret_cast<char *>(const_cast<uint8_t *>(buffer.data())),
+      .length = static_cast<int>(buffer.size())};
+
+  char *result = BTCDPubkeyParse(data);
+  if (!result)
+    return std::nullopt;
```
KY-U: The same point about perhaps documenting or modifying the `nullopt` return.

modules/decredsecp256k1/decredsecp256k1_wrapper/libdecredsecp256k1_wrapper.h
```diff
@@ -99,6 +99,11 @@ extern char* DecredSignDER(ByteArray privKeyData, ByteArray hashData);
 extern int DecredSignVerify(ByteArray privKeyData, ByteArray hashData, ByteArray signData);
 extern char* DecredECDH(ByteArray privKeyData, ByteArray pubKeyData);
 
+// DecredPubkeyParse parses a SEC1-encoded public key and returns the
+// canonical compressed encoding ("OK:<hex>"), or "ERR" on rejection.
+//
```
KY-U, line 104: The same comment is also present in `modules/decredsecp256k1/decredsecp256k1_wrapper/go.mod`

modules/decredsecp256k1/module.cpp
```diff
@@ -100,5 +100,19 @@ Decred_secp256k1::ecdh(std::span<const uint8_t> buffer,
...
 }
+std::optional<std::string>
+Decred_secp256k1::pubkey_parse(std::span<const uint8_t> buffer) const {
+  ByteArray buffer_data{
+      .data = reinterpret_cast<char *>(const_cast<uint8_t *>(buffer.data())),
+      .length = static_cast<int>(buffer.size())};
+
+  char *result = DecredPubkeyParse(buffer_data);
+  if (!result)
+    return std::nullopt;
```
KY-U: The same point about perhaps documenting or modifying the `nullopt` return.

modules/rustk256/module.cpp
```diff
@@ -16,6 +16,16 @@ K256::private_to_public_key(std::span<const uint8_t> buffer) const {
   return s;
 }
 
+std::optional<std::string>
+K256::pubkey_parse(std::span<const uint8_t> buffer) const {
+  char *p = k256_pubkey_parse(buffer.data(), buffer.size());
+  if (!p)
+    return std::nullopt;
```
KY-U: The same point about perhaps documenting or modifying the `nullopt` return.

modules/secp256k1/module.cpp
```diff
@@ -71,6 +71,24 @@ secp256k1_private_to_public_key(std::span<const uint8_t> buffer) {
...
+  secp256k1_pubkey pubkey;
+  if (!secp256k1_ec_pubkey_parse(secp256k1_ctx, &pubkey, buffer.data(),
+                                 buffer.size())) {
+    return "ERR";
+  }
+  size_t out_len = SECP256K1_PUBKEY_COMPRESSED_LEN;
+  std::vector<uint8_t> out(out_len);
+  if (!secp256k1_ec_pubkey_serialize(secp256k1_ctx, out.data(), &out_len,
+                                     &pubkey, SECP256K1_EC_COMPRESSED)) {
+    return std::nullopt;
```
KY-U: The same point about perhaps documenting or modifying the `nullopt` return.

KY-U (commented):
Found no blocking issue. Made some comments about some `nullopt` returns.

modules/bitcoin/module.cpp
```diff
@@ -393,6 +393,25 @@ Bitcoin::deserialize_block(std::span<const uint8_t> buffer) const {
   return block.GetHash().ToString();
 }
 
+std::optional<std::string>
+Bitcoin::pubkey_parse(std::span<const uint8_t> buffer) const {
```
erickcestari: Does it make sense to implement this target for Bitcoin core since it does the same calls as the secp256k1?

modules/btcd/btcd_wrapper/libbtcd_wrapper.h
erickcestari: If I'm not wrong this comment was generated automatically by cgo based on the comments above `BTCDPubkeyParse`.

modules/btcd/btcd_wrapper/wrapper.go
```diff
@@ -235,6 +235,22 @@ func BTCDFreeString(ptr *C.char) {
 	C.free(unsafe.Pointer(ptr))
 }
 
+// BTCDPubkeyParse parses a SEC1-encoded public key via btcec.ParsePubKey and
+// returns the canonical compressed encoding ("OK:<hex>"), or "ERR" on
+// rejection.
+//
+//export BTCDPubkeyParse
+func BTCDPubkeyParse(data C.ByteArray) *C.char {
```
erickcestari: Same as Bitcoin Core. Does it make sense to test BTCD when we are already testing decresecp256k1?

modules/secp256k1/module.cpp
erickcestari: Probably we could even assert it here. If we parsed a pubkey we should be able to serialize it.

modules/bitcoin/module.cpp
erickcestari: I think an assert would be even better.

erickcestari (changes requested):

brunoerg:
Force-pushed:

1. Removed BITCOIN_CORE and BTCD from the pubkey_parse target.
2. driver.cpp. Moved the is_hybrid / is_compact computation inside #ifdef RUST_K256, merging the two comments into one that now sits with the declarations it explains.
3. Replaced 3 `std::nullopt` returns with assert.

modules/btcd/btcd_wrapper/wrapper.go
brunoerg: I was wondering if they were using the libs incorrectly, but nevermind. it doesn't have any major impact in practice. I removed it.

modules/bitcoin/module.cpp
brunoerg: https://github.com/bitcoinfuzz/bitcoinfuzz/pull/635#discussion_r3876056683

modules/secp256k1/module.cpp
brunoerg: Done.

modules/decredsecp256k1/decredsecp256k1_wrapper/libdecredsecp256k1_wrapper.h
brunoerg: Removed.

driver.cpp
brunoerg: Done.

erickcestari (approved):
ACK 76effd6535f927e1027a312a0d5a9aa2c631345e

Nice!

PR #637, opened 2026-08-14
rustbicoin (feat): add BIP324 EllSwift fuzz targets

victorandre957:
Adds Rust Bitcoin to the existing decode_ellswift and roundtrip_ellswift
BitcoinFuzz targets.

These BIP324 EllSwift scenarios were already exercised against other supported
implementations. Including Rust Bitcoin enables differential fuzzing of EllSwift
decoding and secret-key-to-EllSwift round trips, helping catch behavioral
regressions across implementations.


erickcestari (approved):
ACK 3ee205dffddbd93accfcff48878fb77d3155cd52

Thanks for the PR! Rust-bitcoin uses the secp256k1 library for elliptic curve operations, so this target tests the rust-secp256k1 wrapper rather than a new implementation. I think testing the wrapper is valuable as well.

PR #640, opened 2026-08-20
feat(bech32): add bech32_roundtrip and bech32_convert_bits targets

brunoerg:
This PR adds differential targets for the bech32/bech32m codec:

- `bech32_roundtrip` (Core, rust-bitcoin, btcd, libwally): encode `(hrp, witver, program)`, decode it back, compare `ENC:<address>|DEC:v<n>:<program>`. Covers the encode direction, the full HRP range and witness versions 1..16, which `address_parse` never reached.
- `bech32_convert_bits` (Core, btcd): drives the 5<->8 bit regrouping helper directly, both directions with padding on/off, for the padding rules the roundtrip target cannot hit. Groups are kept in range since Core and btcd diverge by design on out-of-range ones.

Also, `address_parse` now compares future witness versions in their own `WITNESS_UNKNOWN` chain instead of skipping them, and mismatch reports are flushed before the assert.


driver.cpp
```diff
@@ -1018,6 +1043,140 @@ void Driver::SilentPaymentsCreateOutputsTarget(
...
+    if (mask && input.from_bits == 5) {
+      for (uint8_t &b : input.data)
+        b &= 0x1f;
+    }
+
+    std::optional<std::string> last_response{std::nullopt};
+    std::string last_module_name;
+
+    for (auto &module : modules) {
+      std::optional<std::string> res{module.second->bech32_convert_bits(input)};
```
erickcestari, line 1094: This new target is actually 2 in one. I think it would be better if we split both. Also there's a lot of logic inside the `driver.cpp` that could be moved to be a custom mutator.

erickcestari (changes requested):

driver.cpp
brunoerg: 100% agreed. The first version was basically fully vibecoded to find a case I had in mind, now I'm improving it including splitting into two targets. You can review again. Thanks.

erickcestari (approved):
ACK 754aafcf009d5beb99aaefe921c0ecbe977f19c2

PR #641, opened 2026-08-21
electrum: add electrum module

ekzyis:
I have found two bip32 issues in electrum with vibe-coded differential fuzzing in my [fuzz branch](https://github.com/ekzyis/electrum/tree/fuzz) with [atheris](https://github.com/google/atheris):

* https://github.com/spesmilo/electrum/pull/10880
* https://github.com/spesmilo/electrum/pull/10881

This PR adds an electrum module so bitcoinfuzz can catch issues like these (and more).

I have confirmed that bitcoinfuzz would have found both with the bip32 custom mutator and the `bip32_deserialize_extended_key` target added in this PR. This should be reproducible if https://github.com/spesmilo/electrum/commit/6cca1b1f6c6614459969096790cade84a5af0245 is installed.

Will keep this in draft until I have fully reviewed Claude's output.

Depends on #642, #645, #646.


modules/electrum/electrum_lib.py
```diff
@@ -0,0 +1,28 @@
+"""Electrum adapter for bitcoinfuzz differential fuzzing."""
+
+from electrum.bip32 import BIP32Node
+
+
+def bip32_deserialize_extended_key(data):
+    try:
+        text = data.decode()
+    except Exception:
+        return None
```
ekzyis: here @erickcestari (https://github.com/bitcoinfuzz/bitcoinfuzz/pull/642#issuecomment-5417340268)

ekzyis: Actually, I could keep `None` here and it wouldn't be a problem because this is an opt-out for electrum

modules/electrum/electrum_lib.py
```diff
@@ -0,0 +1,41 @@
...
+        return (
+            f"depth={node.depth:02x};"
+            f"fp={node.fingerprint.hex()};"
+            f"child={node.child_number.hex()};"
+            f"chaincode={node.chaincode.hex()};"
+            f"key={key.hex()}"
+        )
+    except Exception:
+        # need to return INVALID on any error after parsing to avoid the harness dismissing it as
+        # NULL in C++
```
erickcestari: nit: This comment can be removed.

```suggestion
```

modules/electrum/electrum_lib.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def bip32_deserialize_extended_key(data):
+    try:
+        text = data.decode()
+    except Exception:
+        return None
+    # electrum pins a specific network and by default it's mainnet
+    if not text.startswith(("xprv", "xpub")):
+        return None
```
erickcestari, line 21: Nice. Good use of `None`.

modules/electrum/electrum_lib.py
erickcestari: I think it would be better to return `INVALID` for two reasons:

1. We could have a single `try/catch` block covering the entire code and return `INVALID` from it.
2. Imagine a scenario where rust-bitcoin is able to parse a BIP32 extended key, but the UTF-8 decoding fails. That could indicate an issue in rust-bitcoin, Electrum, or our harness. If we returned `None`, we would silently skip the comparison and never know about it.

.github/workflows/workflow.yml
erickcestari, line 218: The new module needs to be added to the CI for `bip32_master_keygen` and `bip32_deserialize_extended_key`.

erickcestari (changes requested):
I've run locally and it's working great. There's only some small changes needed:

modules/electrum/electrum_lib.py
ekzyis: done in b856c58

.github/workflows/workflow.yml
ekzyis: done in b856c58

modules/electrum/electrum_lib.py
ekzyis: makes sense, done in b856c58

brunoerg (commented):
LGTM 

@erickcestari do you want to do a final look here?

erickcestari (approved):
ACK fff5c3d51994b5e96608130ca06a01a435be0329

Thanks!

PR #642, opened 2026-08-24
embit: ignore if bip32 fuzz input fails to decode

ekzyis:
This was needed to find the bugs I mentioned in #641.

Without this, I think the `bip32_deserialize_extended_key` target for embit is broken.


erickcestari (approved):
Nice! It's way better to handle the exceptions inside the `.py` files then handling in the `.cpp`

ACK 928ff25869dd0b09437812e96d25a0d916cf9249

PR #645, opened 2026-08-25
refactor: set PYTHON_LDFLAGS once in Makefile

ekzyis:
I was about to add another similar block in #641, but Claude and I think this should do the same thing, and then I can just add `-DELECTRUM` in #641.

_Btw, Claude also flexed that `filter` is superior to `findstring` because it matches whole words rather than substrings, haha_


brunoerg (commented):
> Btw, Claude also flexed that filter is superior to findstring because it matches whole words rather than substrings, haha

Wouldn't it break cases like `CXXFLAGS=-DPYCOIN=1` or `-DPYBITCOINKERNEL=1`, since `filter` only match whole words?

brunoerg (approved):
ACK c02197ece1f124180638b6cfdd2b2e86e05e90af

PR #646, opened 2026-08-25
fix(make): fix collisions between *_main.py files

ekzyis:
The first commit fixes inconsistent naming or a bug. I haven't tested this (couldn't build pybitcoinkernel yet), but wouldn't embit and pybitcoinkernel currently overwrite each other, so they can not be fuzzed together?

The second commit refactors the code to avoid such inconsistencies in the future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,8 +14,7 @@ obj/
 *.jar
 *.zip
 __pycache__
-main.py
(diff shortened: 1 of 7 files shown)


brunoerg (approved):
code review ACK

PR #647, opened 2026-08-28
Add `transaction_eval` target to bitcoinkernel module

KY-U:
Adds the `transaction_eval` target to the bitcoinkernel module so that core's bitcoinkernel ABI can be tested against its own internal behaviour.

I've intentionally not added the target to the other kernel bindings as that's what the `kernel_*` targets are for, comparing external consumers of the public bitcoinkernel interface. This PR aims to cover the specific case where a kernel functionality might not be correctly exposed through btccore's own public kernel ABI.

### DIFF
--- a/modules/bitcoinkernel/module.cpp
+++ b/modules/bitcoinkernel/module.cpp
@@ -7,28 +7,29 @@
 #include <cstdint>
 #include <cstring>
 #include <iomanip>
+#include <ranges>
 #include <span>
 #include <sstream>
 #include <string>
 
 namespace {
-std::string bytes_to_hex(const uint8_t *bytes, int length) {
+std::string bytes_to_hex(std::span<const std::byte> bytes) {
   std::stringstream string_stream;
   string_stream << std::hex;
-  for (int i = 0; i < length; ++i) {
+  for (const auto byte : bytes) {
     string_stream << std::setw(2) << std::setfill('0')
-                  << static_cast<int>(bytes[i]);
+                  << std::to_integer<int>(byte);
   }
 
   return string_stream.str();
 }
 
-std::string hash_bytes_to_hex(const uint8_t *bytes, int length) {
+std::string hash_bytes_to_hex(std::span<const std::byte> bytes) {
   std::stringstream string_stream;
   string_stream << std::hex;
-  for (int i = length - 1; i >= 0; --i) {
+  for (const auto byte : bytes | std::views::reverse) {
     string_stream << std::setw(2) << std::setfill('0')
-                  << static_cast<int>(bytes[i]);
+                  << std::to_integer<int>(byte);
   }
 
   return string_stream.str();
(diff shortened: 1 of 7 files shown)


modules/bitcoinkernel/module.cpp
```diff
@@ -180,6 +182,54 @@ char *libbitcoinkernel_block_check(std::span<const uint8_t> buffer) {
     return strdup(result.c_str());
   }
 }
+
+char *libbitcoinkernel_transaction_eval(std::span<const uint8_t> buffer) {
+  try {
+    std::span<const std::byte> raw_span{(const std::byte *)buffer.data(),
```
brunoerg: We should avoid C-style cast in C++ code.

```suggestion
    std::span<const std::byte> raw_span{
    reinterpret_cast<const std::byte*>(buffer.data()),
```

brunoerg: Or perhaps...

```cpp
const auto raw_span = std::as_bytes(buffer);
```

modules/bitcoinkernelvariant/module.cpp
```diff
@@ -221,5 +271,17 @@ std::optional<std::string> BitcoinKernelVariant::kernel_block_check(
   return result;
 }
 
+// Core target
```
brunoerg: nit: this comment isn't necessary

```suggestion
```

modules/bitcoinkernel/module.cpp
```diff
@@ -180,6 +182,54 @@ char *libbitcoinkernel_block_check(std::span<const uint8_t> buffer) {
...
+                                        buffer.size()};
+    btck::Transaction transaction{raw_span};
+    btck::TxValidationState state{};
+    if (!btck::CheckTransaction(transaction, state))
+      return strdup("0");
+
+    // Since bitcoinkernel does not provide a public interface for getting the
+    // witness hash, we calculate it manually using core's internal API.
```
brunoerg, line 185: @sedited Isn't there a way to get the witness hash? I think it would be great to have it.

modules/bitcoinkernel/module.cpp
```diff
@@ -180,6 +182,54 @@ char *libbitcoinkernel_block_check(std::span<const uint8_t> buffer) {
...
+      std::memcpy(witness_hash.data(), txid.data(), txid.size());
+    } else {
+      const std::span<const unsigned char> hash_input{
+          reinterpret_cast<const unsigned char *>(transaction_bytes.data()),
+          transaction_bytes.size()};
+      CHash256{}.Write(hash_input).Finalize(witness_hash);
+    }
+
+    std::string result =
+        hash_bytes_to_hex((const uint8_t *)witness_hash.data(),
```
brunoerg: Avoid C-style cast

modules/bitcoinkernel/module.cpp
sedited: Not yet, will have to add it.

modules/bitcoinkernel/module.cpp
KY-U: Changed casts and iterators for idiomatic c++ patterns.

modules/bitcoinkernel/module.cpp
KY-U: Done.

modules/bitcoinkernelvariant/module.cpp
KY-U: Removed.

brunoerg (commented):
ACK 43ea6f09bf75de9001df646242daaed9dcab726d

I will just wait CI run to finish.

modules/bitcoinkernel/module.cpp
```diff
@@ -7,28 +7,29 @@
...
 }
 
-std::string hash_bytes_to_hex(const uint8_t *bytes, int length) {
+std::string hash_bytes_to_hex(std::span<const std::byte> bytes) {
   std::stringstream string_stream;
   string_stream << std::hex;
-  for (int i = length - 1; i >= 0; --i) {
+  for (const auto byte : bytes | std::views::reverse) {
     string_stream << std::setw(2) << std::setfill('0')
-                  << static_cast<int>(bytes[i]);
```
brunoerg, line 31: `static_cast` is C++ idiomatic but good to see std::to_integer here, lgtm.

brunoerg (approved):
ACK fa19fa7e82a26760423fb0357abbaef06066e34f

Improvements, if necessary, can be done in a follow-up. Thanks.

PR #654, opened 2026-09-18
lnd: fix locale-dependent symbol renaming and char* argument type mismatch

usefahmed07:
Two independent bug fixes in the LND module, unrelated to any pending feature work:

1. **scripts/rename_cgo_symbols.sh**: the script didn't pin `LC_ALL=C`
   before `nm`/`sort`/`comm`. Under a non-C locale (e.g. `en_US.UTF-8`),
   `comm -23` can fail to exclude the exported symbols from the rename
   list, causing them to be renamed by mistake (e.g.
   `LndDeserializeInvoice` becoming `lnd_LndDeserializeInvoice`), which
   breaks linking for all LND exported functions. Fixed by adding
   `LC_ALL=C` before every `nm`/`sort`/`comm` invocation.

2. **modules/lnd/module.cpp**: LND's cgo-exported functions
   (`LndDeserializeInvoice`, `LndParseP2pLightningMessage`,
   `LndDecodeOnion`) take `char*`, not `const char*`, in their generated
   header. The module was passing `const char*` directly, relying on an
   implicit conversion. Fixed by using `const_cast<char*>` at each call
   site.

Also removed `modules/lnd/lnd_wrapper/libscript.h`, a stale
cgo-generated header from an old build that was accidentally tracked in
git and no longer matches the current generated output
(`liblnd_wrapper.h`), and updated `module.cpp`'s include accordingly.

No behavior change intended beyond fixing the two bugs above.


erickcestari (approved):
Thanks!

ACK a009f802e6c9c358733d6dec403a548265c37119

Issue #586, opened 2026-06-24
Implement the deserialize_offer target for LND


erickcestari:
Once the codec is finished and reviewed, we can add LND to the deserialize_offer target.

https://github.com/lightningnetwork/lnd/issues/10736

usefahmed07:
Done in #654

erickcestari:
> Done in [#654](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/654)

Thanks for the PR @usefahmed07  !! It seems that #654 is not implementing the `deserialize_offer` target for LND.

usefahmed07:
> > Done in [#654](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/654)
> 
> Thanks for the PR [@usefahmed07](https://github.com/usefahmed07) !! It seems that [#654](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/654) is not implementing the `deserialize_offer` target for LND.

I will try again at the same time. Can you please guide me to a solution? Thanks!

usefahmed07:
> > Done in [#654](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/654)
> 
> Thanks for the PR [@usefahmed07](https://github.com/usefahmed07) !! It seems that [#654](https://github.com/bitcoinfuzz/bitcoinfuzz/pull/654) is not implementing the `deserialize_offer` target for LND.

My bad, wrong PR link! https://github.com/bitcoinfuzz/bitcoinfuzz/pull/654 is just two unrelated bug fixes, not the target.

The real deserialize_offer target is in https://github.com/bitcoinfuzz/bitcoinfuzz/pull/655 — it's a draft for now since it depends on https://github.com/lightningnetwork/lnd/pull/11199 getting merged first. Thanks.

PR #655, opened 2026-09-18
lnd: implement deserialize_offer target

usefahmed07:
Implements the `deserialize_offer` fuzz target for LND, part of #586.

A heads up on the branch structure: this is stacked on top of #654, not `v2` directly. That's because the include fix and `const_cast<char*>` changes in #654 are actually needed for this code to compile. So #654 should probably go in first, and then this can get rebased onto `v2`.

There's also a real blocker here: `LndDeserializeOffer` (in `wrapper.go`) calls `bolt12.DecodeOffer`, which doesn't exist in upstream LND yet — only on my fork. I opened lightningnetwork/lnd#11199 to export it, but it hasn't been merged. Until it is, building this locally means pointing `go.mod` at my fork with a temporary `replace`, which I've deliberately left out of this PR since it can't be part of the final merge.

What's actually in the diff:
- `module.h` / `module.cpp`: wiring up the `deserialize_offer` override
- `wrapper.go`: the new `LndDeserializeOffer` function, which calls `bolt12.DecodeOffer` and formats the result the same way Eclair's `decodeOffer()` does (`CHAINS=;METADATA=;[AMOUNT=];[CURRENCY=];DESCRIPTION=;FEATURES=;ABSOLUTE_EXPIRY=;PATH_<i>_HOP=;ISSUER=;QUANTITY=;ISSUER_ID=`)

I built it with `CXXFLAGS="-DLND" CXX=clang++ ./scripts/auto_build.py` and ran it against a simple offer TLV (`LOG_OUTPUTS=1 MODULES="LND" FUZZ=deserialize_offer ./bitcoinfuzz -runs=1 <file>`) — output came out as expected.

Two things I'm not fully sure about yet, marked with `// VERIFY` in the code:
- Whether the chain hash byte order LND prints in `CHAINS` actually matches Eclair's `BlockHash.toString()` — Eclair might use a reversed/display convention
- Whether the pubkey format I'm using for `PATH_<i>_HOP` / `ISSUER_ID` (`SerializeCompressed()` as hex) matches Eclair's `PublicKey.toString()`

I didn't want to spend time spinning up Eclair just to check these, but happy to if that's the preferred way to confirm rather than leaving it to review/CI.

Marking this as a draft since it's blocked on #654 and lightningnetwork/lnd#11199.


erickcestari (changes requested):
While we wait for LND to expose a public API for decoding offers, CI caught two issues with this PR: the code formatting and the LND module not compiling. It also needs rebase.

usefahmed07:
> While we wait for LND to expose a public API for decoding offers, CI caught two issues with this PR: the code formatting and the LND module not compiling. It also needs rebase.

IDK why [Docker Image Check] are failing; do you have a suggestion?

erickcestari:
> > While we wait for LND to expose a public API for decoding offers, CI caught two issues with this PR: the code formatting and the LND module not compiling. It also needs rebase.
> 
> IDK why [Docker Image Check] are failing; do you have a suggestion?

`Error: ./wrapper.go:42:23: undefined: bolt12.DecodeOffer`

Oh, it’s expected to be failing. Your upstream PR in `lnd` hasn’t been merged yet, so `DecodeOffer` isn’t a public function. Because of that, this PR won’t compile until that PR is merged, which is fine.

Let’s wait for it. The LND team might prefer a different public API rather than exposing `DecodeOffer`, in which case this PR may need to be updated accordingly.

usefahmed07:
Hi @erickcestari, thanks for the explanation! It's not blocked on lnd anymore.

bitromortac pointed out on the lnd PR that they're keeping `decodeOffer` unexported, and that `DecodeOfferString` already covers what the fuzz target needs. So I changed `LndDeserializeOffer` to use `bolt12.DecodeOfferString` and closed lnd#11199.

To make that work I also had to bump lnd to current master, which needs Go 1.26.8, so I updated the Dockerfiles too. I tried it locally on a spec test vector and it decodes fine, and garbage input gets rejected.

One heads-up: `DecodeOfferString` is a bit stricter than the old path (it checks chain, features and expiry), so if we ever see diffs against other implementations on odd offers, that's probably why.

Let me know if anything looks off!

usefahmed07:
Good Now

erickcestari:
> Hi @erickcestari, thanks for the explanation! It's not blocked on lnd anymore.
> 
> bitromortac pointed out on the lnd PR that they're keeping `decodeOffer` unexported, and that `DecodeOfferString` already covers what the fuzz target needs. So I changed `LndDeserializeOffer` to use `bolt12.DecodeOfferString` and closed lnd#11199.
> 
> To make that work I also had to bump lnd to current master, which needs Go 1.26.8, so I updated the Dockerfiles too. I tried it locally on a spec test vector and it decodes fine, and garbage input gets rejected.
> 
> One heads-up: `DecodeOfferString` is a bit stricter than the old path (it checks chain, features and expiry), so if we ever see diffs against other implementations on odd offers, that's probably why.
> 
> Let me know if anything looks off!

True, I completely forgot that they already a public API. I'll review the PR!

usefahmed07:
> > Hi @erickcestari, thanks for the explanation! It's not blocked on lnd anymore.
> > bitromortac pointed out on the lnd PR that they're keeping `decodeOffer` unexported, and that `DecodeOfferString` already covers what the fuzz target needs. So I changed `LndDeserializeOffer` to use `bolt12.DecodeOfferString` and closed lnd#11199.
> > To make that work I also had to bump lnd to current master, which needs Go 1.26.8, so I updated the Dockerfiles too. I tried it locally on a spec test vector and it decodes fine, and garbage input gets rejected.
> > One heads-up: `DecodeOfferString` is a bit stricter than the old path (it checks chain, features and expiry), so if we ever see diffs against other implementations on odd offers, that's probably why.
> > Let me know if anything looks off!
> 
> True, I completely forgot that they already a public API. I'll review the PR!

Take Your Time

modules/lnd/lnd_wrapper/wrapper.go
```diff
@@ -28,6 +30,124 @@ import (
...
+				sb.WriteString(";")
+			}
+			// VERIFY: byte order against Eclair's BlockHash.toString().
+			// Eclair typically displays chain/block hashes reversed
+			// (big-endian display of a little-endian-stored hash);
+			// this writes the raw bytes as-is. Flip with a reversed
+			// copy here if a differential run shows a mismatch.
+			sb.WriteString(fmt.Sprintf("%x", c))
+		}
+	})
```
erickcestari: If the CHAINS are empty we should fallback to mainet chain.

```suggestion
	// CHAINS. An absent or empty offer_chains means bitcoin mainnet, which
	// is what Eclair and LDK print, so fall back to the same hash here.
	sb.WriteString("CHAINS=")
	var chains [][32]byte
	offer.OfferChains.WhenSomeV(func(rec bolt12.ChainsRecord) {
		chains = rec.Chains
	})
	if len(chains) == 0 {
		chains = [][32]byte{activeChain}
	}
	for i, c := range chains {
		if i > 0 {
			sb.WriteString(";")
		}
		// VERIFY: byte order against Eclair's BlockHash.toString().
		// Eclair typically displays chain/block hashes reversed
		// (big-endian display of a little-endian-stored hash);
		// this writes the raw bytes as-is. Flip with a reversed
		// copy here if a differential run shows a mismatch.
		sb.WriteString(fmt.Sprintf("%x", c))
	}
```

modules/lnd/lnd_wrapper/wrapper.go
```diff
@@ -28,6 +30,124 @@ import (
...
+	// timestamp is pinned to the epoch so results don't depend on wall-clock
+	// time (no offer is treated as expired), and the active chain is mainnet.
+	activeChain := [32]byte(*chaincfg.MainNetParams.GenesisHash)
+
+	offer, err := bolt12.DecodeOfferString(
+		offerStr, time.Unix(0, 0), activeChain,
+	)
+	if err != nil {
+		return C.CString("")
+	}
```
erickcestari, line 71: There are these edge case that is worth skipping.

```suggestion
	if err != nil {
		// Return null so the driver skips lnd for inputs only lnd rejects,
		// instead of reporting a false mismatch.
		switch {
		// lnd applies these reader gates while decoding. The spec
		// scopes them to responding, which is where LDK applies them,
		// and the Eclair and CLN harnesses skip them.
		case errors.Is(err, bolt12.ErrUnknownEvenFeature),
			errors.Is(err, bolt12.ErrUnsupportedChain):
			return nil

		// lnd rejects non-minimal feature vectors; the others accept
		// them.
		// TODO: drop this once the others reject them too, see
		// https://github.com/lightning/bolts/pull/1341.
		case errors.Is(err, bolt12.ErrNonMinimalFeatures):
			return nil
		}

		return C.CString("")
	}
```

erickcestari (changes requested):
Overall LGTM! It's only missing these changes and applying the request changes and it should be good to go!

The LND module needs to be added into the CI for the `deserialize_offer` target and also in the `docker-compose.yml` file

```
- corpus: "deserialize_offer_clean"
            target: "deserialize_offer"
            cxxflags: "-DLDK -DCLIGHTNING -DECLAIR -DLIGHTNING_KMP -DCUSTOM_MUTATOR_BOLT12_OFFER"
            setup_java: "true"
```

usefahmed07:
> Overall LGTM! It's only missing these changes and applying the request changes and it should be good to go!
> 
> The LND module needs to be added into the CI for the `deserialize_offer` target and also in the `docker-compose.yml` file
> 
> ```
> - corpus: "deserialize_offer_clean"
>             target: "deserialize_offer"
>             cxxflags: "-DLDK -DCLIGHTNING -DECLAIR -DLIGHTNING_KMP -DCUSTOM_MUTATOR_BOLT12_OFFER"
>             asan_options: "detect_leaks=0:detect_container_overflow=0"
>             setup_java: "true"
> ```
Thanks @erickcestari! Addressed all three:

- Added `-DLND` to the `deserialize_offer` target's cxxflags in both workflow.yml and docker-compose.yml, so the lnd module is actually included in that CI job.
- CHAINS now falls back to the mainnet genesis hash when offer_chains is empty, matching what Eclair and LDK print.
- lnd-only rejections (unknown even feature, unsupported chain, non-minimal features) now return nil instead of an empty string, so the driver skips lnd for those instead of reporting a false mismatch.

Pushed in cdc5fd8. CI is green — let me know if anything else needs adjusting.

usefahmed07:
> Overall LGTM! It's only missing these changes and applying the request changes and it should be good to go!
> 
> The LND module needs to be added into the CI for the `deserialize_offer` target and also in the `docker-compose.yml` file
> 
> ```
> - corpus: "deserialize_offer_clean"
>             target: "deserialize_offer"
>             cxxflags: "-DLDK -DCLIGHTNING -DECLAIR -DLIGHTNING_KMP -DCUSTOM_MUTATOR_BOLT12_OFFER"
>             asan_options: "detect_leaks=0:detect_container_overflow=0"
>             setup_java: "true"
> ```

Thanks @erickcestari! Addressed all three:

- Added `-DLND` to the `deserialize_offer` target's cxxflags in both workflow.yml and docker-compose.yml, so the lnd module is actually included in that CI job.
- CHAINS now falls back to the mainnet genesis hash when offer_chains is empty, matching what Eclair and LDK print.
- lnd-only rejections (unknown even feature, unsupported chain, non-minimal features) now return nil instead of an empty string, so the driver skips lnd for those instead of reporting a false mismatch.

let me know if anything else needs adjusting.

erickcestari:
> > Overall LGTM! It's only missing these changes and applying the request changes and it should be good to go!
> > The LND module needs to be added into the CI for the `deserialize_offer` target and also in the `docker-compose.yml` file
> > ```
> > - corpus: "deserialize_offer_clean"
> >             target: "deserialize_offer"
> >             cxxflags: "-DLDK -DCLIGHTNING -DECLAIR -DLIGHTNING_KMP -DCUSTOM_MUTATOR_BOLT12_OFFER"
> >             asan_options: "detect_leaks=0:detect_container_overflow=0"
> >             setup_java: "true"
> > ```
> 
> Thanks @erickcestari! Addressed all three:
> 
> * Added `-DLND` to the `deserialize_offer` target's cxxflags in both workflow.yml and docker-compose.yml, so the lnd module is actually included in that CI job.
> * CHAINS now falls back to the mainnet genesis hash when offer_chains is empty, matching what Eclair and LDK print.
> * lnd-only rejections (unknown even feature, unsupported chain, non-minimal features) now return nil instead of an empty string, so the driver skips lnd for those instead of reporting a false mismatch.
> 
> let me know if anything else needs adjusting.

Can you squash the commits into one?

usefahmed07:
> > > Overall LGTM! It's only missing these changes and applying the request changes and it should be good to go!
> > > The LND module needs to be added into the CI for the `deserialize_offer` target and also in the `docker-compose.yml` file
> > > ```
> > > - corpus: "deserialize_offer_clean"
> > >             target: "deserialize_offer"
> > >             cxxflags: "-DLDK -DCLIGHTNING -DECLAIR -DLIGHTNING_KMP -DCUSTOM_MUTATOR_BOLT12_OFFER"
> > >             asan_options: "detect_leaks=0:detect_container_overflow=0"
> > >             setup_java: "true"
> > > ```
> > 
> > 
> > Thanks @erickcestari! Addressed all three:
> > 
> > * Added `-DLND` to the `deserialize_offer` target's cxxflags in both workflow.yml and docker-compose.yml, so the lnd module is actually included in that CI job.
> > * CHAINS now falls back to the mainnet genesis hash when offer_chains is empty, matching what Eclair and LDK print.
> > * lnd-only rejections (unknown even feature, unsupported chain, non-minimal features) now return nil instead of an empty string, so the driver skips lnd for those instead of reporting a false mismatch.
> > 
> > let me know if anything else needs adjusting.
> 
> Can you squash the commits into one?

Done, squashed into one commit

erickcestari (approved):
Thanks!!

ACK 46d75ba7ca24c471057c2a4e4477e16ad3257b1f

